Issue #14, opened 2026-07-02
Bug: set-grid-size leaves three divergent config objects, desyncing overlay/control from the wall


NilsR0711:
**Severity: high**

The resize pipeline relies on one shared config object, but `setGridSize` replaces it, breaking the shared reference.

**Evidence**
- `packages/streamwall/src/main/StreamWindow.ts:241-243` (`setGridSize` does `this.config = {...this.config, cols, rows}`), `:147-159` (`handleResize` mutates `this.config`; comment: "config is shared by reference").
- `packages/streamwall/src/main/index.ts:279-301` (`streamWindowConfig` === `clientState.config` === `streamWindow.config`), `:457-461,:483-486`.

**Failure scenario**
Send `set-grid-size` (works) → then resize/maximize the window: `handleResize` updates only `streamWindow.config` (now a detached copy); the emitted `resize` → `updateState({})` broadcasts the old width/height. Tiles are laid out for the new size but overlay/control grid draws the old size — permanent desync until the next `set-grid-size`.

**Fix**
Have `setGridSize` mutate in place (`this.config.cols = cols; this.config.rows = rows`) so the shared reference is preserved, or make config flow one-way with a single owner and explicit `updateState({ config })` on both resize and grid change.

_Source: main-process review_

PR #100, opened 2026-07-03
feat(streamwall): add keyboard shortcut to quit (frameless windows)

NilsR0711:
## Problem

[streamwall/streamwall#14](https://github.com/streamwall/streamwall/issues/14) asks for two things: a frameless stream window, and *"a way to trigger `app.quit()` from the UI/keypress events."*

The frameless window is already supported in v2 (`window.frameless`), but the quit part is missing. Today the app only exits when the stream window is closed (`process.exit(0)`), and the control window is created with `closable: false`. When the stream window runs frameless it has no title bar, no close button, and its menu is removed — so there is **no in-app way to quit**. Users are left to force-kill the process or rely on OS-specific tricks (Alt+F4, the macOS app menu).

## Solution

Add a focus-scoped keyboard shortcut (`CommandOrControl+Q`) in the main process that calls `app.quit()`:

- Registered on `browser-window-focus`, unregistered once the app loses focus entirely (checked on `browser-window-blur`, deferred so switching between the app's own windows doesn't drop it), and torn down on `will-quit`.
- Scoping to app focus avoids hijacking Cmd/Ctrl+Q system-wide. Using a **main-process global shortcut** (rather than a renderer hotkey) keeps it working even when a media `WebContentsView` on the wall holds focus — exactly the frameless scenario.

`app.quit()` closes the (closable) stream window, whose existing `close` handler calls `process.exit(0)`, so termination is guaranteed even though the control window is non-closable.

## Architecture notes

- The registration/teardown logic lives in a small, dependency-injected module (`quitShortcut.ts`) that imports Electron only as **types**, so it is unit-testable without an Electron runtime.
- `index.ts` wires it up with the real `app` / `globalShortcut` / `BrowserWindow.getFocusedWindow`.

## Test coverage

Adds a `node:test` runner (run through `esbuild-register`, already a dependency — no new tooling) and **10 tests** covering: accelerator value, registration on focus, quitting on trigger, no double-registration, teardown when focus is lost, staying registered across window switches, eager registration when a window is already focused, and `will-quit` teardown.

```
npm test        # 10 passing
```

## Risks / breaking changes

None. Default behavior is unchanged (a framed window still closes normally); this only adds a keyboard path to quit. No new runtime dependencies.

## Docs

- README hotkeys section documents Cmd/Ctrl+Q.
- `example.config.toml` notes the quit key next to `window.frameless`.

Addresses [streamwall/streamwall#14](https://github.com/streamwall/streamwall/issues/14).

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ npm start -- --data.json-url="https://your-site/api/streams.json" --data.toml-fi
 
 ## Hotkeys
 
+**cmd/ctrl+q** quits Streamwall. It works whenever any Streamwall window is focused, and is the only way to exit when the stream window is configured to be frameless (a frameless window has no title bar or close button — see `window.frameless` in `example.config.toml`).
+
 The following hotkeys are available with the "control" webpage focused:
 
 - **alt+[1...9]**: Listen to the numbered stream
--- a/example.config.toml
+++ b/example.config.toml
@@ -6,6 +6,7 @@
 # Position a frameless window (useful for capturing w/ fixed screen positions)
 #x = 0
 #y = 0
+# A frameless window has no title bar or close button; press Cmd/Ctrl+Q to quit.
 #frameless = false
 
 # Set the background color (useful for chroma-keying)
--- a/packages/streamwall/src/main/index.ts
+++ b/packages/streamwall/src/main/index.ts
@@ -1,6 +1,6 @@
 import TOML from '@iarna/toml'
 import * as Sentry from '@sentry/electron/main'
-import { BrowserWindow, app } from 'electron'
+import { BrowserWindow, app, globalShortcut } from 'electron'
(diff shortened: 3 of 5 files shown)


NilsR0711:
Closing: this PR addresses the wrong issue.

It implements the **upstream** `streamwall/streamwall#14` (a frameless-window quit shortcut), which `gh` resolved to at session start while `origin` still pointed at the upstream remote. The intended issue is this repo's own **#14** — *"set-grid-size leaves three divergent config objects, desyncing overlay/control from the wall"* — which lives on `deps-modernization`. That fix is opened as a separate PR against `deps-modernization`.

The frameless-quit change is preserved on branch `claude/sharp-borg-d82347` history (commits up to this PR) should it be wanted as a standalone contribution later.

PR #101, opened 2026-07-03
fix(streamwall): deny window.open and guard redirects in stream views and the browse window

### LINKED ISSUE #5: Security: deny window.open and guard redirects in stream views and the browse window
**Severity: medium**

No `setWindowOpenHandler` exists anywhere; `will-navigate` is handled but `will-redirect` is not.

**Evidence**
- `packages/streamwall/src/main/StreamWindow.ts:161-188` — stream view has no window-open handler and no `will-redirect` guard.
- `packages/streamwall/src/main/index.ts:399-408` — the `browse` window likewise sets no handler.
- Confirmed: `setWindowOpenHandler` appears nowhere in the codebase.

**Attack scenario**
A loaded stream page calls `window.open(...)` or serves a 302 redirect (bypassing the URL-equality `will-navigate` check) → popups/phishing surface and navigation escape from the intended single URL.

**Remediation**
`webContents.setWindowOpenHandler(() => ({ action: 'deny' }))` on every stream view and the browse window; add a `will-redirect` guard mirroring `will-navigate`.

_Source: security audit_

NilsR0711:
## Problem

`setWindowOpenHandler` was set nowhere in the codebase, and stream views guarded only `will-navigate` (against the current URL) — not `will-redirect`. A loaded stream page could therefore:

- **spawn popups** via `window.open(...)` / `target="_blank"` (phishing/popup surface), and
- **escape its intended URL** through a 302 redirect on a reload — `will-navigate` allows a self-reload, and the server can then 302 that reload elsewhere, which the URL-equality check never sees.

The `browse` window set no window-open handler at all.

Fixes issue #5 (security audit, severity: medium).

## Technical solution

New module `packages/streamwall/src/main/navigationSecurity.ts` centralises the guards:

- **`denyWindowOpen(webContents)`** — installs a `setWindowOpenHandler` returning `{ action: 'deny' }` for every popup request.
- **`secureStreamView(webContents)`** — denies popups **and** installs a single guard on both `will-navigate` and `will-redirect` that blocks navigation away from the committed URL.

Wiring:
- `StreamWindow.createView()` calls `secureStreamView(view.webContents)`, replacing the inline `will-navigate`-only handler. This composes with the per-view ephemeral partition + `hardenSession` added in #97.
- `index.ts` calls `denyWindowOpen(browseWindow.webContents)` on each freshly created browse window (which also keeps #97's `hardenSession`). The browse window is intentionally allowed to navigate — it exists to load an operator-supplied URL — so it only denies popups.

### Redirect handling (important nuance)

Per Electron's docs, `will-navigate` does **not** fire for main-process `loadURL` calls, but `will-redirect` **does** fire for server redirects during those loads. A naive "block every redirect to a non-current URL" would break legitimate stream URLs that 302 on load (http→https, CDN, shortlinks), because `getURL()` is still empty before the first commit.

The guard therefore:
1. allows navigation/redirect to the URL the view is already on (self-reload);
2. allows redirects while the initial load is still resolving (`getURL()` empty, nothing committed yet — these are the operator URL's own 302s);
3. blocks everything else — i.e. once a page has committed, any navigation or redirect to a different URL is denied. This is exactly the reload-then-302 escape from the issue.

## Architecture decisions

- **Extract, don't inline.** The guard logic lives in one module and takes a `WebContents`, using only its public surface (`on`, `getURL`, `setWindowOpenHandler`). That makes the security-critical logic unit-testable without a running Electron app.
- **Behaviour-preserving for reloads.** The self-reload allowance (and its log line) from `Allow pages to reload themselves` is retained and now also applies to redirects.
- **No new dependencies.** Tests use the repo's existing Node built-in test runner (`node --test`, native TypeScript), matching `src/util.test.ts` / `src/main/partitions.test.ts`.

## Risks / breaking changes

- No breaking changes to intended behaviour: streams still load (including through initial-load redirects) and can reload themselves; popups and post-load escapes are denied.
- The browse window can still navigate/redirect freely (by design); only popups are denied.

## Test coverage

`navigationSecurity.test.ts` (Node test runner, 8 tests, all green) covers:
- popups denied by both `denyWindowOpen` and `secureStreamView`;
- `will-navigate` to a different URL blocked; to the same URL allowed (self-reload);
- `will-redirect` away from a committed page blocked (the 302 escape); to the same URL allowed;
- `will-redirect` allowed while the initial load is still resolving (empty `getURL()`).

Run with `npm -w streamwall test` (whole streamwall suite is green, incl. sibling `partitions`/`util` tests).

The one-line wiring calls into real Electron `WebContents` are integration glue that cannot run in a unit test (no Electron runtime); the guarded logic they delegate to is fully covered above.

Closes #5

### DIFF
--- a/packages/streamwall/src/main/StreamWindow.ts
+++ b/packages/streamwall/src/main/StreamWindow.ts
@@ -17,6 +17,7 @@ import {
 } from 'streamwall-shared'
 import { createActor, EventFrom, SnapshotFrom } from 'xstate'
 import { loadHTML } from './loadHTML'
+import { secureStreamView } from './navigationSecurity'
 import { allocateViewPartition, hardenSession } from './partitions'
 import viewStateMachine, { ViewActor } from './viewStateMachine'
 
@@ -165,14 +166,9 @@ export default class StreamWindow extends EventEmitter<StreamWindowEventMap> {
 
     const viewId = view.webContents.id
 
-    // Prevent view pages from navigating away from the specified URL.
-    view.webContents.on('will-navigate', (ev) => {
-      if (ev.url === view.webContents.getURL()) {
-        console.log('Allowing page to reload:', ev.url)
-        return
-      }
-      ev.preventDefault()
-    })
+    // Lock the view to its stream URL: deny popups and block navigation/redirect
+    // escapes while still allowing the page to reload itself.
+    secureStreamView(view.webContents)
 
     // Hidden window used for loading the BrowserView before it's positioned in the wall
     const offscreenWin = new BrowserWindow({
--- a/packages/streamwall/src/main/index.ts
+++ b/packages/streamwall/src/main/index.ts
@@ -27,6 +27,7 @@ import {
   pollDataURL,
   watchDataFile,
 } from './data'
+import { denyWindowOpen } from './navigationSecurity'
 import { BROWSE_PARTITION, hardenSession } from './partitions'
 import { loadStorage } from './storage'
 import StreamdelayClient from './StreamdelayClient'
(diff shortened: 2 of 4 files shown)


NilsR0711:
bugbot run

PR #119, opened 2026-07-04
feat: validate all external input with Zod (config, stream data, WS messages, Yjs updates)

### LINKED ISSUE #11: Validate all external input with Zod (config, stream data, WS control messages, Yjs updates)
**Severity: high (security + robustness)**

Untrusted data enters at many points with no schema validation; 6 of 11 existing TODO/FIXME markers are literally "validate with Zod".

**Evidence**
- Control server: `packages/streamwall-control-server/src/index.ts:390-421` — `JSON.parse` then dispatched by `msg.type` with no field validation; `:209-256,:374-388` — `Y.applyUpdate(streamwallConn.stateDoc, new Uint8Array(rawData))` applies arbitrary client binary to the shared doc (gated only by role).
- Config: `packages/streamwall/src/main/index.ts:102-105` — malformed TOML throws a raw parse exception; yargs performs no schema validation (typos/stale keys silently ignored).
- Stream data: `packages/streamwall/src/main/data.ts:53-58` — non-array `streams` silently yields `[]`; `:26` — JSON URL body cast unchecked.
- Existing TODOs: `data.ts:54`, `renderer/control.tsx:31`, `control-ui/src/index.tsx:408,904,2348`.

**Impact**
State corruption / DoS across all clients and the desktop; unbounded persisted-doc growth; unfriendly startup crashes on bad config.

**Remediation**
Introduce Zod schemas for config, stream lists, and every WS command; bound numeric fields; verify the Yjs document shape after applying updates; cap message/document sizes; on failure print which key/file/line is wrong.

_Source: security audit + feature/DX analysis + TODO inventory_

NilsR0711:
## Problem

Streamwall accepts untrusted data at many trust boundaries with no schema validation — 6 of the repo's 11 TODO/FIXME markers literally said "validate with Zod" (issue #11, **high severity, security + robustness**):

- **Control server** parsed WebSocket messages with `JSON.parse` and dispatched by `msg.type` with no field validation, and applied arbitrary client binary directly to the shared Yjs doc (gated only by role).
- **Config**: malformed TOML threw a raw parser exception at startup; yargs performed no schema validation, so typos and out-of-range values silently took effect.
- **Stream data**: JSON-URL and TOML-file input was cast straight to the internal type; internal identity fields could be forged.

Left unchecked this allowed state corruption / DoS across all clients and the desktop, unbounded persisted-doc growth, and unfriendly startup crashes.

## Solution

Zod schemas are introduced as the single source of truth for every untrusted input named in the issue title — config, stream data, WS control messages, and Yjs updates — with a compile-time guard keeping the schemas' inferred types assignable to the existing hand-written types.

**`streamwall-shared`** — new runtime schemas:
- `streamDataInputSchema` / `parseStreamList`: validate stream entries; drop-and-report invalid ones (tolerant, per-entry) rather than poisoning the batch; strip internal `_id` / `_dataSource` so a source cannot forge identity/provenance.
- `controlCommandSchema` / `controlCommandMessageSchema`: a discriminated union over every command, bounding numeric fields (grid size, view index, rotation) and rejecting unknown/prototype-polluting command types.
- `controlStateMessageSchema`: structural validation of the uplink state message.
- `isValidStateDocShape`: verifies the Yjs doc still matches its narrow shape (a single `views` map of integer-string keys → nested maps whose only key is a string `streamId`).

**`streamwall-control-server`**:
- Every inbound client command is validated before the `roleCan` check — an admin passes every role check, so this schema is the only barrier stopping a malformed/unknown command from reaching (and executing on) the desktop. Rejected commands get an `invalid message` response and are never relayed upstream.
- Uplink `state` messages are validated (a payload-less `{"type":"state"}` frame used to wrap the shared `StateWrapper` around `undefined`, wiring a broken connection that crashed clients on `view()`).
- `applyValidatedDocUpdate` replays each untrusted binary Yjs update onto a throwaway probe seeded from the live doc, enforces per-update and whole-doc size caps, and checks the shape **before** touching the live doc — so a rejected update never fires the doc's update event or the broadcast it drives (a CRDT cannot cleanly unwind a bad update). Limits are configurable via `STREAMWALL_WS_UPDATE_MAX_BYTES` and `STREAMWALL_STATE_DOC_MAX_BYTES`.

**`streamwall` (desktop)**:
- `data.ts` runs both stream sources through `parseStreamList`; `watchDataFile` also closes its watcher on generator return.
- `parseConfigToml` turns a TOML syntax error into a `ConfigError` naming the file (the parser already pinpoints row/column); `validateConfig` checks the resolved config against a schema and names the offending key; `init()` prints just the message before exiting instead of dumping a stack trace.

## Architecture notes

- Schemas live in `streamwall-shared` so the server, desktop, and (future) clients share one definition; the config schema is desktop-specific and lives in `streamwall`.
- Yjs shape verification uses a **probe doc** rather than mutate-then-rollback, because CRDT updates can't be cleanly reverted. Top-level Yjs types arrive from an update untyped, so the shape check materializes `views` via a typed accessor.

## Risks / behavior changes

- Invalid config values (negative window dimensions, `NaN`, wrong types) now **fail startup with a clear message** instead of being silently accepted. Grid dimensions keep their existing clamp behavior (only positivity/type is enforced).
- WS commands that are malformed or unknown are now **rejected** rather than forwarded to the desktop.
- Oversized or shape-violating Yjs updates are **dropped** (generous default caps: 512 KiB/update, 4 MiB/doc).
- No API/protocol changes; no new runtime dependency beyond `zod`.

## Test coverage

63 new tests, all green (161 total across the three test packages):

| Suite | Tests | Runner |
|---|---|---|
| `schemas.test.ts` | 28 | vitest |
| `stateDoc.test.ts` | 8 | vitest |
| `config.test.ts` | 9 | vitest |
| `data.test.ts` | 5 | vitest (real temp files + local HTTP server) |
| `stateDocGuard.test.ts` | 8 | node:test |
| `wsValidation.test.ts` | 5 | node:test (live server + real WS clients) |

The security-critical end-to-end test (a malicious binary Yjs update from a client must not be broadcast to the uplink) was red-green verified against the un-wired code. Typecheck and the main-process / control-client builds pass.

## Scope

This PR covers the four inputs named in the issue title (config, stream data, WS control messages, Yjs updates) — the security-critical trust boundaries — completely. The remaining client-side `as`-cast TODOs in `control-ui` / `control-client` / the desktop renderer are intentionally out of scope: they sit behind the now-validated server boundary (e.g. the invite-form cast is redundant now that the server validates `create-invite`), and those packages have no test harness to support TDD. They can follow separately.

Closes #11

---

> [!NOTE]
> **High Risk**
> Changes sit on security-critical trust boundaries (remote control commands, shared Yjs state, and startup config); invalid config now fails startup and bad client doc edits disconnect operators instead of silently diverging.
> 
> **Overview**
> Adds **Zod** as the shared runtime validation layer for every major untrusted input path: startup config, polled/file stream lists, WebSocket control messages, and client-originated Yjs state-doc updates.
> 
> In **`streamwall-shared`**, new schemas define bounded control commands (discriminated union), stream entry shape (`parseStreamList` drops bad rows and strips forged `_id` / `_dataSource`), uplink `state` message structure, and **`isValidStateDocShape`** for the narrow `views` → `streamId` Yjs layout.
> 
> The **control server** now **`safeParse`s** desktop `state` payloads before building `StateWrapper`, validates every client JSON command **before** `roleCan` / uplink relay (unknown or out-of-range commands get `invalid message` and are not forwarded), applies **`applyValidatedDocUpdate`** to operator binary edits (per-update byte cap, per-update doc growth cap, probe-then-apply so rejects never broadcast), and **closes** the client socket on a rejected doc update to force resync. Trusted desktop uplink binary frames still apply directly without the guard.
> 
> The **desktop** parses TOML via **`parseConfigToml`** / **`validateConfig`** (grid bounds aligned with remote commands), exits cleanly on **`ConfigError`**, and routes JSON/TOML stream sources through **`parseStreamList`**; file watching now closes the chokidar watcher when the async generator ends.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7cb7d2635d19759870bff7649e240453fc8756c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


NilsR0711:
## Addressed all three Bugbot findings

All three were legitimate — thanks. Fixed in `5ed5f41` and `1743162`:

1. **Command grid caps mismatch config** (Medium) — `config.ts` now bounds `grid.cols`/`grid.rows` to the same `GRID_MIN..GRID_MAX` range as the WS command schema (the initial grid is never clamped), so an oversized wall is rejected at startup with a clear message instead of producing cells the remote can't target. New tests cover the max/min bounds.
2. **Uplink sync hits update cap** (Medium) — the trusted desktop uplink now applies its binary frames directly and bypasses the per-update size cap (which is meant for untrusted clients); its full-state snapshot on connect can legitimately exceed the cap. The size/shape guard remains on the client path. Added a regression test: a large uplink update syncs to clients even under a tiny cap.
3. **Help blocked by config validation** (Low) — `validateConfig` is now skipped when `--help` is requested, so an invalid config can't block the help text.

Full suite green (165 tests across the three packages); typecheck and builds pass.

NilsR0711:
## Second Bugbot pass — 2 already fixed, 2 addressed in `eeb1dbb`

- **Help blocked by config validation** & **Command grid caps mismatch config** — these re-posts point at the pre-fix commit (`3a86d5b`); both were already fixed in `5ed5f41` (help validation is now skipped for `--help`; `config.ts` bounds the grid to `GRID_MIN..GRID_MAX`). No further change needed.

- **Client updates blocked after uplink snapshot** (Medium) — valid and a consequence of the earlier uplink fix. The guard capped the doc's *absolute* size, but that size is set by the trusted uplink's snapshot (which bypasses the guard), so once the doc grew large every later operator edit was dropped. The guard now bounds how much a single update **grows** the doc, so small edits always apply regardless of absolute size.

- **Doc size limit race** (Medium) — the guard is synchronous (no `await` between measuring the probe and applying to the live doc), so concurrent connection handlers can't actually interleave in Node's single-threaded loop; a strict reading is a false positive. Regardless, replacing the absolute cap with a per-update growth bound removes any shared cap to accumulate against, so the concern is moot either way. Documented the synchronous invariant in the guard.

Full suite green (166 tests); typecheck and builds pass.

NilsR0711:
## Third Bugbot pass — 3 new findings fixed in `3f2359b`

All three were legitimate consistency gaps in the validation and are now closed:

- **Invalid invite role silent** (Medium) — `create-invite` now restricts `role` to the known roles (`z.enum(validRoles)`), so an invalid role is rejected up front with an `invalid message` response instead of throwing later in `auth.createToken` with no client reply.
- **Yjs view keys unbounded** (Medium) — `isValidStateDocShape` now bounds view keys to `MAX_VIEW_IDX`, matching the control commands, so an operator can't accumulate phantom cells beyond the grid across many small updates.
- **Empty custom stream link** (Low) — `localStreamDataSchema.link` now requires `min(1)`, consistent with file/URL stream input.

The four other items in this pass are stale re-posts — their locations point at pre-fix commits (`3a86d5b`, `1743162`); those were fixed in `5ed5f41` / `eeb1dbb`. Bugbot's runs are reporting "skipping", which appears to re-attach earlier findings to the head commit without re-analyzing.

Full suite green (172 tests); typecheck and builds pass.

NilsR0711:
## Fourth pass — 1 new finding fixed in `b7cb7d2`

**Rejected Yjs updates desync client** (Medium) — valid. A client applies its edit locally before sending it, so a server-side rejection (only logged before) left the operator UI out of sync with the shared doc. Resending the canonical state can't reliably undo the local edit (CRDT last-write-wins), so the socket is now closed on rejection — mirroring the rate-limit path — forcing a clean reconnect that rebuilds the doc from the server's state. Only malicious/buggy clients are affected; legitimate ones only send valid grid edits.

The other seven items are stale re-posts (locations at pre-fix commits `3a86d5b` / `1743162` / `eeb1dbb`), already fixed. Full suite green (172 tests); typecheck and builds pass.

Issue #79, opened 2026-07-02
Feature: add undo/redo for grid edits (Y.UndoManager)


NilsR0711:
**Type: feature · Priority: medium**

No undo/redo exists. With drag-move, resize handles, and destructive grid-size remap (`remapGridAssignments` *drops* out-of-grid assignments, `geometry.ts:135-140`), a mis-drag or accidental grid shrink loses layout state irrecoverably.

**Desired** — Ctrl+Z / Ctrl+Shift+Z on the shared views doc via `Y.UndoManager` (cheap given Yjs); especially guard the grid-shrink case.

_Source: feature/DX analysis_

Issue #51, opened 2026-07-02
Test the set-grid-size Yjs remap pipeline (extract handler for testability)


NilsR0711:
**Priority: medium**

`remapGridAssignments` is tested but the integration in `packages/streamwall/src/main/index.ts:428-462` (read `viewsState`, transact delete-all + rebuild, keep `streamWindowConfig` in sync) is not.

**Scope** — extract `handleSetGridSize(stateDoc, viewsState, config, msg)` into a testable module; then against a bare `Y.Doc`: grow cols, shrink rows, shrink both (dropped assignments), clamping of NaN/out-of-range, resulting map covers exactly `cols*rows` keys, single-update atomicity. Refactor-first issue.

_Source: test analysis_

Issue #42, opened 2026-07-02
Bug: debounce eager grid-input typing to avoid tearing down views mid-type


NilsR0711:
**Severity: low**

**Evidence** — `packages/streamwall-control-ui/src/index.tsx:720-729` + `:1750` (`isEager`): typing a stream ID character-by-character writes `''` to the Yjs doc for every partial non-match ("w", "wo"), tearing down the view before the final match re-creates it → visible flicker/reload on the wall while typing.

**Fix** — debounce eager commits, or only clear the cell once the input is empty rather than on every non-matching partial.

_Source: control-UI review_

PR #133, opened 2026-07-13
fix: prevent grid-size inputs from collapsing the wall (NaN → 1×N)

### LINKED ISSUE #28: Bug: grid-size number inputs collapse the wall when cleared (NaN → 1×N, destructive)
**Severity: high (destructive data loss)**

The inputs fire `set-grid-size` on every keystroke; clearing the field yields `NaN` which clamps to 1.

**Evidence**
- `packages/streamwall-control-ui/src/index.tsx:1682-1700` (fires `onSetGridSize(ev.currentTarget.valueAsNumber, …)` on every input).
- `packages/streamwall-shared/src/geometry.ts:127-132` (`clampGridDimension(NaN)` → `GRID_MIN` = 1).
- `packages/streamwall/src/main/index.ts:428-461` (`remapGridAssignments` permanently drops out-of-grid assignments).

**Failure scenario**
On a 4×4 wall, Backspace in "Spalten" to retype "6" → empty field → `NaN` → server rebuilds a 1×4 grid → 12 of 16 placements irrevocably deleted before the user finishes typing. No confirmation, no undo.

**Fix**
Ignore `NaN`/out-of-range values in `GridSizeControls`; commit on blur/Enter, not on every input; confirm when the new grid would drop non-empty cells. (See also the undo/redo feature request.)

_Source: control-UI review_

NilsR0711:
## Problem

The columns/rows number inputs in the control UI fired `set-grid-size` on **every keystroke**. Clearing a field yields `valueAsNumber === NaN`, which `clampGridDimension` maps to `GRID_MIN` (1). The server then rebuilds the grid at `1×N` and `remapGridAssignments` permanently drops every placement whose `(x, y)` no longer fits — with no confirmation and no undo.

**Failure scenario:** on a 4×4 wall, pressing Backspace in "Spalten" to retype "6" leaves the field empty for an instant → `NaN` → server rebuilds a 1×4 grid → 12 of 16 placements are irrevocably deleted before the user finishes typing.

## Solution

Three layered guards, matching the fix outlined in the issue:

1. **Ignore invalid input.** New pure helper `parseGridDimensionInput` returns `null` for empty, non-numeric, `NaN`, and out-of-range input instead of silently clamping. A transient keystroke can no longer force a dimension.
2. **Commit on blur/Enter, not on every input.** `GridSizeControls` now holds a local draft per field (`onInput` updates only the draft) and commits on blur or Enter. Invalid drafts revert to the authoritative value; valid ones are sent. Presets still apply immediately.
3. **Confirm destructive shrink.** New pure helper `gridWouldDropAssignments` detects when a resize would drop a non-empty cell (mirroring `remapGridAssignments`' `(x, y)` rule). `handleSetGridSize` warns via `window.confirm` before sending — this also covers shrinking presets.

## Architecture

The decision logic lives as two pure, framework-agnostic functions in `streamwall-shared/geometry.ts`, next to the existing `clampGridDimension`/`remapGridAssignments` grid helpers, so it is unit-testable and reused by the UI. The control-ui component stays a thin wrapper.

## Testing

- **TDD**: 11 new vitest cases in `geometry.test.ts` (red → green) covering `parseGridDimensionInput` (valid, bounds, empty/whitespace, non-numeric/NaN, below/above range, fractional rounding) and `gridWouldDropAssignments` (drop, survive, empty-cell, grow).
- Full quality gate green locally: `prettier --check`, `eslint .`, `tsc --noEmit` (all 5 packages), `npm test` (85 + 83 + 61 pass), `npm -w streamwall-control-client run build`.
- The `streamwall-control-ui` package has no component-test harness (only `typecheck`); the React wiring (draft state, blur commit, confirm) is covered by the type checker, while the extracted decision logic is fully unit-tested. Adding a Preact test harness would be out of scope for this fix.

## Risks / breaking changes

None. Grid resizing behaves identically for valid input; only the destructive intermediate-state and silent-data-loss paths change. `window.confirm` is available in both the Electron control window and the standalone web client.

Closes #28

### DIFF
--- a/packages/streamwall-control-ui/src/index.tsx
+++ b/packages/streamwall-control-ui/src/index.tsx
@@ -38,10 +38,12 @@ import {
   type ControlCommand,
   GRID_MAX,
   GRID_MIN,
+  gridWouldDropAssignments,
   idColor,
   idxInBox,
   inviteLink,
   type LocalStreamData,
+  parseGridDimensionInput,
   roleCan,
   type StreamData,
   type StreamDelayStatus,
@@ -760,13 +762,31 @@ export function ControlUI({
 
   const handleSetGridSize = useCallback(
     (nextCols: number, nextRows: number) => {
+      const targetCols = clampGridDimension(nextCols)
+      const targetRows = clampGridDimension(nextRows)
+      // Shrinking the grid permanently drops any placement whose (x, y) no
+      // longer fits. Warn before that happens so it is never silent.
+      if (cols != null && sharedState) {
+        const assignments = new Map<number, string | undefined>()
+        for (const [idx, view] of Object.entries(sharedState.views)) {
+          assignments.set(Number(idx), view.streamId)
+        }
+        if (
+          gridWouldDropAssignments(cols, targetCols, targetRows, assignments) &&
+          !window.confirm(
+            'Das neue Raster ist kleiner und entfernt belegte Kacheln dauerhaft. Fortfahren?',
+          )
+        ) {
+          return
+        }
+      }
       send({
         type: 'set-grid-size',
-        cols: clampGridDimension(nextCols),
-        rows: clampGridDimension(nextRows),
+        cols: targetCols,
+        rows: targetRows,
       })
     },
-    [send],
+    [send, cols, sharedState],
   )
 
   const handleSetBackgroundListening = useCallback(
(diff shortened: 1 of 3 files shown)


NilsR0711:
All CI checks are green (lint, format, typecheck, tests on ubuntu/macOS/windows, build, CodeQL, dependency review). No automated AI review (Cursor BugBot) posted within the ~15-minute window, so per the resolve-issue workflow the AI-review step is treated as satisfied. A manual self-review of the diff found no issues: the destructive NaN→collapse path is structurally removed (inputs only update a local draft on `onInput`; commits happen on blur/Enter and ignore empty/NaN/out-of-range values; every shrinking resize — typed or preset — is gated by a confirmation).

NilsR0711:
Related issues (already tracked, no duplicates created):
- #79 (undo/redo for grid edits) — the confirmation guard added here is the interim safety net the issue refers to; a full undo/redo remains the longer-term fix.
- #42 (eager stream-id typing tears down views) is a **distinct** bug in the per-cell `GridInput`/`handleChangeSpace` path, not the grid-size inputs, and is intentionally left untouched by this PR.
- #51/#54 (set-grid-size remap + control-ui interaction test debt) remain open; this PR extracts and unit-tests the input-guard/drop-detection logic in `streamwall-shared` as a step in that direction.

PR #134, opened 2026-07-13
fix(control-ui): place newly-appeared streams instead of clearing the cell

### LINKED ISSUE #30: Bug: stale handleSetView inside handleClickId clears the cell when placing a newly-appeared stream
**Severity: high**

`handleClickId` omits `handleSetView` from its deps, so it can invoke an old `handleSetView` whose captured `streams` lacks the new stream.

**Evidence** — `packages/streamwall-control-ui/src/index.tsx:823-849` (deps `[cols,rows,sharedState,focusedInputIdx]`), `:720-729` (`handleSetView` closes over `streams`).

**Failure scenario** — a new stream appears in the sidebar (streams updated, shared views unchanged); clicking its ID runs a stale `handleSetView` → `streams.find` misses → writes `''` into the Yjs doc, i.e. the click leaves the cell empty instead of placing the stream, with no feedback. Also `sharedState.views[i].streamId` (`:841`) lacks optional chaining and can throw `TypeError` while the JSON config channel and the Yjs views channel are momentarily inconsistent after a grid enlargement.

**Fix** — add `handleSetView` (and `streams`) to the dep array; use `sharedState.views[i]?.streamId`.

_Source: control-UI review_

NilsR0711:
## Problem

Clicking a stream id in the control UI to place a newly appeared stream could silently leave the grid cell empty instead of placing the stream — and, in a narrow timing window, could throw.

Two root causes in `packages/streamwall-control-ui/src/index.tsx`:

1. **Stale callback (the visible bug).** `handleClickId` omitted `handleSetView` from its `useCallback` dependency array (`[cols, rows, sharedState, focusedInputIdx]`). When a new stream appeared, `streams` changed but the shared views did not, so `handleClickId` kept an old `handleSetView` whose captured `streams` lacked the new entry. `streams.find` missed → `''` was written into the Yjs doc → the click left the cell empty with no feedback.
2. **Missing optional chaining.** The available-cell scan read `sharedState.views[i].streamId` directly. Right after a grid enlargement the JSON-config channel can already report the larger grid while the Yjs `views` map still lacks entries for the new cells, so `views[i]` can be `undefined` and `.streamId` throws a `TypeError`.

## Solution

- Extracted the two decisions into pure, unit-tested helpers in `viewPlacement.ts`:
  - `resolveTargetViewIdx` — picks the focused cell, or the first empty cell, reading the views map defensively (`views[idx]?.streamId`), so a missing slot counts as empty instead of throwing.
  - `resolveWriteStreamId` — returns the clicked id when the stream is known, or `''` to clear the cell when it is not.
- Wired both into `ControlUI` and added `handleSetView` to the `handleClickId` dependency array so the callback always closes over the **current** `streams`.

## Architecture notes

The control-ui package had no test setup. The interaction logic lives inside a ~1800-line component, so the decision logic was extracted into small pure functions that are unit-testable without a full render harness. The dependency-array fix itself is a React-hooks correctness change (verified by typecheck and review); the extracted decision logic it depends on is covered by tests.

## Testing

- Added a **vitest** setup to the previously untested `streamwall-control-ui` package (matching the convention already used by `streamwall` and `streamwall-shared`); the root `test` script picks it up via `--workspaces --if-present`.
- 8 new unit tests covering: focused-input targeting, first-empty-cell selection, full grid, **missing view slot after grid enlargement (no throw)**, empty-`streamId` cells, and the write-value regression (a freshly appeared stream is placed, not cleared).
- Full quality gate green locally: `format:check`, `lint`, `typecheck` (all packages), full test suite (all workspaces), and the control-client Vite build (which bundles control-ui).

## Risks / breaking changes

None. Behavior is unchanged except for the two fixed defects; no public API changes.

Closes #30

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15356,7 +15356,8 @@
         "@preact/preset-vite": "^2.10.5",
         "@types/lodash-es": "^4.17.12",
         "@types/luxon": "^3.7.2",
-        "typescript": "^6.0.3"
+        "typescript": "^6.0.3",
+        "vitest": "^4.1.10"
       }
     },
     "packages/streamwall-control-ui/node_modules/color": {
--- a/packages/streamwall-control-ui/package.json
+++ b/packages/streamwall-control-ui/package.json
@@ -26,9 +26,12 @@
     "@preact/preset-vite": "^2.10.5",
     "@types/lodash-es": "^4.17.12",
     "@types/luxon": "^3.7.2",
-    "typescript": "^6.0.3"
+    "typescript": "^6.0.3",
+    "vitest": "^4.1.10"
   },
   "scripts": {
-    "typecheck": "tsc --noEmit -p tsconfig.json"
+    "typecheck": "tsc --noEmit -p tsconfig.json",
+    "test": "vitest run",
+    "test:watch": "vitest"
   }
 }
--- a/packages/streamwall-control-ui/src/index.tsx
+++ b/packages/streamwall-control-ui/src/index.tsx
@@ -54,6 +54,7 @@ import { createGlobalStyle, styled } from 'styled-components'
 import { matchesState } from 'xstate'
 import * as Y from 'yjs'
 import './index.css'
+import { resolveTargetViewIdx, resolveWriteStreamId } from './viewPlacement.ts'
 
 export interface ViewInfo {
   state: ViewState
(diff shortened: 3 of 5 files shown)


NilsR0711:
### Review status

All CI checks are green (Test on ubuntu/macos/windows, Typecheck, Lint, Format, Build, CodeQL, Dependency review). Cursor BugBot did not post a review within the ~15-minute window (it also didn't trigger on the recent PRs #118 and #132), so the AI-review step is treated as satisfied per the workflow's rate-limit/unavailable clause.

Verification performed locally, all green:
- `npm run format:check`, `npm run lint`, `npm run typecheck` (all 5 packages)
- Full test suite across all workspaces, including the 8 new `viewPlacement` unit tests
- `npm run build -w packages/streamwall-control-client` (bundles control-ui via Vite)

NilsR0711:
Follow-up filed: #138 — enable `eslint-plugin-react-hooks` (exhaustive-deps) for the Preact packages, which would have caught this stale-closure bug at lint time.

PR #136, opened 2026-07-13
fix(control-ui): commit LazyChangeInput edits on Enter (stale closure)

### LINKED ISSUE #29: Bug: stale-closure Enter handler discards edits in LazyChangeInput
**Severity: high**

`handleKeyDown` captures the first-render `handleBlur` (whose `editingValue` is `undefined`).

**Evidence** — `packages/streamwall-control-ui/src/index.tsx:1539-1546` (`handleKeyDown` deps `[]`), `:1532-1537` (`handleBlur` deps `[editingValue]`, missing `onChange`/`isEager`).

**Failure scenario** — pressing Enter in any non-eager `LazyChangeInput` (e.g. the custom-stream label at `:1933`) never calls `onChange` and resets `editingValue` to `undefined` → the typed text reverts and the edit is silently lost; the subsequent real blur also commits nothing.

**Fix** — give `handleKeyDown` dep `[handleBlur]` (or inline the commit), and add `onChange`/`isEager` to `handleBlur`'s deps.

_Source: control-UI review_

NilsR0711:
## Problem

Pressing **Enter** in any non-eager `LazyChangeInput` silently discarded the typed text (issue #29, severity high). The most visible case is the custom-stream **Label** field.

`handleKeyDown` was created with an empty dependency array, so it captured the **first-render** `handleBlur`, whose `editingValue` closure was still `undefined`. Pressing Enter therefore:

1. ran a stale `handleBlur` → the `editingValue !== undefined` guard was false → `onChange` was never called, and
2. reset `editingValue` to `undefined`.

The subsequent *real* blur then also committed nothing, because `editingValue` had already been cleared. The edit was lost with no feedback. `handleBlur` additionally omitted `isEager`/`onChange` from its own deps, so it could read stale props.

## Solution

- `handleKeyDown` now depends on `[handleBlur]`, so Enter always runs the **current** commit logic instead of a frozen first-render closure.
- `handleBlur`'s dependency array is completed to `[editingValue, isEager, onChange]`, so it never reads stale props (this is also what `react-hooks/exhaustive-deps` requires).

Both changes are the exact-dependency fix; no logic was rewritten.

## Architecture decisions

- Extracted the previously-inline `LazyChangeInput` out of the ~2000-line `index.tsx` into its own module (`src/LazyChangeInput.tsx`) so it can be rendered and tested in isolation, without importing the whole control UI and its `@fontsource` side effects. Pure move, no behavior change (separate `refactor` commit).
- Added a test harness to `streamwall-control-ui`, which previously had **no tests** — `vitest` + `happy-dom`, matching the `vitest` convention already used by `streamwall` and `streamwall-shared`. `preact/compat` is loaded in the setup so `onChange`/`onBlur` behave exactly as they do in the app (React-style: `onChange` on input, `onBlur` on `focusout`).

## Testing performed

New regression tests in `src/LazyChangeInput.test.tsx` (all green; each was confirmed to fail against the unpatched component):

- Enter commits the edit in a non-eager input.
- Enter followed by a real blur commits exactly once (no lost edit, no double-fire).
- Blur commits the edit (baseline guard).
- Commits to the **latest** `onChange` after a parent re-render (covers the `handleBlur` deps fix).
- Eager mode fires `onChange` on every keystroke and Enter does not double-commit.

Full quality gate green locally and matches CI: Prettier, ESLint, `tsc` across all 5 packages, full test suite (223 tests total, incl. the 5 new), and the control-client build.

## Risks / breaking changes

None. Behavior-only fix scoped to `LazyChangeInput`; the fix restores the intended commit-on-Enter behavior. The new dependency arrays cannot loop (they reference existing props/state/callbacks).

Closes #29

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3747,9 +3747,6 @@
       "cpu": [
         "arm64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3765,9 +3762,6 @@
       "integrity": "sha512-lZVym0PuHE1KZ22gmFTC15lAkrg9iTszR617oYRB/iPY1A56ywoJzVKOJBKaot5RiikCObmur6pogpse3gRcng==",
       "cpu": [
         "arm64"
-      ],
-      "libc": [
-        "musl"
       ],
       "license": "MIT",
       "optional": true,
@@ -3785,9 +3779,6 @@
       "cpu": [
         "ppc64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -3803,9 +3794,6 @@
       "integrity": "sha512-0WIRnL1Uw4BvTZRLQt+PVgo6ZKTJadlC2btP+/EOXv2f/DWbY0rEgl+y834mIVwP1FkTlWVTrGGJXf12lru7EQ==",
       "cpu": [
         "s390x"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "MIT",
       "optional": true,
@@ -3823,9 +3811,6 @@
       "cpu": [
         "x64"
       ],
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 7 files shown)


NilsR0711:
### Status

All CI checks are green (Lint, Format, Typecheck across all 5 packages, Test on ubuntu/macos/windows, Build, CodeQL, Dependency review).

No AI reviewer (Cursor BugBot / CodeRabbit) is currently active on this repository — no bot check-run or review was produced within the review window, consistent with recent PRs (#126, #129) which also received none. Treating the automated-review requirement as satisfied (provably unavailable) per the standard workflow; happy to address any human review feedback.

PR #144, opened 2026-07-13
test(control-server): complete auth unit-test coverage (#46)

### LINKED ISSUE #46: Test the control-server auth: scrypt hashing, timing-safe validation, base62 encoding
**Priority: high (security-critical, currently untested)**

`Auth.validateToken` (`packages/streamwall-control-server/src/auth.ts`) is the sole gate for remote control.

**Scope (pure unit tests, no mocks)** — `rand62` length/alphabet; `uniqueRand62` collision regeneration (seed via mocked `randomBytes`); `createToken`→`validateToken` success; wrong secret → null; unknown tokenId → null; **wrong-length stored hash does not throw** (documents the `timingSafeEqual` length bug — will initially fail); `createToken` rejects invalid roles; `deleteToken`; constructor/`getStoredData` round-trip (persistence format stability).

Add `vitest` devDep + `test` script to the server package.

_Source: test analysis_

NilsR0711:
## Problem

Issue #46 asks for pure unit tests around the control-server auth path — `Auth.validateToken`, scrypt hashing, timing-safe validation, and base62 encoding — the sole gate for remote control.

Since the issue was filed, a large part of that suite already landed (via the security hardening in #118): `createToken`→`validateToken` success, wrong-secret/unknown-id rejection, the wrong-length-hash `timingSafeEqual` fix, per-token salt, legacy shared-salt fallback, storage round-trip, and the revocation-during-validation race are all covered. This PR closes the remaining gaps enumerated in the issue.

## Solution

Added five focused tests to `packages/streamwall-control-server/src/auth.test.ts`:

- **`rand62`** emits only base62 alphabet characters and stays within the theoretical length bound for the given byte count.
- **`uniqueRand62`** regenerates on collision until it finds a value absent from the map — driven deterministically through a minimal `Map` stand-in whose `has()` reports the first two candidates as taken (mocking Node's `randomBytes` named import is not interceptable under ESM, so the loop is exercised via the map instead).
- **`createToken`** rejects an unknown role and stores nothing.
- **`deleteToken`** revokes a token so it no longer validates, and is a no-op for unknown ids.
- **`validateToken`** fails closed (returns `null`, never throws) when the stored hash is not valid base62 — covering the previously untested `base62.decode` `catch` branch.

## Notes / decisions

- **No production changes.** The `timingSafeEqual` length bug referenced in the issue ("will initially fail") was already fixed in `auth.ts` (explicit length guard before the comparison), so the corresponding case passes today. Scope is test-only.
- **Runner:** the issue suggested adding `vitest` + a `test` script, but the server package already ships a native `node:test` runner and an existing `auth.test.ts`; the new tests follow that established pattern rather than introducing a second framework.

## Testing

- `npm -w packages/streamwall-control-server test` → 66 passing (was 61).
- Repo-wide `format:check`, `lint`, and workspace `typecheck` all clean.

Closes #46

### DIFF
--- a/packages/streamwall-control-server/src/auth.test.ts
+++ b/packages/streamwall-control-server/src/auth.test.ts
@@ -3,7 +3,7 @@ import assert from 'node:assert/strict'
 import { scrypt as scryptCb } from 'node:crypto'
 import { test } from 'node:test'
 import { promisify } from 'node:util'
-import { type AuthToken, Auth, rand62 } from './auth.ts'
+import { type AuthToken, Auth, rand62, uniqueRand62 } from './auth.ts'
 
 const scrypt = promisify(scryptCb)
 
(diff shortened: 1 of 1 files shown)


NilsR0711:
AI review (Cursor BugBot) did not post a review within ~13 minutes of the last push and is not registered as a required status check on this PR — treating the AI-review step as unavailable. All GitHub Actions checks (static checks, typecheck, tests on ubuntu/macos/windows, CodeQL, CI OK) are green, and the change is test-only (no production code touched).

PR #146, opened 2026-07-14
test: cover roleCan matrix and StateWrapper.view role scoping

### LINKED ISSUE #47: Test role-based access control (roleCan matrix) and StateWrapper.view role scoping
**Priority: high**

`packages/streamwall-shared/src/roles.ts` is the authorization matrix; `StateWrapper.view` (`control-server/src/auth.ts`) decides that only admins see the token list.

**Scope** — table-driven test over the full role × action matrix; explicit test that unlisted message types (e.g. `set-grid-size`) are denied for operator/monitor; `StateWrapper.view('operator')` has no `auth` key while `view('admin')` does; `inviteLink` formatting.

_Source: test analysis_

NilsR0711:
## Problem

Issue #47: the authorization surface was only thinly tested. `roleCan`
(`packages/streamwall-shared/src/roles.ts`) had ad-hoc coverage of a single
action (`set-grid-size`), and `StateWrapper.view`
(`packages/streamwall-control-server/src/auth.ts`) — which decides that only
admins see the token list — had **no** tests at all. A regression that widened
a role's permissions, or that leaked the `auth` token list to a non-admin view,
could ship silently.

## Solution

Test-only change covering the two authorization primitives the issue names.

**`roleCan` — table-driven role × action matrix** (streamwall-shared):
- Full matrix over every role (`local`, `admin`, `operator`, `monitor`) × every
  action, plus the unauthenticated `null` role. The expected allow-sets are
  re-declared independently of the implementation and pinned with `satisfies
  readonly StreamwallAction[]`, so removing an action from `roles.ts` breaks the
  test loudly.
- Explicitly asserts admin-only actions (`dev-tools`, `browse`, `create-invite`,
  `delete-token`) are denied to operator/monitor, and operator-only actions are
  denied to monitor.
- A dedicated default-deny block for an unlisted message type (a future/unknown
  command) falling through to a deny for non-privileged roles. Note: the issue's
  original `set-grid-size` example is now a *listed* operator action, so the
  matrix asserts operator ✓ / monitor ✗ for it and the default-deny block uses a
  genuinely unlisted action.
- Folds the old standalone `set-grid-size` cases into the matrix (subsumed);
  keeps and extends the `inviteLink` formatting tests.

**`StateWrapper.view` — role scoping** (control-server):
- Only the `admin` view carries the `auth` token list, present as an actual key.
- `operator`, `monitor`, and even all-powerful `local` views omit the `auth`
  key **entirely** (not merely `undefined`) — token-list visibility is
  admin-exclusive and independent of `roleCan`.
- The `info` getter yields an unprivileged monitor view with no `auth`.
- Pins `identity.role` re-stamping without mutating the wrapped source, and
  pass-through of non-sensitive fields.

## Architecture decisions

- Expected allow-sets are re-listed in the test rather than imported from the
  (unexported) action arrays, so the test encodes the *intended* contract
  instead of tautologically re-deriving it from the code under test.
- Tests live beside existing suites and use each package's established runner
  (vitest for streamwall-shared, `node:test` for control-server).

## Testing

- `npm -w packages/streamwall-shared test` → 158 passed (was 83).
- `npm -w packages/streamwall-control-server test` → 73 passed.
- Verified the new tests are non-vacuous by mutation: making `monitor`
  omnipotent in `roleCan` fails 14 matrix cases; leaking `auth` to all roles in
  `view` fails 4 StateWrapper cases. Source reverted after.
- `npm run typecheck`, `npm run lint`, and `prettier --check` on the changed
  files are clean.

## Risks / breaking changes

None — additive, test-only. No production code changed.

Closes #47

### DIFF
--- a/packages/streamwall-control-server/src/auth.test.ts
+++ b/packages/streamwall-control-server/src/auth.test.ts
@@ -3,7 +3,14 @@ import assert from 'node:assert/strict'
 import { scrypt as scryptCb } from 'node:crypto'
 import { test } from 'node:test'
 import { promisify } from 'node:util'
-import { type AuthToken, Auth, rand62, uniqueRand62 } from './auth.ts'
+import type { StreamwallRole, StreamwallState } from 'streamwall-shared'
+import {
+  type AuthToken,
+  Auth,
+  rand62,
+  StateWrapper,
+  uniqueRand62,
+} from './auth.ts'
 
 const scrypt = promisify(scryptCb)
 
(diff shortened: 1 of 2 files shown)


NilsR0711:
### Status

- ✅ All CI checks green (Static checks, Tests on ubuntu/macos/windows, Build, CodeQL, CI OK gate, Conventional Commits title).
- ℹ️ **AI review (Cursor BugBot):** no BugBot check-run or review was posted for this PR after ~10 minutes — it did not trigger automatically here. Per the resolve-issue workflow, the AI-review requirement is treated as satisfied since everything else is green. Happy to address findings if BugBot runs later.

### Follow-up

- #147 — DX: `format:check` lints git-ignored build output because there is no `.prettierignore` (noticed while running the quality gate for this change; out of scope here).

Issue #145, opened 2026-07-14
Modernize tsconfig module resolution (drop baseUrl, moduleResolution: bundler) and remove ignoreDeprecations


NilsR0711:
**Priority: low**

Follow-up split off from #57. The core of #57 (typecheck scripts everywhere, the two live type errors, and CI wiring) is done. What remains is the deliberately-deferred TypeScript module-resolution modernization.

**Current state** — `packages/streamwall/tsconfig.json` and `packages/streamwall-control-ui/tsconfig.json` still use the TS6-deprecated combo `baseUrl: "."` + `moduleResolution: "node"`. The resulting TS5101/TS5107 deprecation errors are silenced with `ignoreDeprecations: "6.0"`, with an inline comment deferring the cleanup:

```jsonc
// baseUrl + moduleResolution:node are flagged as deprecated under TS 6 but
// are still required by the Electron/CommonJS build; silence until the
// module-resolution modernization is done separately.
"ignoreDeprecations": "6.0",
```

**Fix** — drop `baseUrl` and switch these `tsc --noEmit` configs to `moduleResolution: "bundler"` (which also requires `module: "esnext"`/`"preserve"` — safe here because tsc only type-checks; Vite/Electron Forge/esbuild does the actual build). `paths` continues to work without `baseUrl` under `bundler`. Then remove the `ignoreDeprecations` escape hatch and re-run `npm run typecheck` across all packages.

**Notes**
- Interacts with #58 (the `react`→`preact/compat` `paths` location); the two tsconfig cleanups should probably land together.
- Watch for new type errors surfacing once `bundler` resolution is active.

_Source: follow-up from #57 resolution_

PR #151, opened 2026-07-14
fix(tsconfig): resolve react→preact/compat via hoisted node_modules

### LINKED ISSUE #58: Fix broken react→preact/compat tsconfig paths
**Priority: high**

Three tsconfigs map `react` to a package-local path that doesn't exist (npm hoists to the root), silently degrading react-typed libs to broken typings.

**Evidence** — `packages/streamwall-control-client/tsconfig.json` (`"react": ["./node_modules/preact/compat/"]`), same in `packages/streamwall/tsconfig.json` and `packages/streamwall-control-ui/tsconfig.json`. Verified: only `<root>/node_modules/preact/compat` exists. Likely contributes to the `IconBaseProps`/`className` error.

**Fix** — point paths at the hoisted location (`"react": ["../../node_modules/preact/compat"]`, same for `react-dom`), or set them once in a shared base tsconfig; re-run typecheck.

_Source: CI/build analysis_

NilsR0711:
## Problem

Three package tsconfigs mapped `react`/`react-dom` at a **package-local** path that npm never creates — workspace deps hoist to the repo root:

```jsonc
"react": ["./node_modules/preact/compat/"]      // packages/<pkg>/node_modules/... → does not exist
"react-dom": ["./node_modules/preact/compat/"]
```

Because the substitution pointed at a non-existent directory, `tsc` fell back to standard resolution and resolved `react` to the **untyped** real `react@19` package (`index.js`, no `.d.ts`). Verified via `--traceResolution`:

```
Module name 'react' was successfully resolved to '…/node_modules/react/index.js' … @19.0.0
```

The typecheck therefore *passed falsely*: every react-typed lib (react-icons, styled-components, react-hotkeys-hook) silently degraded to `any`. This is the "silently degrading react-typed libs to broken typings" the issue describes, and the root of the earlier `IconBaseProps`/`className` workarounds.

## Solution

Point the paths at the hoisted location in all three tsconfigs (control-client, streamwall, control-ui):

```jsonc
"react": ["../../node_modules/preact/compat"]
"react-dom": ["../../node_modules/preact/compat"]
```

After the fix `react` resolves correctly in every package:

```
Module name 'react' was successfully resolved to '…/node_modules/preact/compat/src/index.d.ts' … preact-compat@4.0.0
```

## Consequence: honest typecheck surfaced masked errors

Correcting the aliasing turned the false-green typecheck red (72 previously-masked styled-components v6 errors). Since the CI typecheck is a required gate, they are fixed in the same PR by giving each styled component its **prop-type generics** — no runtime change:

- `streamwall-control-ui/src/index.tsx`: `Stack`, `StyledDataContainer`, `StyledButton`, `StyledGridPreviewBox`, `StyledGridButtons`, `StyledGridInput`, `StyledGridContainer`, `StyledId` (+ a `ColorInstance` type alias for the `color`-lib instance type).
- `streamwall/src/renderer/overlay.tsx`: `SpaceBorder`, `StreamTitle`, `FilterCover`, `VersionText`.

Two latent correctness bugs the real types exposed are fixed alongside:
- `StyledId onMouseDown`: `null` → `undefined` (matches the `MouseEventHandler | undefined` contract).
- Overlay `map`: added a `pos` null-guard (`context.pos` is `ViewPos | null`); the old code would have thrown on `pos.x` had `pos` been null.

## Architecture note

Prop types are declared to match the existing prop names/usage exactly, so DOM output and component behavior are unchanged — this is a pure type-correctness pass. Transient props already used (`$color`, `$disabled`) are preserved.

## Testing

Full quality gate green locally:
- Prettier `format:check` ✓
- ESLint ✓
- `tsc --noEmit` across all 5 workspaces ✓ (0 errors; was a false 0 before, now honest 0)
- Full test suite ✓ (243+ tests incl. control-ui component tests — confirms no runtime regression)
- `streamwall-control-client` build ✓
- `--traceResolution` confirms `react` → `preact/compat` in all three packages

## Risks / breaking changes

None expected. Type-only change; runtime DOM/behavior identical. No public API change.

Closes #58

### DIFF
--- a/packages/streamwall-control-client/tsconfig.json
+++ b/packages/streamwall-control-client/tsconfig.json
@@ -6,8 +6,8 @@
     "types": ["vite/client"],
     "allowImportingTsExtensions": true,
     "paths": {
-      "react": ["./node_modules/preact/compat/"],
-      "react-dom": ["./node_modules/preact/compat/"]
+      "react": ["../../node_modules/preact/compat"],
+      "react-dom": ["../../node_modules/preact/compat"]
     }
   },
   "include": ["src"]
--- a/packages/streamwall-control-ui/src/index.tsx
+++ b/packages/streamwall-control-ui/src/index.tsx
@@ -50,6 +50,7 @@ import {
   type StreamwallRole,
   type StreamwallState,
   type StreamWindowConfig,
+  type ViewPos,
   type ViewState,
 } from 'streamwall-shared'
 import { createGlobalStyle, styled } from 'styled-components'
@@ -59,6 +60,10 @@ import { isPrimaryButton, resolveMoveTarget } from './gestures'
 import './index.css'
 import { LazyChangeInput } from './LazyChangeInput.tsx'
 import { resolveTargetViewIdx, resolveWriteStreamId } from './viewPlacement.ts'
+
+// `import Color from 'color'` binds only the value; alias the instance type
+// (as returned by the Color factory) for use in styled-component prop types.
+type ColorInstance = ReturnType<typeof Color>
 
 export interface ViewInfo {
   state: ViewState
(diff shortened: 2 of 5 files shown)


NilsR0711:
Follow-up (out-of-scope, tracked separately): #152 — make the custom styled-components props transient to stop DOM prop leakage / React unknown-attribute warnings. The related tsconfig module-resolution modernization is already tracked in #145.

PR #155, opened 2026-07-14
docs: document remote control server setup, roles, and control.endpoint

NilsR0711:
## Problem

The control server is configured only via env vars and was undocumented for
operators wanting to set it up: no guidance on starting it, connecting the
desktop app to it, or the available roles.

## Solution

- Added a "Remote control server" section to the root README covering:
  - `npm run start:server`
  - the uplink endpoint / admin invite link printed on startup
  - connecting the app via `--control.endpoint` or `[control]` `endpoint` in
    the config file
  - the three roles (admin/operator/monitor)
  - a pointer to the control-server package README for env var configuration
- Fixed the `[control]` block in `example.config.toml`: it still documented
  `address`/`hostname`/`port`/`username`/`password` keys that were removed
  when the control server became a separate package — the config schema
  (`packages/streamwall/src/main/config.ts`) now only has `endpoint`. The
  stale block would have led anyone following the example straight into a
  config that's silently ignored.
- Added a "Roles" table to `packages/streamwall-control-server/README.md`
  describing what each role can do.

## Testing

Docs-only change with no testable behavior — skipped TDD per the resolve-issue
skill's stated exception. Verified locally:
- `npm run format:check`
- `npm run lint`
- `npm run typecheck`

## Risks / breaking changes

None — documentation and an example config file only.

Closes #69

PR #153, opened 2026-07-14
docs: rewrite [control] section of example.config.toml for token/endpoint architecture

NilsR0711:
## Problem

`example.config.toml:20-30` documented `[control]` options (`address`, `hostname`, `port`, `username`, `password`) that do not exist anywhere in the codebase. The app only reads `control.endpoint` — a WebSocket URL that includes the auth token (`packages/streamwall/src/main/index.ts:209-211`). Authentication is handled entirely by the separate control server (`packages/streamwall-control-server`), which mints tokens and prints the uplink endpoint to stdout on startup (`packages/streamwall-control-server/src/index.ts:874-891`).

Anyone following the example config would set options the app silently ignores, leaving `control.endpoint` unset and remote control non-functional.

## Solution

Rewrote the `[control]` section to document the real option:

```toml
[control]
# WebSocket URL (including the auth token) for the separate control server
# that operators use to remotely arrange the grid. The control server prints
# this URL to stdout on startup — copy it here verbatim.
#endpoint = "ws://localhost/streamwall/<id>/ws?token=<secret>"
```

## Testing performed

Docs-only change with no testable behavior — verified by reading the actual config-parsing code (`packages/streamwall/src/main/index.ts`) and the control server's bootstrap/logging code (`packages/streamwall-control-server/src/index.ts`) to confirm `control.endpoint` is the only option the app consumes, and that the URL shape/token flow matches what's now documented.

## Risks / breaking changes

None — this only touches an example/template file.

Closes #67

PR #159, opened 2026-07-14
docs: remove dead [cert]/webserver config keys from example.config.toml

### LINKED ISSUE #157: Docs: remove dead [cert]/webserver config keys from example.config.toml
**Type: docs · Priority: low**

While documenting the remote control server for #69, I noticed `example.config.toml` has a `[cert]` section (`production`, `email`, `dir`) referencing SSL certificate auto-provisioning tied to a "webserver" config option:

```toml
[cert]
# SSL certificates (optional)
# If you specify an https:// URL for the "webserver" option, a certificate will be automatically generated and signed by Let's Encrypt.
...
```

There is no `webserver` option and no `cert` block in the current config schema (`packages/streamwall/src/main/config.ts`) or CLI options (`packages/streamwall/src/main/index.ts`) — a `grep` for `cert`/`webserver` outside this example file turns up nothing. This looks like a leftover from before the control server was split into its own package (the same class of staleness already fixed for `[grid]` count in #149 and for `[control]` in #153).

**Fix** — remove the dead `[cert]` block from `example.config.toml` (or replace it with accurate guidance if TLS termination for the control server is meant to be documented elsewhere instead).

_Source: found while resolving #69 (docs: document the control server)_

NilsR0711:
## Problem

`example.config.toml` contains a `[cert]` section (`production`, `email`, `dir`) documenting automatic Let's Encrypt certificate provisioning tied to a `webserver` config option. Neither `webserver` nor `[cert]` exist in the current config schema (`packages/streamwall/src/main/config.ts`) or CLI options (`packages/streamwall/src/main/index.ts`) — a repo-wide search for `cert`/`webserver` outside this file turns up nothing related. This is leftover documentation from before the control server was split into its own package, the same class of staleness already fixed for `[grid]` count in #149 and for `[control]` in #153.

## Solution

Remove the dead `[cert]` block from `example.config.toml`. No replacement guidance was added since TLS termination for the control server has already been documented separately in #155 (control server setup docs).

## Testing

Docs-only change with no testable behavior — skipped TDD per the resolve-issue workflow. Verified via `git diff` that only the dead block was removed, and confirmed `grep -rn "cert|webserver"` across the config/CLI source has no matches tied to this section.

## Risks / Breaking changes

None — this is a comment/config-example cleanup with no code or runtime impact.

Closes #157

### DIFF
--- a/example.config.toml
+++ b/example.config.toml
@@ -69,20 +69,6 @@ json-url = []
     #template = "Switching to #<%- selectedIdx %> (with <%- voteCount %> votes)"
     #interval = 15
 
-[cert]
-# SSL certificates (optional)
-# If you specify an https:// URL for the "webserver" option, a certificate will be automatically generated and signed by Let's Encrypt.
-# For this to work, Streamwall must be accessible from both ports 80 and 443.
-
-# Uncomment to obtain a real certificate (otherwise a testing cert will be generated)
-#production = true
-
-# Email address to use when registering SSL certificate with Let's Encrypt
-#email = "you@example.com"
-
-# Directory to store certificate
-#dir = "../streamwall-cert"
-
 [streamdelay]
 #endpoint = "http://localhost:8404"
 #key = "super-secret"


NilsR0711:
Superseded by #163, which landed the identical fix (with an improved TLS-guidance note) while this PR was awaiting review. Closing without merging to avoid a duplicate/empty diff.

Issue #67, opened 2026-07-02
Docs: rewrite the [control] section of example.config.toml for the token/endpoint architecture


NilsR0711:
**Type: docs · Priority: high**

`example.config.toml:20-30` documents `address`/`hostname`/`port`/`username`/`password` — none exist. The app reads only `control.endpoint` (a WS URL incl. token, `index.ts:184-187,513`); auth is token/invite-based via the separate control server.

**Fix** — show `[control] endpoint = "ws://host/streamwall/<id>/ws?token=<secret>"` and explain the URL is printed by the control server at startup (`control-server/src/index.ts:486-489`).

_Source: feature/DX analysis_

Issue #68, opened 2026-07-02
Docs: remove the dead [cert]/Let's Encrypt section from example.config.toml


NilsR0711:
**Type: docs · Priority: medium**

`example.config.toml:60-72` promises automatic Let's Encrypt certificates, but there is no ACME/greenlock code anywhere and no `cert` key in `StreamwallConfig` (`index.ts:39-84`).

**Fix** — delete the section; document TLS-behind-reverse-proxy for the control server instead.

_Source: feature/DX analysis_

PR #160, opened 2026-07-14
docs: remove stale [cert] section, document TLS via reverse proxy

### LINKED ISSUE #154: Docs: [cert] section of example.config.toml documents a nonexistent "webserver" option and unimplemented cert automation
**Type: docs · Priority: low**

While fixing #67 (rewriting the `[control]` section of `example.config.toml`), found that the neighboring `[cert]` section is also stale:

```toml
[cert]
# SSL certificates (optional)
# If you specify an https:// URL for the "webserver" option, a certificate will be automatically generated and signed by Let's Encrypt.
# For this to work, Streamwall must be accessible from both ports 80 and 443.

# Uncomment to obtain a real certificate (otherwise a testing cert will be generated)
#production = true

# Email address to use when registering SSL certificate with Let's Encrypt
#email = "you@example.com"

# Directory to store certificate
#dir = "../streamwall-cert"
```

There is no `webserver` config option anywhere in the codebase (`grep -rn "webserver" packages/ --include="*.ts"` returns nothing), and the current control server (`packages/streamwall-control-server`) has no automatic Let's Encrypt / cert-generation logic — it just reads its `baseURL` from the `STREAMWALL_CONTROL_URL` env var (`packages/streamwall-control-server/src/index.ts:925`) and expects TLS termination to be handled externally (reverse proxy, etc.).

**Suggested fix** — either:
- Remove the `[cert]` section entirely and document externally-terminated TLS (reverse proxy) as the supported approach, or
- If automatic cert provisioning is still a desired feature, implement it and then fix the docs to match.

This needs a maintainer/product decision on which direction to take, so filing as a follow-up rather than resolving inline.

_Source: found while resolving #67 (PR #153)_

NilsR0711:
## Problem

The `[cert]` section in `example.config.toml` documented:
- An `https://` value for a `"webserver"` config option that does not exist anywhere in the codebase (`grep -rn "webserver" packages/ --include="*.ts"` returns nothing).
- Automatic Let's Encrypt certificate provisioning that `streamwall-control-server` never implemented — the server only reads its `baseURL` from `STREAMWALL_CONTROL_URL` and expects TLS to be terminated externally.

Found while resolving #67 (PR #153) and filed as #154 for a follow-up decision.

## Solution

Per the direction suggested in the issue and confirmed by the current control-server behavior (`STREAMWALL_CONTROL_URL`'s scheme just selects http/https behavior; the README already requires `wss://` for non-loopback remote endpoints), removed the `[cert]` section entirely and documented externally-terminated TLS (reverse proxy) as the supported approach:

- `example.config.toml`: removed the stale `[cert]` section.
- `README.md`: added a short note under "Remote control server" pointing to reverse-proxy TLS termination (nginx, Caddy, etc.) for `wss://`/`https://` endpoints.

## Testing

Docs-only change; no testable behavior. Verified `npx prettier --check README.md` passes.

## Risks / breaking changes

None — this only corrects documentation to match existing behavior; no code changed.

Closes #154

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ See
 for environment variable configuration (hostname/port, storage location, rate
 limits).
 
+`streamwall-control-server` does not generate or manage TLS certificates —
+for `wss://`/`https://` endpoints, terminate TLS with a reverse proxy (e.g.
+nginx, Caddy) in front of the server and set `STREAMWALL_CONTROL_URL` to the
+public `https://` address.
+
 ## Data sources
 
 Streamwall can load stream data from both JSON APIs and TOML files. Data sources can be specified in a config file (see `example.config.toml` for an example) or the command line:
--- a/example.config.toml
+++ b/example.config.toml
@@ -69,20 +69,6 @@ json-url = []
     #template = "Switching to #<%- selectedIdx %> (with <%- voteCount %> votes)"
     #interval = 15
 
-[cert]
-# SSL certificates (optional)
-# If you specify an https:// URL for the "webserver" option, a certificate will be automatically generated and signed by Let's Encrypt.
-# For this to work, Streamwall must be accessible from both ports 80 and 443.
-
-# Uncomment to obtain a real certificate (otherwise a testing cert will be generated)
-#production = true
-
-# Email address to use when registering SSL certificate with Let's Encrypt
-#email = "you@example.com"
-
-# Directory to store certificate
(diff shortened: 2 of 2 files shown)


NilsR0711:
Superseded by #163, which merged the identical fix (same [cert] section removal + reverse-proxy TLS note) for the duplicate issue #68. Closing to avoid a redundant/conflicting merge.

Issue #106, opened 2026-07-03
Security: revalidate redirects and HLS sub-requests loaded into WebContentsView (network-layer SSRF)


NilsR0711:
**Severity: high** — follow-up to #1 (fixed by #94, merged to `main`).

The SSRF guard added in #94 hardens `ensureValidURL` (`packages/streamwall/src/util.ts`), which validates the **initial URL string** before it is loaded into a `WebContentsView`. Two vectors remain that a single-URL string check structurally cannot cover, because the dangerous request is issued *after* validation:

**1. HTTP redirects.** `wc.loadURL(content.url)` (`viewStateMachine.ts` `loadPage`) and `browseWindow.loadURL(msg.url)` (`main/index.ts` browse handler) follow Chromium-level 3xx redirects with no revalidation. An operator supplies `http://attacker-public.example/` (passes validation), which responds `302 Location: http://169.254.169.254/latest/meta-data/`. Chromium follows it → SSRF into cloud metadata / loopback / LAN. No DNS rebinding required.

**2. HLS manifest contents.** For `.m3u8` URLs only the manifest URL is validated; `loadHTML(wc, 'playHLS', { query: { src } })` hands it to the renderer's hls.js, which then fetches the variant/segment URIs **named inside the manifest**, none of which pass through `ensureValidURL`. A public-IP host can serve a playlist whose segment URIs point at `http://169.254.169.254/…` or other internal hosts.

(Related: active DNS rebinding — `ensureValidURL` resolves the host, but the loader re-resolves afterwards (TOCTOU), so a rebinding attacker is not defended against. Documented in #94.)

**Remediation**
Move enforcement to the network layer of the view's session so every request (including redirects and sub-resources) is checked against the same non-public-address rules:
- `session.webRequest.onBeforeRequest` / `onBeforeRedirect` on the `WebContentsView`'s session, rejecting requests whose resolved address is non-public; and/or
- render operator/custom URLs in a locked-down session partition with a request-level allowlist (builds on the per-view partitions from #97).

This reuses the address classification already in `ensureValidURL`; consider extracting it to `streamwall-shared` so both the string check and the network hook share one blocklist.

_Source: adversarial review of #94._

PR #173, opened 2026-07-14
fix(security): classify ws:/wss: requests in the network-layer SSRF guard

### LINKED ISSUE #168: Security: WebSocket (ws/wss) requests bypass the network-layer SSRF guard
**Severity: medium** — follow-up to #106.

The network-layer SSRF guard added in #106 (`installRequestSSRFGuard` in `packages/streamwall/src/main/partitions.ts`) revalidates every request a view/browse session issues via `session.webRequest.onBeforeRequest`. To match the vectors reported in #106 (HTTP redirects and HLS sub-resources) and to avoid breaking the Vite dev-server HMR socket, the request decision (`findRequestBlockReason` in `packages/streamwall/src/util.ts`) only governs `http:`/`https:` and returns `null` (allow) for every other scheme — including `ws:`/`wss:`.

**Vector.** A page loaded into a `WebContentsView` (a compromised/hostile stream page, or content reached via an allowed redirect) can open a WebSocket to an internal host:

```js
new WebSocket('ws://169.254.169.254/') // or ws://127.0.0.1:<port>/, ws://<LAN host>/
```

The handshake is surfaced to `onBeforeRequest` with a `ws://`/`wss://` URL, which the guard currently lets through. This is lower-likelihood than the #106 vectors (it needs script execution inside the loaded page rather than a plain redirect/manifest), but it is the same class of loopback/LAN/cloud-metadata SSRF and is not covered by the current guard.

**Remediation**
- Extend `findRequestBlockReason` to also classify `ws:`/`wss:` against the same non-public-address rules.
- Because the check is origin-allow-listed for the dev server over `http`, the Vite HMR socket (`ws://localhost:<port>`) must be allow-listed too — extend the allow-list passed from `StreamWindow`/`loadHTML` to include the dev server's `ws:` origin, or allow-list by host rather than by full origin.
- Add unit tests mirroring the existing `findRequestBlockReason` / `installRequestSSRFGuard` tests for `ws://` and `wss://` targets (blocked private, allowed public, allowed dev socket).

_Source: adversarial review while implementing #106._

NilsR0711:
## Problem

Follow-up to #106/#170. The network-layer SSRF guard (`installRequestSSRFGuard` / `findRequestBlockReason`) revalidates every request a view/browse session issues, but `findRequestBlockReason` only classified `http:`/`https:` targets — every other scheme, including `ws:`/`wss:`, was passed through unconditionally.

A page loaded into a `WebContentsView` (a compromised/hostile stream page, or content reached via an allowed redirect) can open a WebSocket to an internal host from script:

```js
new WebSocket('ws://169.254.169.254/')
```

That handshake is surfaced to `onBeforeRequest` as a `ws://` URL and was let through, the same class of loopback/LAN/cloud-metadata SSRF as the already-covered http(s) redirect/sub-resource vectors.

## Solution

- `findRequestBlockReason` (`packages/streamwall/src/util.ts`) now also classifies `ws:`/`wss:` targets with the same non-public-address rules used for `http:`/`https:` (loopback hostname fast-path, IP-literal check, DNS resolution + fail-open-on-unresolvable).
- The dev-server allow-list in `installRequestSSRFGuard` (`packages/streamwall/src/main/partitions.ts`) previously matched by full origin (scheme+host+port). Since the Vite HMR socket connects over `ws://localhost:<port>` while the allow-listed entry is the dev server's `http://localhost:<port>` origin, those origins never matched and the HMR socket would have been cancelled once `ws:` became a governed scheme. Switched the match to compare by **host** (hostname+port) instead, so one allow-list entry covers both the `http:` origin and its `ws:` counterpart on the same host — as suggested in the issue.

## Testing

TDD: added failing tests first, then implemented.

- `packages/streamwall/src/util.test.ts`: `findRequestBlockReason` now has ws:/wss: coverage — blocks the cloud-metadata literal, blocks a loopback hostname, blocks a public host resolving to a private address, allows a public host resolving to a public address, and fails open on an unresolvable host (mirrors the existing http(s) test matrix).
- `packages/streamwall/src/main/partitions.test.ts`: `installRequestSSRFGuard` now has coverage for the host-based allow-list — a `ws:` request to the allow-listed dev-server host is let through without consulting the reason lookup, while a `ws:` request to a different, non-allow-listed host is still blocked.
- Full quality gate run locally: prettier, eslint, `tsc --noEmit` (all workspaces), and the full test suite (`packages/streamwall` 126/126, `streamwall-control-server` 73/73, `streamwall-control-ui` 22/22) — all green.

## Risks / breaking changes

None expected. This only tightens an existing security guard to cover a scheme it previously ignored; legitimate `ws:`/`wss:` traffic to public hosts (and the dev-server's own HMR socket) continues to work.

Closes #168

### DIFF
--- a/packages/streamwall/src/main/partitions.test.ts
+++ b/packages/streamwall/src/main/partitions.test.ts
@@ -191,6 +191,25 @@ test('installRequestSSRFGuard allows an explicitly allowed origin without consul
   )
 })
 
+test('installRequestSSRFGuard allows a ws: request to the allow-listed dev server host', async () => {
+  // The Vite HMR socket connects over ws: to the same host:port the dev
+  // server's http: origin is allow-listed for; the allow-list must match by
+  // host so this is not treated as a different, unlisted origin.
+  const session = guardWith(
+    { 'ws://localhost:5173/vite-hmr': 'loopback host' },
+    ['http://localhost:5173'],
+  )
+  assert.equal(await session.requestURL('ws://localhost:5173/vite-hmr'), false)
+})
+
+test('installRequestSSRFGuard still blocks a ws: request to a different, non-allow-listed host', async () => {
+  const session = guardWith(
+    { 'ws://169.254.169.254/': 'blocking request to private-network address' },
+    ['http://localhost:5173'],
+  )
+  assert.equal(await session.requestURL('ws://169.254.169.254/'), true)
+})
+
 test('installRequestSSRFGuard fails open if the reason lookup itself throws', async () => {
   const session = fakeSession()
   installRequestSSRFGuard(session, {
--- a/packages/streamwall/src/main/partitions.ts
+++ b/packages/streamwall/src/main/partitions.ts
@@ -63,20 +63,32 @@ interface RequestFilteringSession {
 export interface RequestGuardOptions {
   /**
    * Origins that bypass the address check entirely (e.g. the Vite dev server on
-   * loopback, which serves the HLS renderer page). Empty in a packaged build.
+   * loopback, which serves the HLS renderer page). Matched by host (hostname
+   * plus port), not full origin, so the dev server's ws: HMR socket is covered
+   * by the same entry as its http: origin. Empty in a packaged build.
    */
   allowedOrigins?: readonly string[]
   /** Overridable for tests; defaults to the real address classifier. */
   findBlockReason?: (url: string) => Promise<string | null>
 }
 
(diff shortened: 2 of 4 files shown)


NilsR0711:
Follow-up filed for an unrelated flaky test noticed while running the quality gate: #174 (`wsValidation.test.ts` times out under the full-workspace `npm run test`, but passes reliably standalone; reproduced on `main` too).

PR #175, opened 2026-07-14
fix(security): resolve validated hosts through the loader's own session

### LINKED ISSUE #169: Security: defend against active DNS rebinding for view/browse loads (TOCTOU)
**Severity: medium** — tracking the known limitation carried by #1/#94/#106.

Both SSRF layers resolve the target host and then hand the *URL* to Chromium, which re-resolves it independently:

- `ensureValidURL` (`packages/streamwall/src/util.ts`) resolves the host, checks every address, then returns — after which `wc.loadURL()` re-resolves.
- The #106 network guard (`findRequestBlockReason` / `installRequestSSRFGuard`) likewise resolves in `onBeforeRequest`, then Chromium performs its own resolution for the actual connection.

This is a time-of-check/time-of-use gap: an attacker who controls a DNS name can answer the guard's lookup with a public address and Chromium's subsequent lookup with a private one (short TTL / DNS rebinding), reaching loopback/LAN/cloud-metadata despite both checks passing. #106 explicitly documented this as out of scope; this issue exists so the remaining limitation is tracked rather than only living in commit/issue prose.

**Possible remediations** (need evaluation)
- Pin the resolved-and-vetted address for the connection so check and use share one resolution — e.g. a custom `session.resolveHost`/`resolveProxy` path, or resolving ourselves and connecting by IP with SNI/Host preserved. Electron/Chromium support here is limited and needs research.
- Or enforce at the OS/network level (deny the app process egress to private ranges).
- At minimum, add a regression test asserting the current documented behavior so any future change is deliberate.

**Impact.** Requires an attacker-controlled domain and rebinding setup; the literal-address and static-DNS vectors (the common operator-SSRF cases) are already blocked by #94 and #106.

_Source: adversarial review of #94, reaffirmed while implementing #106._

NilsR0711:
## Problem

Both SSRF layers (`ensureValidURL` and the `#106` network guard,
`findRequestBlockReason`/`installRequestSSRFGuard`) resolved a hostname
themselves — via Node's independent `dns.lookup` — to classify it as
public/private, then handed the *URL string* to Chromium, which re-resolves
it independently to actually connect.

This is a time-of-check/time-of-use gap: an attacker who controls a DNS name
can answer our lookup with a public address and Chromium's later lookup with
a private one (DNS rebinding), reaching loopback/LAN/cloud-metadata despite
both checks passing. #106 explicitly documented this as out of scope; #169
tracks the remaining limitation.

## Solution

Added `createSessionHostResolver` (`packages/streamwall/src/util.ts`), which
builds a `HostAddressResolver` backed by `session.resolveHost(...)` — the
same resolver, and same cache, Chromium itself uses to connect — instead of
Node's independent `dns.lookup`.

Wired it into every call site that has a session available:
- `main/index.ts`: browse-window `ensureValidURL` uses
  `browseWindow.webContents.session`.
- `main/viewStateMachine.ts`: view-load `ensureValidURL` uses
  `view.webContents.session`.
- `main/partitions.ts`: `installRequestSSRFGuard`'s default resolver (used
  by every session `hardenSession` protects) is now bound to the guarded
  session itself, instead of falling back to `findRequestBlockReason`'s
  Node-`dns`-backed default.

### Why this helps, and what it doesn't fix

Sharing the resolver/cache means the address we validate is very likely the
exact one Chromium reuses moments later for the real connection — narrowing
the gap to only a short/zero-TTL adversarial DNS response racing the two
lookups by milliseconds, rather than being wide open on every single
request regardless of TTL (the previous state, where the two lookups used
entirely unrelated resolvers).

It does **not** close the gap outright. Electron/Chromium expose no API to
pin a single request's connection to a pre-validated address — the
command-line `--host-resolver-rules` switch that Chromium's network stack
does honor is static and set once at process launch, so it can't be used to
pin arbitrary, dynamically-supplied operator hostnames discovered at
runtime. This is documented in code (`createSessionHostResolver`'s JSDoc)
so any future change to close the gap further is deliberate, per the
issue's own "at minimum" bar.

The network-layer guard (`findRequestBlockReason`/`installRequestSSRFGuard`)
therefore remains necessary as defense in depth, not a complete fix on its
own.

## Testing

TDD: failing tests written first (`createSessionHostResolver` didn't exist),
confirmed red, then implemented to green.

- `packages/streamwall/src/util.test.ts`: new tests for
  `createSessionHostResolver` (endpoint flattening, empty-endpoint case) and
  for `ensureValidURL`/`findRequestBlockReason` classifying addresses
  reported by a session resolver.
- `packages/streamwall/src/main/partitions.test.ts`: `fakeSession()` now
  exposes a stubbable `resolveHost`; new tests assert
  `installRequestSSRFGuard`'s default resolver is the guarded session's own
  `resolveHost` (not an independent lookup), for both the allow and block
  paths. Also tightened the existing `hardenSession also installs the
  network-layer SSRF guard` test to use the stub instead of relying on a
  real DNS lookup for `cdn.twitch.tv`, removing a hidden network dependency
  from the unit suite.

All existing SSRF/loopback/private-range regression tests still pass
unchanged. Full quality gate green locally: format, lint, typecheck, full
workspace test suite (`npm run test` across all packages), and
`electron-forge package` build.

## Risks / breaking changes

None expected. Purely additive resolver wiring; the classification logic
(`isBlockedAddress`, `isLoopbackHostname`) is unchanged. No behavior change
for callers that don't have a session (none currently — `ensureValidURL` is
called at both real call sites with a session-bound resolver now).

Closes #169

### DIFF
--- a/packages/streamwall/src/main/index.ts
+++ b/packages/streamwall/src/main/index.ts
@@ -18,7 +18,7 @@ import { updateElectronApp } from 'update-electron-app'
 import WebSocket from 'ws'
 import yargs from 'yargs'
 import * as Y from 'yjs'
-import { ensureValidURL } from '../util'
+import { createSessionHostResolver, ensureValidURL } from '../util'
 import { ConfigError, parseConfigToml, validateConfig } from './config'
 import ControlWindow from './ControlWindow'
 import {
@@ -503,7 +503,10 @@ async function main(argv: ReturnType<typeof parseArgs>) {
       if (msg.type === 'browse') {
         console.debug('Attempting to browse URL:', msg.url)
         try {
-          await ensureValidURL(msg.url)
+          await ensureValidURL(
+            msg.url,
+            createSessionHostResolver(browseWindow.webContents.session),
+          )
           browseWindow.loadURL(msg.url)
         } catch (error) {
           console.error('Invalid URL:', msg.url)
--- a/packages/streamwall/src/main/partitions.test.ts
+++ b/packages/streamwall/src/main/partitions.test.ts
@@ -31,6 +31,11 @@ function fakeSession() {
         requestListener = listener
       },
     },
+    // Overridable per test; the default reports every hostname as public so
+    // tests that don't care about DNS classification aren't forced to stub it.
+    resolveHost: async (): Promise<{
+      endpoints: { address: string; family: 'ipv4' | 'ipv6' }[]
+    }> => ({ endpoints: [{ address: '93.184.216.34', family: 'ipv4' }] }),
     request(permission: string): boolean {
       assert.ok(handler, 'a permission request handler must be registered')
       let granted: boolean | undefined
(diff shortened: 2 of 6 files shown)


NilsR0711:
Follow-up: opened #178 to evaluate OS/network-level egress enforcement as a deeper, structural mitigation for the residual DNS-rebinding gap (defense in depth beyond this PR's session-resolver narrowing).

PR #156, opened 2026-07-14
feat: surface stream load errors on the wall and add auto-retry with backoff

NilsR0711:
## Problem

When a view entered `displaying.error` the overlay filtered it out entirely, leaving a silent black cell. The `error` state was terminal (no auto-retry), and the `stalled` playback state had no watchdog, so a transient network blip permanently lost a cell until an operator manually reloaded it. The only failure signal in the control UI was a red border, with no reason.

Closes #75.

## Solution

**State machine (`viewStateMachine.ts`)**
- Records a human-readable `error` reason in context (via a new `setError` action + `formatError` helper), cleared on a fresh load/recovery.
- Auto-retries `error` and `stalled` views on an exponentially growing, capped backoff (`min(delay · 2^n, maxDelay)`) via xstate delayed transitions, guarded by a `canRetry` guard (`enabled && retryCount < maxRetries`).
- Adds a watchdog to the `stalled` state so a view that outlasts `stalledTimeout` is reloaded.
- Resets the retry budget when the view reaches `running` (recovery) or on a manual `RELOAD`.

**Wiring**
- `ViewState.context` now carries `error`, serialized by `StreamWindow.emitState()`; `ViewStateValue` also gains the previously-missing `running.playback` sub-state.
- New `[retry]` config section (`enabled`, `delay`, `max-delay`, `max-retries`, `stalled-timeout`) plumbed through yargs, the zod validation schema, and `StreamWindow` (seconds in config → ms in the machine; sensible defaults, retry on by default).

**UI**
- Overlay: error views are no longer filtered out — they render an error tile (warning icon, stream label, failure reason, red border).
- Control UI: adds a warning badge with the reason (full reason as tooltip; icon-only on small cells) alongside the existing red border.

## Architecture decisions

- Backoff lives in the state machine as delayed transitions so retry state is per-view, self-cancelling on recovery/content-change, and unit-testable with fake timers.
- The `displaying.error → displaying.loading` retry is an internal transition (LCA `displaying`), so the `displaying` entry reset does **not** clobber the incremented `retryCount`; a content change re-targets `.displaying` and *does* reset. Both paths are covered by tests.
- Retry config passed to `StreamWindow` as an optional second constructor arg (defaulting to `DEFAULT_RETRY_CONFIG`) rather than folded into `StreamWindowConfig`, to avoid leaking retry internals into the broadcast client state.

## Testing

[...]

Issue #75, opened 2026-07-02
Feature: surface stream load errors on the wall and add auto-retry with backoff


NilsR0711:
**Type: feature · Priority: high**

When a view enters `displaying.error` the overlay filters it out entirely (`overlay.tsx:37-41`) — a silent black cell. The `error` state is terminal (`viewStateMachine.ts:353`), there's no auto-retry, and the `stalled` playback state (`:308-314`) has no watchdog.

**Desired** — an error badge/reason on the wall tile and in the control UI (currently just a red border, `control-ui/src/index.tsx:1181,2159-2160`), plus configurable auto-retry with backoff for error/stalled views.

_Source: feature/DX analysis_

PR #176, opened 2026-07-14
test: cover the overlay error tile and control-UI error badge rendering

### LINKED ISSUE #158: test: cover the overlay error tile and control-UI error badge rendering
## Context

PR #156 (issue #75) added user-facing error rendering:

- **Overlay** (`packages/streamwall/src/renderer/overlay.tsx`): error views are no longer filtered out — they render an error tile (`ErrorCover`/`ErrorHeading`/`ErrorReason`) with a warning icon, label, reason, and red border.
- **Control UI** (`packages/streamwall-control-ui/src/index.tsx`): errored grid-preview boxes now show a `StyledGridError` badge with the reason (tooltip on hover, icon-only on small cells).

The underlying state-machine logic (error reason capture, backoff, stalled watchdog) is thoroughly unit-tested in `viewStateMachine.test.ts`, but the **presentational rendering** of these two pieces has no direct test.

## Problem

- `overlay.tsx` has no component tests at all, and its `Overlay` component is not exported (the module renders straight to `document.body`), so there is no seam to assert on. A regression in the error-tile branch (e.g. the `matchesState('displaying.error', …)` filter, or a missing reason) would go unnoticed.
- The control-UI error badge (`StyledGridError`) is only reachable through the ~1800-line `ControlUI` component and is not currently exercised by the happy-dom component tests.

## Suggested solution

- Export the `Overlay` component (or extract the per-view tile into a small exported component) and add happy-dom/vitest component tests asserting: error views render the badge + reason, non-error views do not, and the reason falls back to "Stream error" when `context.error` is null.
- Add a focused control-UI test (mirroring the existing `LazyChangeInput`/component-test setup) that a view in `displaying.error` renders `StyledGridError` with the reason, and that the tooltip/collapse-on-small behaviour holds.

## Impact

Low risk (presentational), but locks in the new error-surfacing UX so future refactors of the overlay/grid preview don't silently drop it.

NilsR0711:
## Problem

PR #156 (issue #75) added user-facing error rendering:

- **Overlay** (`packages/streamwall/src/renderer/overlay.tsx`): errored views render an error tile (warning icon, label, reason, red border).
- **Control UI** (`packages/streamwall-control-ui/src/index.tsx`): errored grid-preview boxes show a `StyledGridError` badge with the reason.

Both pieces were only reachable through large, self-rendering component trees with no seam to test against, so neither had direct coverage.

## Solution

- Extracted the per-view error/non-error tile out of `overlay.tsx` into a new, exported `OverlayViewTile` component (`packages/streamwall/src/renderer/OverlayViewTile.tsx`). This avoids the module-level `render(<App />, document.body)` side effect that made the original module unsafe to import in a test.
- Extracted the per-view grid-preview tile out of the ~1800-line `ControlUI` component into a new, exported `GridPreviewBox` component in `packages/streamwall-control-ui/src/index.tsx`. No behavior change - the JSX and props are unchanged, just given a name and a seam.
- Added component tests for both:
  - `OverlayViewTile.test.tsx`: errored views render the warning badge + reason; non-error views render neither; the heading falls back to "Stream error" when there is no label/reason.
  - `GridPreviewBox.test.tsx`: errored views render `StyledGridError` with the reason as text and as a `title` tooltip; non-error views render no badge; the heading falls back to "Stream error" with no reason; the info panel gets the `.small` class that collapses the reason text on small cells.
- Added a `vitest.config.ts` + `happy-dom` devDependency to the `streamwall` package (mirroring `streamwall-control-ui`'s existing setup) so its Preact renderer components can be tested with a DOM, opted into per-file via `// @vitest-environment happy-dom`.
- Excluded `**/*.test.tsx` from `streamwall`'s production `tsc` typecheck, matching the existing `**/*.test.ts` exclusion (there were previously no `.test.tsx` files in that package).

### A note on the test setup

Rendering `react-icons` components and `styled-components` elements directly under this repo's `preact` + `preact/compat` + `happy-dom` test stack currently breaks: react-icons' `Context.Consumer` usage throws (`Cannot add property __, object is not extensible`), and any `styled.div`-produced element ends up with a garbled `tagName` (e.g. `.SC-EGGEI`) instead of `DIV`. Both reproduce identically under `jsdom` too, so this isn't a happy-dom quirk - it's a real incompatibility between this `styled-components`/`react-icons` version pairing and Preact's non-compat renderer internals. Neither of the two new test files hit this in the app itself (it works fine in the real Electron/browser render), only in this specific test harness.

To keep this PR scoped to the issue, both test files stub out the icon components via `vi.mock` and assert on `textContent`/attributes/`parentElement` rather than tag names or CSS selectors that assume real `<div>`s. I opened a follow-up issue to track fixing (or at least documenting) the underlying incompatibility so future component tests don't have to rediscover it. See below.

## Testing

- `npm run typecheck --workspaces --if-present`
- `npm test --workspaces --if-present` (all packages, incl. the two new test files)
- `npm run lint` / `npx prettier --check` on all changed files
- `npm run package` in `packages/streamwall` (Electron Forge build) to confirm the `overlay.tsx` extraction doesn't break the production build

## Risks / breaking changes

None - both changes are pure extractions (no behavior change) plus new tests. No production code paths changed.

## Follow-up issues

- Follow-up filed for the `react-icons`/`styled-components` + Preact test-environment incompatibility discovered while writing these tests (link added once created).

Closes #158

### DIFF
--- a/packages/streamwall-control-ui/src/GridPreviewBox.test.tsx
+++ b/packages/streamwall-control-ui/src/GridPreviewBox.test.tsx
@@ -1,0 +1,92 @@
+import Color from 'color'
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+import { GridPreviewBox } from './index.tsx'
+
+// react-icons renders through preact/compat's Context.Consumer, which
+// currently crashes under this package's happy-dom test environment
+// (unrelated to the markup under test here) - stub the icon out so the
+// component's own rendering logic can be exercised in isolation.
+vi.mock('react-icons/fa', () => ({
+  FaExclamationTriangle: () => <i data-icon="warning" />,
+}))
+vi.mock('react-icons/md', () => ({
+  MdOutlineStayCurrentLandscape: () => null,
+  MdOutlineStayCurrentPortrait: () => null,
+}))
+
+let container: HTMLDivElement | undefined
+
+afterEach(() => {
+  if (container) {
+    act(() => render(null, container!))
+    container.remove()
+    container = undefined
+  }
+})
+
+function renderBox(
+  props: Partial<Parameters<typeof GridPreviewBox>[0]> = {},
+): HTMLDivElement {
+  container = document.createElement('div')
+  document.body.appendChild(container)
+  act(() => {
+    render(
+      <GridPreviewBox
+        streamId="abc"
+        color={Color('red')}
+        pos={{ x: 0, y: 0, width: 100, height: 100, spaces: [0] }}
+        windowWidth={100}
+        windowHeight={100}
+        isListening={false}
+        isSmall={false}
+        isError={false}
+        errorReason={null}
+        orientation={null}
+        source="Example Source"
+        city={undefined}
+        state={undefined}
+        {...props}
+      />,
+      container!,
+    )
+  })
+  return container
+}
+
+describe('GridPreviewBox', () => {
+  test('renders the error badge with the reason when the view is errored', () => {
+    const box = renderBox({
+      isError: true,
+      errorReason: 'Connection lost',
+    })
+
+    const badge = box.querySelector('[data-icon="warning"]')?.parentElement
+    expect(badge?.textContent).toContain('Connection lost')
+    expect(badge?.getAttribute('title')).toBe('Connection lost')
+  })
+
+  test('falls back to a generic label when the errored view has no reason', () => {
+    const box = renderBox({
+      isError: true,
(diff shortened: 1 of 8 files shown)


NilsR0711:
Follow-up filed for the react-icons/styled-components + Preact test-environment incompatibility discovered while writing these tests: #177

NilsR0711:
CI is fully green (all required checks, including "CI OK", passed). No Cursor BugBot review has appeared after ~20 minutes, so treating the AI-review requirement as satisfied per the repo's convention for unavailable/rate-limited review runs.

PR #181, opened 2026-07-14
fix(test): resolve react-icons and styled-components to preact/compat in tests

### LINKED ISSUE #177: react-icons and styled-components break under the preact/happy-dom test stack
## Context

While adding component tests for the overlay error tile and control-UI error badge (#158, PR #176), rendering the app's actual `react-icons` and `styled-components` elements under this repo's test stack (`preact` + `preact/compat` aliasing + `happy-dom`) turned out to be broken:

1. **`react-icons`**: any icon component (e.g. `FaExclamationTriangle` from `react-icons/fa`) throws when rendered:
   ```
   TypeError: Cannot add property __, object is not extensible
       at type ../../node_modules/preact/src/diff/children.js:219:15
   ```
   `react-icons`' `IconBase` renders through `IconContext.Consumer` (a `React.createElement`-based render-prop pattern via the `react` → `preact/compat` alias). This crashes regardless of `preact/compat` being preloaded, and regardless of whether `render`/`createElement` come from `preact` or `preact/compat`.

2. **`styled-components`**: any `styled.div`-produced element renders with a garbled `tagName` instead of `DIV`. E.g. a `Box = styled.div\`...\`` rendered as `<Box />` ends up as an element whose `tagName` is the generated CSS selector (`.SC-BDVVNZ`) rather than `DIV`. `Box` itself is a `React.forwardRef`-shaped object (`$$typeof: Symbol(react.forward_ref)`) matching preact/compat's own `forwardRef` symbol, but something in the render pipeline still resolves the element's tag incorrectly.

Both reproduce identically whether the environment is `happy-dom` or `jsdom` (jsdom throws `InvalidCharacterError: ".sc-bdvvNz" did not match the QName production`, confirming an invalid tag name is genuinely being passed to `document.createElement`, not a lenient-DOM quirk). Neither issue reproduces in the real app (Electron/browser render) - only in this specific `vitest` + `preact` (non-React) test harness.

## Impact

Any future component test that renders a `react-icons` icon or a `styled-components` element directly will hit this. The two new test files added in PR #176 work around it by:
- `vi.mock`-ing the relevant `react-icons/*` imports with stub components
- Avoiding tag-name/CSS-selector-based DOM queries (e.g. using `element.parentElement` instead of `element.closest('div')`, and asserting on `className`/attributes/`textContent` rather than `tagName`)

This is a workable pattern for now, but it's easy to rediscover this the hard way (confusing `TypeError`/`InvalidCharacterError` with no obvious cause) and it silently limits what a component test can assert (e.g. can't verify an element renders as a real `<div>`).

## Suggested solution

Investigate the root cause - likely a version mismatch or missing preact/compat shim for one of:
- `react-icons`' `IconContext.Consumer` pattern under preact's non-compat internal diffing
- `styled-components` v6's `forwardRef`-based component resolution under preact/compat

Options to explore:
- Pin/bump `preact`, `preact/compat`, `styled-components`, or `react-icons` to versions known to interoperate correctly in tests
- Add an explicit compat shim/setup (beyond the existing `import 'preact/compat'` in `streamwall-control-ui/src/test-setup.ts`)
- If unfixable within a reasonable window, document the workaround (mock icons, avoid tag-based queries) in a testing guide so future contributors don't have to rediscover it

## Where to look

- `packages/streamwall-control-ui/vitest.config.ts` and `src/test-setup.ts` (existing preact/compat + happy-dom setup)
- `packages/streamwall/vitest.config.ts` (new, added in PR #176, same alias/jsx approach)
- Reproduction: render any `react-icons/fa` icon, or any `styled.div`-based component, directly via `preact`'s `render()` in a `happy-dom` or `jsdom` test environment.

NilsR0711:
## Problem

`react-icons` and `styled-components` break when rendered directly under this repo's `vitest` + `preact` (non-React) + `happy-dom` component test stack:

- `react-icons` (e.g. `FaExclamationTriangle`) throws `TypeError: Cannot add property __, object is not extensible` from Preact's reconciler.
- `styled-components` (e.g. `styled.div`) renders elements with their generated CSS class name as the DOM tag (`.sc-bdvvNz`) instead of the real tag (`DIV`).

Neither reproduces in the real Electron/browser app - only under this test harness. The tests added in #176 worked around both by mocking `react-icons` and avoiding tag-based DOM queries.

## Root cause

`vitest.config.ts` aliases `react`/`react-dom` to `preact/compat` so these libraries pick up Preact instead of React at runtime. But `react-icons` and `styled-components` each ship both a CJS build and an ESM build:

- Their **CJS build**'s internal `require('react')` is resolved as an opaque external module by vitest's SSR-like module runner, which bypasses `resolve.alias` entirely.
- Their **ESM build**'s `import ... from 'react'` goes through Vite's resolver, which correctly honors the alias.

Under the existing config, vitest was loading the CJS build for both packages, so they resolved the real `react` package - installed automatically as their npm peer dependency - instead of `preact/compat`. That's what caused the crash (a frozen React element hitting Preact's non-compat reconciler) and the wrong-tag rendering (styled-components' `forwardRef`/`$$typeof` interop checks - see [styled-components#5736](https://github.com/styled-components/styled-components/pull/5736) - never got a chance to run against a `preact/compat`-shaped component).

## Fix

In both `packages/streamwall/vitest.config.ts` and `packages/streamwall-control-ui/vitest.config.ts`:

- `resolve.mainFields: ['browser', 'module', 'main']` makes Vite prefer `react-icons`'/`styled-components`' ESM/browser builds.
- `test.server.deps.inline: [/react-icons/, /styled-components/]` forces both through Vite's transform pipeline instead of being loaded as opaque external CJS modules, so their ESM `import` specifiers actually resolve through the `react` alias.

One residual gap: `svg-loaders-react` (used via `styled(TailSpin)` in `OverlayViewTile`, for the loading spinner) ships CJS only, with no ESM build for this fix to route through - it still resolves real `react` and still crashes. It's mocked in `OverlayViewTile.test.tsx` the same way `react-icons` used to be, since there's no ESM build to redirect through the alias.

## Testing

- Added `packages/streamwall-control-ui/src/preactCompatInterop.test.tsx`, a focused regression test that renders a real (unmocked) `react-icons` icon and a real `styled-components` element and asserts the actual rendered DOM (a real `<svg>`, and the real `DIV` tag) - verified red (fails without the config fix) before green (passes with it).
- Removed the `vi.mock('react-icons/...')` workarounds from `GridPreviewBox.test.tsx` and `OverlayViewTile.test.tsx` now that the underlying resolution is fixed, switching their assertions from a mock-only `data-icon` attribute to querying the real rendered `<svg>` / DOM tag - this also gives the styled-components error badge wrapper a genuine tag-name assertion, which wasn't previously possible per the issue.
- Full workspace test suite (`npm run test --workspaces`), `npm run typecheck --workspaces`, and `npm run lint` all pass.

## Risks / breaking changes

None expected - this is test-tooling only (`vitest.config.ts` in the two Preact packages, plus the two test files); no production code paths change.

Closes #177

### DIFF
--- a/packages/streamwall-control-ui/src/GridPreviewBox.test.tsx
+++ b/packages/streamwall-control-ui/src/GridPreviewBox.test.tsx
@@ -1,20 +1,8 @@
 import Color from 'color'
 import { render } from 'preact'
 import { act } from 'preact/test-utils'
-import { afterEach, describe, expect, test, vi } from 'vitest'
+import { afterEach, describe, expect, test } from 'vitest'
 import { GridPreviewBox } from './index.tsx'
-
-// react-icons renders through preact/compat's Context.Consumer, which
-// currently crashes under this package's happy-dom test environment
-// (unrelated to the markup under test here) - stub the icon out so the
-// component's own rendering logic can be exercised in isolation.
-vi.mock('react-icons/fa', () => ({
-  FaExclamationTriangle: () => <i data-icon="warning" />,
-}))
-vi.mock('react-icons/md', () => ({
-  MdOutlineStayCurrentLandscape: () => null,
-  MdOutlineStayCurrentPortrait: () => null,
-}))
 
 let container: HTMLDivElement | undefined
 
@@ -62,7 +50,8 @@ describe('GridPreviewBox', () => {
       errorReason: 'Connection lost',
     })
 
-    const badge = box.querySelector('[data-icon="warning"]')?.parentElement
+    const badge = box.querySelector('svg')?.parentElement
+    expect(badge?.tagName).toBe('DIV')
     expect(badge?.textContent).toContain('Connection lost')
     expect(badge?.getAttribute('title')).toBe('Connection lost')
   })
@@ -73,7 +62,7 @@ describe('GridPreviewBox', () => {
       errorReason: null,
     })
 
-    const badge = box.querySelector('[data-icon="warning"]')?.parentElement
+    const badge = box.querySelector('svg')?.parentElement
     expect(badge?.textContent).toContain('Stream error')
     expect(badge?.hasAttribute('title')).toBe(false)
   })
@@ -81,7 +70,7 @@ describe('GridPreviewBox', () => {
   test('does not render the error badge for a non-error view', () => {
     const box = renderBox({ isError: false })
 
-    expect(box.querySelector('[data-icon="warning"]')).toBeNull()
+    expect(box.querySelector('svg')).toBeNull()
   })
 
   test('marks the info panel as small so the reason text collapses on small cells', () => {
--- a/packages/streamwall-control-ui/src/preactCompatInterop.test.tsx
+++ b/packages/streamwall-control-ui/src/preactCompatInterop.test.tsx
@@ -1,0 +1,45 @@
+// @vitest-environment happy-dom
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import { FaExclamationTriangle } from 'react-icons/fa'
+import styled from 'styled-components'
+import { afterEach, describe, expect, test } from 'vitest'
+
+// Regression test for https://github.com/NilsR0711/streamwall/issues/177:
+// react-icons and styled-components must resolve `react` to `preact/compat`
+// (see vitest.config.ts) rather than the real `react` package that npm
+// installs as their peer dependency, or rendering either throws / produces
+// the wrong DOM tag.
+
+const Box = styled.div`
+  color: red;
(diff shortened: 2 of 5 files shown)


NilsR0711:
Filed a follow-up for the one residual gap this PR doesn't reach: #182 (`svg-loaders-react` has no ESM build, so `styled(TailSpin)` in `OverlayViewTile` still can't resolve `preact/compat` and stays mocked in `OverlayViewTile.test.tsx`).

PR #191, opened 2026-07-14
fix(electron): hide window on macOS close and flush storage before quit

### LINKED ISSUE #88: DX: handle the macOS window-close convention and flush storage on quit
**Type: dx · Priority: low**

Closing the stream window hard-exits via `streamWindow.on('close', () => process.exit(0))` (TODO at `index.ts:508-511`). On macOS the convention is hide + reopen from dock; `process.exit(0)` also skips graceful teardown (throttled storage writes at `:342-347` may be lost — see the persistence-flush bug).

**Desired** — hide on macOS with dock reopen; flush storage before quit on all platforms.

_Source: feature/DX analysis_

NilsR0711:
## Problem

Closing the stream window hard-exited via `streamWindow.on('close', () => process.exit(0))`:

- On macOS this skips the platform convention: closing the main window should hide it and keep the app running in the dock, reopening on activate (dock icon click).
- `process.exit(0)` bypasses the app's quit sequence entirely, so a pending throttled `stateDoc` write (persisted at most once per second, see the `throttle(..., 1000)` around the storage persist) can still be in-flight and never reaches disk -- recent grid/view changes are silently lost.

## Solution

- Extracted the hide-vs-quit decision into a pure, unit-tested `shouldHideInsteadOfQuit(platform, isQuitting)` helper (`windowCloseBehavior.ts`): hide the window on macOS unless the app is actually quitting; quit on every other platform (and on macOS once quitting has started).
- Wired an app-level `activate` handler to reshow the stream window from the dock, matching the standard Electron single-window pattern.
- Added `flushStorage(db, flushPendingUpdate)` to `storage.ts`: it synchronously forces the pending throttled update to run and then awaits `db.write()`, guaranteeing the latest state is actually on disk before the process exits.
- Hooked the flush into `before-quit` (deferring the real quit with `event.preventDefault()` until the flush settles, guarded by a one-shot flag), so it covers every quit path -- the stream window's close button, Cmd+Q, and the app menu -- not just the single spot the old code handled.

## Testing performed

- `windowCloseBehavior.test.ts`: covers hide-on-macOS-while-running, quit-on-macOS-while-quitting, and quit on Windows/Linux regardless of quitting state.
- `storage.test.ts`: covers forcing a pending throttled update to run before persisting, persisting already-current data, and that `db.data` is actually written to the adapter.
- Full workspace quality gate run locally: `format:check`, `lint`, `typecheck` (all workspaces), `npm test` (all workspaces, 153 tests in the `streamwall` package), and the control-client build -- all green.
- Manual verification of the actual Electron app run wasn't performed in this sandboxed environment (no GUI); behavior is covered by the extracted unit tests plus code review of the Electron API usage (`BrowserWindow.hide()`/`show()`, `app.on('activate'|'before-quit')`, `event.preventDefault()`).

## Risks / breaking changes

None expected. Behavior on Windows/Linux is unchanged (window close still quits the app), except storage is now guaranteed to flush first. On macOS, closing the window no longer quits the app -- this is the desired convention change requested by the issue, but it is a user-visible behavior change worth calling out.

Closes #88

### DIFF
--- a/packages/streamwall/src/main/StreamWindow.ts
+++ b/packages/streamwall/src/main/StreamWindow.ts
@@ -1,5 +1,11 @@
 import assert from 'assert'
-import { BrowserWindow, ipcMain, WebContents, WebContentsView } from 'electron'
+import {
+  BrowserWindow,
+  Event as ElectronEvent,
+  ipcMain,
+  WebContents,
+  WebContentsView,
+} from 'electron'
 import EventEmitter from 'events'
 import intersection from 'lodash/intersection'
 import isEqual from 'lodash/isEqual'
@@ -35,7 +41,7 @@ function getDisplayOptions(stream: StreamData): ContentDisplayOptions {
 
 export interface StreamWindowEventMap {
   load: []
-  close: []
+  close: [ElectronEvent]
   state: [ViewState[]]
   resize: []
 }
@@ -71,7 +77,7 @@ export default class StreamWindow extends EventEmitter<StreamWindowEventMap> {
     })
     win.removeMenu()
     win.loadURL('about:blank')
-    win.on('close', () => this.emit('close'))
+    win.on('close', (event) => this.emit('close', event))
 
     win.once('ready-to-show', () => {
       win.show()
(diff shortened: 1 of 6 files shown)


NilsR0711:
BugBot has not posted a review in the ~13 minutes since the last push, and CI (\`CI OK\`) is fully green. Treating the AI-review requirement as satisfied per the review-not-running fallback.

Filed a follow-up for a related gap noticed while implementing this fix: #193 (ControlWindow close doesn't quit the app; no `window-all-closed` handler). Out of scope for this PR, which is specifically about the stream window's close/quit behavior per #88.

PR #194, opened 2026-07-14
feat(control-ui): support touch/pointer input for tile drag-move and resize

### LINKED ISSUE #80: Feature: support touch/pointer input for tile drag-move and resize

The new interactions are mouse-only (`handleGridMouseDown`, `window.addEventListener('mouseup', …)`, `control-ui/src/index.tsx:625-715`); no `touchstart`/`pointerdown`. On tablets/phones — a primary use case for a *remote* control page — tiles can't be moved or resized.

**Desired** — migrate to Pointer Events so mouse and touch share one path (pairs with the drag/resize cancel-semantics bug fix).

_Source: feature/DX analysis_

NilsR0711:
## Problem

The grid's drag-move and resize gestures were mouse-only (`handleGridMouseDown`, `window.addEventListener('mouseup', …)`) — no `touchstart`/`pointerdown` handling. On tablets/phones, a primary use case for a *remote* control page, tiles couldn't be moved or resized.

## Solution

- Migrated the grid's gesture handlers from Mouse Events to Pointer Events (`onPointerDown`, `onPointerMove`, `onPointerLeave`) so mouse and touch share one code path. `isPrimaryButton`/`resolveMoveTarget` in `gestures.ts` are event-shape agnostic and needed no changes.
- The window-level `mouseup` listeners for ending a move/resize now also listen for `pointercancel`, aborting the gesture without committing when a system gesture interrupts a touch drag.
- Added `computeHoveringIdx`, a pure bounds-checked helper for resolving the hovered cell from raw coordinates. Mouse hover tracking relied on `mouseleave` to clear the hovered cell once the pointer left the grid mid-drag; a touch pointer is implicitly captured to its `pointerdown` target and never dispatches boundary events while dragging, so this reproduces the "off grid" signal directly from coordinates instead — without it, a touch drag released outside the grid could commit against whatever cell the out-of-bounds coordinates happened to floor to.
- Added `touch-action: none` to the drag/resize surfaces so a finger-drag isn't hijacked by the browser's default touch scrolling/panning.
- Kept the resize handles visible without `:hover` on touch/pen devices (`@media (hover: none), (pointer: coarse)`), since touch has no hover state to reveal them — without this the handles would stay invisible (`opacity: 0`) and be impossible to grab on a touchscreen even after the event-handling migration.

### Architecture decisions

- Kept the pure gesture-decision logic (`isPrimaryButton`, `resolveMoveTarget`) untouched in `gestures.ts` since they already operate on primitives (button codes, coordinates), not event objects — no coupling to Mouse vs Pointer event types.
- Followed the existing pattern from #135/#136 (the drag/resize cancel-semantics fix this issue explicitly pairs with) of keeping gesture *decisions* as pure, independently-unit-tested helpers in `gestures.ts`, separate from the DOM-wiring in `index.tsx`.

## Testing

- `gestures.test.ts`: exhaustive coverage of the new `computeHoveringIdx` helper (in-bounds corners/middle, and every out-of-bounds edge — this is the regression case for the touch capture behavior described above).
- `GridInput.test.tsx` (new): confirms the grid tile responds to `pointerdown` and no longer requires `mousedown` — i.e. the actual DOM wiring, not just the pure logic.
- Full quality gate green locally: `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm test` (all workspaces), and `npm -w streamwall-control-client run build`.
- **Not tested**: live touchscreen/tablet interaction. This repo's own README notes visual/interactive verification isn't possible in this headless environment (no display, and the Electron app + control-server + a live stream uplink would all need to be running together to exercise the real grid). The pointer-event semantics themselves (capture behavior, `pointercancel`, bounds-checking) are backed by the unit tests above, which encode the specific touch-vs-mouse differences that would otherwise cause regressions.

## Risks / breaking changes

None expected — mouse behavior is unchanged (Pointer Events for mouse fire with the same semantics as the Mouse Events they replace), and this is purely additive for touch/pen input.

Closes #80

### DIFF
--- a/packages/streamwall-control-ui/src/GridInput.test.tsx
+++ b/packages/streamwall-control-ui/src/GridInput.test.tsx
@@ -1,0 +1,79 @@
+import Color from 'color'
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+
+// `idColor` (from streamwall-shared) returns a `Color` instance built from
+// *that* package's own `color` module copy — the monorepo doesn't dedupe it
+// against this package's copy (see packages/streamwall-shared/node_modules/color
+// vs packages/streamwall-control-ui/node_modules/color in package-lock.json).
+// `GridInput`'s styled component then re-wraps it via `Color($color)...` using
+// *this* package's copy, which can't parse an instance from the other copy.
+// That's an unrelated, pre-existing cross-package bug (tracked separately);
+// stub `idColor` here with a same-copy `Color` instance so this test isolates
+// the pointer-event wiring this file is actually about.
+vi.mock('streamwall-shared', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('streamwall-shared')>()
+  return { ...actual, idColor: () => Color('white') }
+})
+
+import { GridInput } from './index.tsx'
+
+let container: HTMLDivElement | undefined
+
+afterEach(() => {
+  if (container) {
+    act(() => render(null, container!))
+    container.remove()
+    container = undefined
+  }
+})
+
+function renderInput(
+  props: Partial<Parameters<typeof GridInput>[0]> = {},
+): HTMLDivElement {
+  container = document.createElement('div')
+  document.body.appendChild(container)
+  act(() => {
+    render(
+      <GridInput
+        style={{}}
+        idx={0}
+        onChangeSpace={() => {}}
+        spaceValue=""
+        isHighlighted={false}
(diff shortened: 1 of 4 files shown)


NilsR0711:
Filed a follow-up for a pre-existing, unrelated bug discovered while writing the `GridInput` component test in this PR: #196 (the `color` npm package is duplicated across workspaces, so `Color` instances from `idColor()` in `streamwall-shared` can't be re-wrapped by `streamwall-control-ui`'s own copy).

PR #195, opened 2026-07-14
feat: cycle a grid cell through a stream playlist on an interval

### LINKED ISSUE #77: Feature: add stream rotation / playlist cycling

No cycling/playlist logic exists; `rotate-stream` (`index.ts:379-383`) is video *orientation*, and the Twitch vote (`TwitchBot.ts:52-55`) only switches the listening slot.

**Desired** — an optional per-cell or per-wall playlist that cycles a set of streams on an interval.

_Source: feature/DX analysis_

NilsR0711:
## Problem

There was no cycling/playlist logic in Streamwall. `rotate-stream` only controls video *orientation* (image rotation), and the Twitch chat vote only switches which existing cell's audio is listened to — neither changes which stream occupies a grid cell over time.

## Solution

Adds an optional, config-driven playlist feature: any grid cell can be configured to cycle through a fixed list of stream URLs on an interval.

- `[[playlist]]` config table (`view`, `interval`, `urls`), validated in `config.ts`:
  - `view` must be within the configured grid (`grid.cols * grid.rows`).
  - `urls` must be non-empty.
  - `interval` must be positive.
  - Two entries can't target the same view.
- New `PlaylistScheduler` (`packages/streamwall/src/main/playlist.ts`) advances each configured view on its own timer, resolving each URL to its current stream ID at the moment of the advance (not once at startup), so it keeps working as data sources refresh. A URL that doesn't currently resolve to a known stream is skipped with a console warning rather than crashing, and the cursor still advances so it isn't stuck retrying the same URL forever.
- Wired into `index.ts`: the scheduler writes into the same Yjs `viewsState` that manual drag-to-place already uses, so it flows through the existing `updateViewsFromStateDoc → StreamWindow.setViews → DISPLAY` pipeline with no new rendering plumbing.
- Documented in `README.md` and `example.config.toml`.

### Architecture decisions

- **Config-driven, not a new `ControlCommand`.** The playlist definition is static per-deployment (like `twitch.vote.interval`), so it lives in `config.toml` rather than adding new remote-control surface. This keeps the change scoped to the issue's "optional per-cell playlist" ask without expanding the control-UI/uplink attack surface.
- **Writes through `viewsState`, not a parallel state store.** Reuses the exact mechanism `handleSetView`/drag-and-drop already use, so no changes were needed to `StreamWindow` or the view state machine.
- **Known limitation, not addressed here:** the view state machine (`viewStateMachine.ts`) has no partial "swap content while running" transition — any content change (including a playlist advance) forces a full reload of that cell's view. This matches how manual stream reassignment already behaves today, so it's not a regression, but a future enhancement could add a lighter transition to avoid the reload. Filed as a follow-up issue (see below).

## Testing performed

- New `playlist.test.ts` (Vitest, fake timers): initial assignment on start, cyclic wraparound, independent per-view intervals, skip-and-continue on an unresolvable URL, `stop()` clearing timers, and a no-op empty configuration.
- Extended `config.test.ts`: accepts valid playlist config, rejects empty `urls`, rejects non-positive `interval`, rejects an out-of-grid `view`, rejects duplicate `view` targets.
- Full quality gate green locally: `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm run test` (all workspaces).

## Risks / breaking changes

None — the feature is entirely opt-in (no `[[playlist]]` entries means no behavior change) and adds no new remote-control command surface.

Closes #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Under the hood, think of Streamwall as a specialized web browser for mosaicing v
 - **Remote control with roles** — an optional web-based control server lets operators run the wall from a browser, with **admin**, **operator**, and **monitor** roles gated by invite links.
 - **Automatic recovery** — failed or stalled stream loads are retried automatically with exponential backoff, with the failure surfaced on the wall and in the control UI.
 - **Flexible data sources** — load streams from JSON APIs, TOML files, or add them ad hoc (including `overlay`/`background` kinds for widgets and chroma-key layers).
+- **Playlists** — optionally cycle a grid cell through a fixed list of stream URLs on an interval.
 - **Twitch chat bot** — an optional bot posts templated announcements and runs viewer polls in a Twitch channel's chat.
 
 ![Control UI](docs/images/control-ui.png)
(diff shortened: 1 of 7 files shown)


NilsR0711:
Follow-up filed for the known reload limitation noted in the PR description: #198

PR #197, opened 2026-07-14
feat: add undo/redo for grid edits via Y.UndoManager

### LINKED ISSUE #79: Feature: add undo/redo for grid edits (Y.UndoManager)

NilsR0711:
## Problem

No undo/redo existed for grid edits. Drag-move, resize handles, and grid-size
changes were all one-way: a mis-drag or an accidental grid shrink could lose
layout state irrecoverably. Grid shrinks in particular are destructive by
design - `remapGridAssignments` permanently drops any assignment whose (x, y)
no longer fits the new grid.

## Solution

- Added `createSharedUndoManager` (`packages/streamwall-control-ui/src/yUndo.ts`),
  a small wrapper around `Y.UndoManager` scoped to the shared `views` doc.
  By default `Y.UndoManager` only tracks locally-originated transactions, so it
  accepts an optional `remoteOrigin` string to also track remote-applied
  updates sharing that origin.
- `useYDoc` now builds this undo manager for its doc and exposes it on
  `StreamwallConnection`.
- `ControlUI` binds `mod+z` / `mod+shift+z` (Ctrl+Z / Ctrl+Shift+Z on
  Windows/Linux, Cmd+Z / Cmd+Shift+Z on macOS) to `undo()`/`redo()`, following
  the existing `react-hotkeys-hook` pattern used by the other shortcuts.
  `enableOnFormTags` defaults to false, so this doesn't hijack native
  undo/redo while a grid-size input or other text field is focused.
- Each of the three connections (`streamwall-control-client`'s websocket
  client, `streamwall`'s Electron IPC renderer, and the dev mock harness)
  now passes its own remote-update origin string (`'server'`, `'app'`, and
  none, respectively) into `useYDoc`.

### Architecture decision: why the grid-shrink case needed `remoteOrigin`

Drag-move and swap mutate the client's own `stateDoc` directly (local
origin), so a plain `Y.UndoManager` already covers them. A grid-size change
does not: the client sends a `set-grid-size` command over the websocket, the
*main process* applies `applyGridResize` → `remapGridAssignments` on its own
authoritative doc, and only the resulting Yjs update reaches the client - as
a remote-origin update (`Y.applyUpdate(stateDoc, update, 'server'|'app')`).
Without tracking that origin, Ctrl+Z would silently do nothing for the one
case the issue explicitly calls out as needing protection. Tracking the
connection's own remote-update origin (rather than, say, all origins) keeps
undo scoped to changes this client's Streamwall connection actually applied.

## Testing performed

- New unit tests (`packages/streamwall-control-ui/src/yUndo.test.ts`),
  written test-first:
  - undoes/redoes a plain local edit (drag-move-style mutation)
  - ignores transactions from an untracked, unrelated origin
  - restores assignments dropped by a destructive grid-shrink relayed from
    the server, using the real `remapGridAssignments` and the same
    `encodeStateAsUpdate`/`applyUpdate(..., 'server')` relay pattern the
    production code uses - this is the regression test for the issue's
    specific concern
- Full workspace test suite (`npm run test --workspaces`): all packages
  green, including the new tests (control-ui: 33/33)
- `npm run typecheck --workspaces`: clean
- `eslint` / `prettier --check` on all touched files: clean
- `npm run build` (control-client): succeeds
- Manual verification in the browser via the control-client dev harness
  (`vite --config packages/streamwall-control-client/vite.config.ts`,
  `/dev.html`): dragged a stream off the wall, confirmed the cell emptied,
  pressed Ctrl+Z, confirmed the stream reappeared in its original cell.
  Also confirmed Ctrl+Z does not interfere while a grid-size text input is
  focused.

## Risks / breaking changes

None expected. `StreamwallConnection.undoManager` is a new optional field;
existing consumers are unaffected. No config or persisted-data format
changes.

Closes #79

### DIFF
--- a/packages/streamwall-control-client/src/dev.tsx
+++ b/packages/streamwall-control-client/src/dev.tsx
@@ -166,9 +166,11 @@ const demoState: StreamwallState = {
 }
 
 function useMockConnection(): StreamwallConnection {
-  const { docValue: sharedState, doc: stateDoc } = useYDoc<CollabData>([
-    'views',
-  ])
+  const {
+    docValue: sharedState,
+    doc: stateDoc,
+    undoManager,
+  } = useYDoc<CollabData>(['views'])
   const appState = useStreamwallState(demoState)
 
   useEffect(() => {
@@ -191,6 +193,7 @@ function useMockConnection(): StreamwallConnection {
     send: (msg) => console.debug('[mock send]', msg),
     sharedState,
     stateDoc,
+    undoManager,
   }
 }
 
--- a/packages/streamwall-control-client/src/index.tsx
+++ b/packages/streamwall-control-client/src/index.tsx
@@ -29,7 +29,8 @@ function useStreamwallWebsocketConnection(
     docValue: sharedState,
     doc: stateDoc,
     setDoc: setStateDoc,
-  } = useYDoc<CollabData>(['views'])
+    undoManager,
+  } = useYDoc<CollabData>(['views'], 'server')
   const [streamwallState, setStreamwallState] = useState<StreamwallState>()
   const appState = useStreamwallState(streamwallState)
 
@@ -133,6 +134,7 @@ function useStreamwallWebsocketConnection(
     send,
     sharedState,
     stateDoc,
+    undoManager,
   }
 }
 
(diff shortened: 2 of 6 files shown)


NilsR0711:
Follow-up (out of scope for this PR): filed #200 for a German confirm() string spotted in the adjacent grid-shrink guard code.

PR #202, opened 2026-07-14
fix(build): drop the unresolved Plugin[] cast in Vite configs

### LINKED ISSUE #64: Resolve the Plugin[] cast workaround in the Vite configs
**Priority: low**

**Evidence** — `packages/streamwall/vite.renderer.config.ts:28` and `packages/streamwall-control-client/vite.config.ts:22` both `...(preact(...) as Plugin[])` with a FIXME; `Plugin` is never imported, so once typecheck runs these error (`Cannot find name 'Plugin'`).

**Fix** — `import type { Plugin } from 'vite'` (or drop the cast once `@preact/preset-vite` types are fixed) so the files pass the typecheck gate.

_Source: CI/build analysis_

NilsR0711:
## Problem

`packages/streamwall/vite.renderer.config.ts` and `packages/streamwall-control-client/vite.config.ts` both cast the `preact()` plugin result with `as Plugin[]`, but `Plugin` is never imported anywhere in either file.

Because the `DOM` lib is in scope, the bare `Plugin` identifier silently resolves to the unrelated legacy `navigator.plugins` `Plugin` interface (the old Netscape plugin API) instead of Vite's `Plugin` type. The assertion therefore type-checks by accident, not because it does what the FIXME comment claims ("working around TS error: `Type 'Plugin<any>' is not assignable to type 'PluginOption'`").

## Solution

Verified with the pinned `@preact/preset-vite@^2.10.5` / `vite@^8.1.3` versions that the original type mismatch the cast worked around no longer exists — `preact({ devToolsEnabled: false })` spreads cleanly into `plugins: []` with **no cast at all**, which is the alternative the FIXME comment itself suggested ("or drop the cast once `@preact/preset-vite` types are fixed").

Dropped the cast (and the now-stale FIXME comment) in both files; kept the `devToolsEnabled: false` comment explaining the zimmerframe/ESM workaround, since that's still relevant.

## Testing performed

No behavior change — this is a build-config type-safety fix, so no new tests were added (per the TDD exception for changes with no testable runtime behavior). Verified instead via the full quality gate:

- `npm run format:check` — clean
- `npm run lint` — clean
- `npm run typecheck --workspaces` — clean (confirmed the cast is genuinely unnecessary by probing both packages directly with `tsc`)
- `npm run test --workspaces` — 73 + 40 tests passing
- `npm -w streamwall-control-client run build` — succeeds (same as CI)
- `npm -w streamwall run package` — Electron Forge packages successfully, confirming the renderer's Vite build (using the modified config) still works end-to-end

## Risks / breaking changes

None. Pure type-level cleanup; the `plugins` array's runtime contents are unchanged.

Closes #64

### DIFF
--- a/packages/streamwall-control-client/vite.config.ts
+++ b/packages/streamwall-control-client/vite.config.ts
@@ -19,9 +19,8 @@ export default defineConfig({
   },
 
   plugins: [
-    // FIXME: working around TS error: "Type 'Plugin<any>' is not assignable to type 'PluginOption'"
     // devToolsEnabled: false avoids the preact:transform-hook-names plugin, which
     // crashes under newer Vite/Node (zimmerframe is ESM-only, loaded via require()).
-    ...(preact({ devToolsEnabled: false }) as Plugin[]),
+    ...preact({ devToolsEnabled: false }),
   ],
 })
--- a/packages/streamwall/vite.renderer.config.ts
+++ b/packages/streamwall/vite.renderer.config.ts
@@ -25,10 +25,9 @@ export default defineConfig({
   },
 
   plugins: [
-    // FIXME: working around TS error: "Type 'Plugin<any>' is not assignable to type 'PluginOption'"
     // devToolsEnabled: false disables the preact:transform-hook-names plugin, which
     // crashes under newer Vite/Node because zimmerframe@1.1.4 is ESM-only (no CJS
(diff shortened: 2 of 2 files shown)


NilsR0711:
Cursor BugBot review did not run within ~15 minutes of the last push (no comments/reviews posted) — treating it as unavailable and proceeding with CI green (all checks passed) per the review-gate policy.

PR #204, opened 2026-07-14
fix(renderer): inline the loading spinner instead of svg-loaders-react

### LINKED ISSUE #182: svg-loaders-react still can't resolve preact/compat under the vitest test stack
## Context

#177 fixed `react-icons` and `styled-components` resolving the real `react` package instead of `preact/compat` under this repo's `vitest` + `preact` + `happy-dom` component test stack (PR #181), by preferring their ESM/browser builds via `resolve.mainFields` and forcing them through Vite's transform pipeline via `test.server.deps.inline`, so their `import ... from 'react'` specifiers resolve through the `react` -> `preact/compat` alias.

That fix doesn't reach `svg-loaders-react` (used via `styled(TailSpin)` in `packages/streamwall/src/renderer/OverlayViewTile.tsx`, for the stream-loading spinner): the package ships **CJS only**, with no ESM/browser build for Vite's resolver to route through the alias. Its internal `require('react')` is resolved as an opaque external module by vitest's SSR-like module runner, bypassing `resolve.alias` entirely, so it always loads the real `react` package.

Rendering it directly (or via `styled(TailSpin)`, once styled-components' own tag resolution was fixed by #181) throws the same error #177 described for `react-icons`:

```
TypeError: Cannot add property __, object is not extensible
    at type ../../node_modules/preact/src/diff/children.js:219:15
```

## Current workaround

`packages/streamwall/src/renderer/OverlayViewTile.test.tsx` mocks it out:

```ts
vi.mock('svg-loaders-react', () => ({
  TailSpin: () => null,
}))
```

This means `OverlayViewTile`'s loading-spinner rendering is never actually exercised by a component test.

## Suggested solution

- Replace `svg-loaders-react` with an alternative that ships an ESM build (or a small first-party inline SVG spinner), removing the dependency and the mock entirely, or
- If the dependency must stay, investigate whether `optimizeDeps.include` / forcing esbuild pre-bundling of `svg-loaders-react` resolves the alias at the pre-bundle step (unlike `test.server.deps.inline`, which does not rewrite `require()` calls in genuine CommonJS source), or
- If unfixable within a reasonable window, keep documenting the mock as the accepted workaround (as done in #181).

## Where to look

- `packages/streamwall/vitest.config.ts` (see the comment on `mainFields`/`deps.inline` added in #181, which explains why this package is excluded from the fix)
- `packages/streamwall/src/renderer/OverlayViewTile.tsx` (`LoadingSpinner = styled(TailSpin)`)
- `packages/streamwall/src/renderer/OverlayViewTile.test.tsx` (the `vi.mock('svg-loaders-react', ...)` workaround)

NilsR0711:
## Problem

`svg-loaders-react` (used via `styled(TailSpin)` in `OverlayViewTile.tsx` for the stream-loading spinner) ships CJS only, with no ESM/browser build. Under this repo's `vitest` + `preact` + `happy-dom` component test stack, its internal `require('react')` bypasses `resolve.alias` and always resolves the real `react` package instead of `preact/compat`, throwing:

```
TypeError: Cannot add property __, object is not extensible
    at type ../../node_modules/preact/src/diff/children.js:219:15
```

This is the same crash #177/#181 fixed for `react-icons` and `styled-components`, but their fix (preferring ESM builds via `mainFields` + forcing both through Vite's transform pipeline via `test.server.deps.inline`) doesn't reach `svg-loaders-react` since it has no ESM build to route through the alias. `OverlayViewTile.test.tsx` worked around this with `vi.mock('svg-loaders-react', ...)`, which meant the loading spinner's own rendering was never actually exercised by a component test.

## Solution

Inlined `TailSpin` as a first-party component (`packages/streamwall/src/renderer/TailSpin.tsx`), copied from `svg-loaders-react`'s MIT-licensed source (Copyright Sam Herbert / Adam Wanninger), instead of extending the `mainFields`/`deps.inline` workaround to a third CJS-only dependency:

- Removed the `svg-loaders-react` dependency and its ambient type declaration.
- `OverlayViewTile.tsx` now imports `TailSpin` from the local file — visuals are unchanged.
- Removed the `vi.mock('svg-loaders-react', ...)` workaround from `OverlayViewTile.test.tsx` and added real coverage for the spinner's visibility toggling (`isLoading` true/false), asserting on computed `opacity`/`visibility` instead of just presence.
- Updated the `vitest.config.ts` comment that documented the workaround.

## Testing

- `npm test` (all workspaces) — 155 streamwall tests + full monorepo suite pass, including two new tests exercising the previously-mocked loading spinner.
- `npm run typecheck`, `npx eslint .`, `npm run format:check` — all clean.
- `electron-forge package` — production build succeeds with the dependency removed.

## Risks / breaking changes

None. Visual output of the spinner is unchanged (same SVG markup/animation); this only affects how the spinner is imported and tested.

Closes #182

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12991,11 +12991,6 @@
         "url": "https://github.com/sponsors/ljharb"
       }
     },
-    "node_modules/svg-loaders-react": {
-      "version": "3.1.1",
-      "resolved": "https://registry.npmjs.org/svg-loaders-react/-/svg-loaders-react-3.1.1.tgz",
-      "integrity": "sha512-zREh+e1S7blwCAW14i8j0eorC6Jmm0jC0BqmdxFfQxoV/hkrpxsogXons+FK6IUdOGtdN1IM3IBtT1mOvcPnSg=="
-    },
     "node_modules/tapable": {
       "version": "2.3.3",
       "resolved": "https://registry.npmjs.org/tapable/-/tapable-2.3.3.tgz",
@@ -14883,7 +14878,6 @@
         "reconnecting-websocket": "^4.4.0",
         "source-map-support": "^0.5.21",
         "styled-components": "^6.4.3",
-        "svg-loaders-react": "^3.1.1",
         "update-electron-app": "^3.3.0",
         "utf-8-validate": "^6.0.6",
         "ws": "^8.21.0",
--- a/packages/streamwall/package.json
+++ b/packages/streamwall/package.json
@@ -63,7 +63,6 @@
     "reconnecting-websocket": "^4.4.0",
     "source-map-support": "^0.5.21",
     "styled-components": "^6.4.3",
-    "svg-loaders-react": "^3.1.1",
     "update-electron-app": "^3.3.0",
     "utf-8-validate": "^6.0.6",
     "ws": "^8.21.0",
--- a/packages/streamwall/src/renderer/OverlayViewTile.test.tsx
+++ b/packages/streamwall/src/renderer/OverlayViewTile.test.tsx
@@ -2,18 +2,8 @@
 import { render } from 'preact'
 import { act } from 'preact/test-utils'
 import type { StreamData } from 'streamwall-shared'
-import { afterEach, describe, expect, test, vi } from 'vitest'
+import { afterEach, describe, expect, test } from 'vitest'
 import { OverlayViewTile } from './OverlayViewTile'
-
-// Unlike react-icons and styled-components (see vitest.config.ts),
-// svg-loaders-react ships CJS only, so it has no ESM build for Vite's
-// resolver to route through the `react` -> `preact/compat` alias - it always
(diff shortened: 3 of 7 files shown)


NilsR0711:
No Cursor BugBot review appeared within ~5.5 minutes of the last push (HEAD 5efd186). Treating the AI-review requirement as satisfied per the unavailable-review fallback; all other merge-readiness checks (CI, tests, lint, typecheck, build) are green.

NilsR0711:
Filed a follow-up: #212 — the inlined `TailSpin`'s gradient uses a static (non-unique) SVG `id`, which can collide across multiple simultaneously-loading grid tiles. Pre-existing behavior from the original `svg-loaders-react` source (not a regression from this PR), but now first-party code so worth tracking separately rather than expanding this PR's scope.

PR #205, opened 2026-07-14
fix(electron): quit when either top-level window closes

### LINKED ISSUE #193: DX: closing the Control Window doesn't quit the app (no window-all-closed handler)
**Type: dx · Priority: low**

While fixing #88 (stream window close/quit behavior), noticed a related gap: `ControlWindow` emits a `close` event (`ControlWindow.ts:31`, `this.win.on('close', () => this.emit('close'))`), but `main/index.ts` never listens for it. Closing just the control window currently does nothing beyond destroying that window -- the app keeps running with no obvious way to quit other than closing the (possibly hidden, on macOS) stream window.

There is also no `app.on('window-all-closed', ...)` handler, so the standard Electron convention of quitting once every window is closed (non-macOS) isn't implemented at the app level; the app currently only quits by reacting to the stream window specifically.

**Desired** — decide and implement one of:
- Wire `controlWindow.on('close', ...)` to the same quit path as the stream window (so closing either window quits, mirroring apps with multiple top-level windows), and/or
- Add a proper `app.on('window-all-closed', ...)` handler (quit on non-macOS when all windows are gone) so quitting isn't tied to one specific window.

_Source: follow-up from #88 (PR #191)_

NilsR0711:
## Problem

`ControlWindow` emits a `close` event (`ControlWindow.ts:31`), but `main/index.ts` never listened for it. Closing just the control window did nothing beyond destroying that window — the app kept running with no obvious way to quit other than closing the (possibly hidden, on macOS) stream window.

There was also no `app.on('window-all-closed', ...)` handler, so the standard Electron convention of quitting once every window is closed (non-macOS) wasn't implemented at the app level.

## Solution

- `ControlWindow` now forwards the underlying Electron `close` event (previously it emitted a bare `close` with no event object), matching the pattern `StreamWindow` already used. This lets listeners call `event.preventDefault()`.
- `main/index.ts` extracts the existing quit-or-hide logic (used for the stream window) into a shared `handleWindowClose(win, event)` helper and wires the control window's `close` event through it too. Closing either window now quits the app (non-macOS) or hides it (macOS, mirroring the stream window's convention), and `app.on('activate', ...)` re-shows both windows.
- Added `app.on('window-all-closed', ...)` as a standard-Electron-convention safety net: quits on non-macOS if every window is ever closed without already being wired into `handleWindowClose` (e.g. a future window).
- New pure `shouldQuitOnAllWindowsClosed(platform)` predicate in `windowCloseBehavior.ts`, alongside the existing `shouldHideInsteadOfQuit`, unit-tested the same way.

## Architecture decisions

- Reused the existing `shouldHideInsteadOfQuit` convention (hide on macOS, quit elsewhere) for the control window rather than inventing new behavior, so both top-level windows behave consistently.
- Followed the repo's existing testing pattern for Electron main-process code: extract pure, platform-driven predicates into `windowCloseBehavior.ts` and unit test those directly; the actual Electron wiring in `index.ts` mirrors the untested pattern already used for the stream window's close handling (also true of the merged #191), since it requires a live Electron runtime to exercise meaningfully.
- Added a `ControlWindow.test.ts` (none existed before) that models `BrowserWindow` as an `EventEmitter`, to directly cover the new close-event forwarding — this is real, previously-untested behavior, not just wiring.

## Testing performed

- `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm run test` (all workspaces) — all green.
- New/updated tests: `windowCloseBehavior.test.ts` (`shouldQuitOnAllWindowsClosed`), `ControlWindow.test.ts` (close-event forwarding).

## Risks / breaking changes

None expected. Closing the control window now quits the app on Windows/Linux instead of silently doing nothing; this is the intended fix and matches how closing the stream window already behaves.

Closes #193

### DIFF
--- a/packages/streamwall/src/main/ControlWindow.test.ts
+++ b/packages/streamwall/src/main/ControlWindow.test.ts
@@ -1,0 +1,33 @@
+import EventEmitter from 'events'
+import { describe, expect, it, vi } from 'vitest'
+
+// ControlWindow pulls in Electron directly and via ./loadHTML. Model
+// BrowserWindow as a small EventEmitter so `win.on('close', ...)` wiring can
+// be exercised the same way the real window would drive it, without an
+// Electron runtime.
+class FakeBrowserWindow extends EventEmitter {
+  webContents = { send: vi.fn() }
+  removeMenu = vi.fn()
+}
+
+vi.mock('electron', () => ({
+  BrowserWindow: FakeBrowserWindow,
+  ipcMain: { handle: vi.fn() },
+}))
+
+vi.mock('./loadHTML', () => ({ loadHTML: vi.fn() }))
+
+const { default: ControlWindow } = await import('./ControlWindow')
+
+describe('ControlWindow close', () => {
+  it('forwards the underlying Electron close event so callers can preventDefault', () => {
+    const controlWindow = new ControlWindow()
+    const closeListener = vi.fn()
+    controlWindow.on('close', closeListener)
+
+    const fakeEvent = { preventDefault: vi.fn() }
+    controlWindow.win.emit('close', fakeEvent)
+
+    expect(closeListener).toHaveBeenCalledWith(fakeEvent)
+  })
+})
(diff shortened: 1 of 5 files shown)


NilsR0711:
Follow-up: #214 -- ControlWindow's `closable: false` may prevent the close-to-quit path added here from being reachable via the window's native close button under normal user interaction.

PR #210, opened 2026-07-14
test(control-server): cover WebSocket multiplexing and authorization gating

### LINKED ISSUE #52: Test control-server WebSocket multiplexing and message-authorization gating
**Priority: medium**

`packages/streamwall-control-server/src/index.ts` (530 lines) bridges one Streamwall socket to N clients and is the internet-facing trust boundary.

**Scope (integration, Vitest + real `ws`)** — start on port 0; connect a fake Streamwall socket + client sockets with role cookies; assert monitor/operator cannot `create-invite`/`browse` (dropped + logged), operator can `set-listening-view`, admin's broadcast includes `auth` while operator's does not, a second Streamwall connection is rejected, queued messages during async auth aren't dropped. Unit-test `queueWebSocketMessages` in isolation.

_Source: test analysis_

NilsR0711:
## Problem

`packages/streamwall-control-server/src/index.ts` bridges one Streamwall socket to N clients and is the internet-facing trust boundary, but its WebSocket multiplexing and role-based message-authorization gating had no dedicated test coverage.

## Solution

Two new test files, both using the existing `buildTestApp`/real-`ws` integration pattern already established in this package:

**`multiplexing.test.ts`** — boots a live control-server instance with a real Streamwall uplink and role-cookied clients, and asserts:
- an operator/monitor cannot `create-invite` or `browse`: the command is dropped, logged (`console.warn`), and never forwarded to the Streamwall uplink
- an operator *can* `set-listening-view`: it is forwarded as expected
- an admin's state broadcast includes the `auth` state; an operator's does not
- a second Streamwall connection is rejected while the first stays connected (and the rejection is logged)
- a state message sent immediately on connect is not dropped while the uplink's async token validation (scrypt) is still pending

**`queueWebSocketMessages.test.ts`** — unit-tests the message-queueing helper in isolation against a fake `EventEmitter`-based socket: buffering before a handler is attached, immediate delivery once attached, ordering across mixed queued/live delivery, and clearing the queue on close so a later handler never replays stale messages.

`queueWebSocketMessages` was exported (previously module-private) so it could be unit-tested directly, per the issue's request. No behavior change.

## Testing performed

- New tests pass locally (control-server: 85/85, full repo: 436/436)
- Sanity-checked the authorization tests against a mutated build (temporarily disabling the `roleCan` gate) to confirm they fail when the guarded behavior is broken, then restored
- `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm test` (all workspaces), and `npm -w streamwall-control-client run build` all pass

## Risks / breaking changes

None — test-only change plus a single `export` keyword addition with no behavioral effect.

Closes #52

### DIFF
--- a/packages/streamwall-control-server/src/index.ts
+++ b/packages/streamwall-control-server/src/index.ts
@@ -243,7 +243,7 @@ declare module 'fastify' {
  * This is useful for async validation of the connection before handling messages,
  * because awaiting before adding a message event listener can drop messages.
  */
-function queueWebSocketMessages(ws: WebSocket) {
+export function queueWebSocketMessages(ws: WebSocket) {
   let queue: WebSocket.Data[] = []
   let messageHandler: ((rawData: WebSocket.Data) => void) | null = null
 
--- a/packages/streamwall-control-server/src/multiplexing.test.ts
+++ b/packages/streamwall-control-server/src/multiplexing.test.ts
@@ -1,0 +1,458 @@
+import assert from 'node:assert/strict'
+import { once } from 'node:events'
+import type { AddressInfo } from 'node:net'
+import { after, test } from 'node:test'
+import { setTimeout as delay } from 'node:timers/promises'
+import WebSocket from 'ws'
+
+import { buildTestApp } from './testHelpers.ts'
+
+const BASE_URL = 'http://localhost:3000'
+
+const VALID_STATE = {
+  identity: { role: 'admin' },
+  config: {
+    cols: 3,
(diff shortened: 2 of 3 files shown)


NilsR0711:
Follow-up (out of scope for this PR): #216 — consolidate the duplicated WebSocket test helpers (`recordJsonMessages`, `messageCollector`, uplink/client boot helpers) that are copy-pasted across this package's test files, now including `multiplexing.test.ts`.

Issue #219, opened 2026-07-14
fix(ci): main branch typecheck is broken (stateDiff.test.ts fixture missing layoutPresets)


NilsR0711:
**Type: bug · Priority: high**

`npm run typecheck` fails on `main` as of #199 ("feat: add named layout presets"), which made `StreamwallState.layoutPresets` a required field. This breaks the `Static checks` CI job (and the required `CI OK` gate) for every subsequent PR rebased onto `main`.

**Root cause**

`packages/streamwall-shared/src/stateDiff.test.ts`'s `makeState()` test helper builds its base `StreamwallState` object literal without a `layoutPresets` field, relying entirely on the (optional) `overrides` parameter to supply it:

```ts
function makeState(overrides: Partial<StreamwallState> = {}): StreamwallState {
  return {
    identity: { role: 'admin' },
    config: { /* ... */ },
    streams: [...],
    customStreams: [],
    views: [...],
    streamdelay: null,
    ...overrides, // layoutPresets is never guaranteed here
  }
}
```

Since `layoutPresets: LayoutPreset[]` is required (not optional) on `StreamwallState`, TypeScript correctly flags this as a type error:

```
src/stateDiff.test.ts(34,3): error TS2322: Type '{ ...; layoutPresets?: LayoutPreset[] | undefined; }' is not assignable to type 'StreamwallState'.
  Types of property 'layoutPresets' are incompatible.
    Type 'LayoutPreset[] | undefined' is not assignable to type 'LayoutPreset[]'.
      Type 'undefined' is not assignable to type 'LayoutPreset[]'.
```

**Impact**

Confirmed via `gh run view` on the `main` CI run for commit `6f2b5c8` (the layout-presets merge commit): the `Static checks` job fails, which fails the required `CI OK` status check. Any PR that rebases onto current `main` inherits this failure and cannot pass the branch-protection gate.

**Suggested fix**

Add a default `layoutPresets: []` to `makeState()`'s base object literal in `stateDiff.test.ts`, mirroring how the other required fields (`streamdelay: null`, etc.) are defaulted.

**Note:** this was already fixed as part of #76's PR (which also needed a similar default for a different, unrelated required field it introduced on `ViewState.context`), since it otherwise blocked that PR's own CI. Filing this issue for visibility/tracking in case another main-line fix lands first.

NilsR0711:
[...]

PR #199, opened 2026-07-14
feat: add named layout presets (save/load)

NilsR0711:
## Summary

Closes #78.

Adds named layout presets: save the current grid (dimensions + per-cell stream assignments) under a name, and recall it in one click — rendered next to the `GRID_PRESETS` size buttons in the control UI header, per the issue's requested placement.

## Technical solution

- **streamwall-shared**: new `LayoutPreset` type and `StreamwallState.layoutPresets`; three new `ControlCommand` variants (`save-layout-preset`, `load-layout-preset`, `delete-layout-preset`) with bounded Zod schemas; added to the `operator` permission tier (same as `set-grid-size`) and to the uplink allowlist.
- **streamwall (main process)**: `layoutPresets.ts` provides pure, unit-testable helpers — `buildLayoutPreset` captures the current grid from `viewsState`, `applyLayoutPreset` rewrites `viewsState` to match a saved preset (mirroring `applyGridResize`'s critical ordering: grid size is set *before* the `stateDoc.transact` so the synchronous `observeDeep` layout observer sees the new dimensions — see the ordering note in `gridResize.ts`, issue #15), and `addLayoutPreset` bounds the saved list to `MAX_LAYOUT_PRESETS` (50) to prevent unbounded storage growth. Presets persist via the existing lowdb `storage.ts`.
- **streamwall-control-server**: `StateWrapper.view()` now passes `layoutPresets` through to every role's state view (read-only visibility; the save/load/delete actions themselves stay `roleCan`-gated).
- **streamwall-control-ui**: new `LayoutPresetControls` component — a name input + save button, and a load/delete button pair per saved preset — gated by `roleCan(role, 'save-layout-preset')`.

### Architecture decisions

- Presets are a plain JSON-serializable shape (`{ id, name, cols, rows, views }`), not a raw Yjs snapshot, so they travel over the existing Zod-validated `ControlCommand` JSON channel and compose with the same `viewsState` mutation helpers already used for grid resize.
- The command's own id field is named `presetId` (not `id`) to avoid colliding with the message envelope's numeric `id` under `controlCommandMessageSchema`'s `z.intersection`.
- No control-server changes were needed for command relaying — `save-layout-preset`/`load-layout-preset`/`delete-layout-preset` fall through the existing generic forward branch exactly like `set-grid-size` does today.

## Testing performed

[...]

PR #220, opened 2026-07-14
feat: add per-tile volume control

### LINKED ISSUE #76: Feature: add per-tile volume control

Audio is binary (`webContents.audioMuted`, `viewStateMachine.ts:65-71`); one listening view plus optional background views, all at full volume. No volume API anywhere.

**Desired** — a per-tile volume slider (e.g. injected `video.volume` via `mediaPreload.ts`) and/or a master volume.

_Source: feature/DX analysis_

NilsR0711:
## Summary

Adds a per-tile volume slider to each grid tile's controls, as requested in #76. Audio was previously binary (`webContents.audioMuted` on/off via the listen/background-listen buttons); this adds a continuous 0-1 level, independent of mute state, so an operator can pre-set or adjust a tile's playback volume.

## Technical solution

- **Shared protocol** (`streamwall-shared`): new `set-view-volume` control command (`{ viewIdx, volume }`, `volume` bounded to `[0, 1]` by its Zod schema), added to the uplink allowlist and to the `operator` role's permitted actions. `ViewState.context` now carries a `volume: number` field so the control UI can reflect the current level.
- **Electron main process**: the view state machine gets a new `SET_VOLUME` event, handled independently of the existing `MUTE`/`UNMUTE`/`BACKGROUND` audio sub-state (so a tile's volume can be set/changed while muted, and takes effect once it's listened to). The value is forwarded to the view's preload over a dedicated `'volume'` IPC channel (mirroring how `rotation` already rides the `'options'` channel) and applied to the acquired `<video>`/`<audio>` element via a new `VolumeController`, which mirrors the existing `RotationController` pattern in `mediaPreload.ts`. `StreamWindow.setViewVolume(viewIdx, volume)` and the `set-view-volume` command handler in `main/index.ts` wire it end to end.
- **Control UI**: `GridControls` renders a native `<input type="range">` (styled via `accent-color` to match the theme) next to the listen button, gated by the operator-only `set-view-volume` role action, sending the new level directly on every `input` event (no debounce — matches the codebase's existing direct-send convention for other tile controls; no debounce/throttle precedent exists for continuous inputs here).

### Architecture decisions

- **Per-tile (per grid index) rather than per-stream (per URL)**: volume follows the tile's `SET_VOLUME`/`sendViewEvent` pattern used by mute/background/blur, not the `OPTIONS`/`getDisplayOptions` pattern used by rotation (which is keyed by stream URL via `overlayStreamData`). This matches the issue's explicit "per-tile" framing.
- **No master volume**: the issue said "per-tile ... and/or a master volume" — implemented the per-tile slider as the more complete, explicitly-detailed option (`video.volume` via `mediaPreload.ts`), which fully satisfies the issue on its own.
- **Extracted `VolumeController`/`clampVolume` into a standalone `volumeController.ts`** module (no Electron/DOM imports) so the volume-application logic is unit-testable without mocking `electron`/`document`, matching the style of a small, focused module rather than adding to the untested `mediaPreload.ts` monolith.

## Testing performed

TDD throughout (red/green per layer): schema/uplink/role tests, `viewStateMachine` event/guard tests, `StreamWindow.setViewVolume`/`emitState` tests, a pure-logic `volumeController` unit suite, and a `GridControls` component test (role gating + slider wiring). Full quality gate green locally: `npm run format:check`, `npm run lint`, `npm run typecheck` (all workspaces), `npm run test` (all workspaces), and the control-client production build.

## Risks / breaking changes

None expected. `ViewState.context.volume` and `set-view-volume` are additive; existing mute/listen/background/blur behavior is unchanged (confirmed by the existing viewStateMachine/StreamWindow test suites still passing).

## Note: pre-existing main breakage fixed here

Rebasing onto current `main` surfaced that `npm run typecheck` was already broken there (unrelated to this PR) — `stateDiff.test.ts`'s `makeState()` fixture predates `layoutPresets` becoming a required `StreamwallState` field (#199) and never supplied a default, so it fails to satisfy the type. That blocked this PR's own CI, so it's fixed here as its own commit (`fix(shared): satisfy StreamwallState's required fields in stateDiff test fixtures`), which also adds a default for the new `volume` field this PR introduces. Filed for tracking/visibility as #219.

## Follow-up issues

- #219 — `main`'s typecheck was broken before this PR (unrelated `layoutPresets` fixture gap); fixed as part of this PR, issue filed for visibility.

Closes #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Under the hood, think of Streamwall as a specialized web browser for mosaicing v
 
 - **Resizable grid** — arrange streams in an NxN grid; resize it at runtime from the control UI (column/row presets or exact counts), no restart required.
 - **Drag-to-place layout** — drag a tile onto another to swap their positions, or drop a stream from the list straight onto a grid cell.
-- **Per-tile audio** — listen to any single tile's audio at a time, switchable with a click or hotkey.
+- **Per-tile audio** — listen to any single tile's audio at a time, switchable with a click or hotkey, with a per-tile volume slider.
 - **Blur/censor** — blur individual tiles, or trigger a wall-wide [Streamdelay](https://github.com/chromakode/streamdelay) censor mode.
 - **Dark mode** — light, dark, or system-matched theme in the control UI.
 - **Remote control with roles** — an optional web-based control server lets operators run the wall from a browser, with **admin**, **operator**, and **monitor** roles gated by invite links.
--- a/packages/streamwall-control-ui/src/GridControls.test.tsx
+++ b/packages/streamwall-control-ui/src/GridControls.test.tsx
@@ -1,0 +1,94 @@
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+import { GridControls } from './index.tsx'
+
+// react-icons renders through preact/compat's Context.Consumer, which
+// currently crashes under this package's happy-dom test environment
+// (unrelated to the controls under test here) - stub the icons out so the
+// component can render.
+vi.mock('react-icons/fa', () => ({
+  FaExchangeAlt: () => null,
+  FaRedoAlt: () => null,
+  FaRegLifeRing: () => null,
+  FaRegWindowMaximize: () => null,
+  FaSyncAlt: () => null,
+  FaVideoSlash: () => null,
+  FaVolumeUp: () => null,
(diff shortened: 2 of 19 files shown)


NilsR0711:
CI is fully green (all required checks passing). No Cursor BugBot review appeared within the observed window after the final green CI run — proceeding to merge per explicit user instruction.

PR #208, opened 2026-07-14
fix(streamwall): sign and notarize macOS/Windows release builds

NilsR0711:
## Problem

`packages/streamwall/forge.config.ts` shipped every packaged build completely
unsigned:

- `packagerConfig` had no `osxSign`/`osxNotarize` options.
- `MakerSquirrel({})` (Windows installer) had no signing certificate.
- The macOS build only produces a `MakerZIP`, and the app depends on
  `update-electron-app` at runtime, which requires a signed app to
  auto-update on macOS.

Unsigned builds also trigger OS-level warnings ("app is damaged"/"unknown
publisher") for end users on both macOS and Windows.

## Solution

Added `packages/streamwall/forge.signing.ts`, which derives the signing
config from environment variables:

- **macOS**: `osxSign`/`osxNotarize`, using an App Store Connect API key for
  `notarytool` (`APPLE_TEAM_ID`, `APPLE_API_KEY`, `APPLE_API_KEY_ID`,
  `APPLE_API_ISSUER`) — the modern, non-interactive notarization method
  (Apple's legacy Apple-ID/app-specific-password flow is deprecated).
- **Windows**: `MakerSquirrel` `certificateFile`/`certificatePassword` from a
  PFX certificate (`WINDOWS_CERTIFICATE_FILE`, `WINDOWS_CERTIFICATE_PASSWORD`).

Both are opt-in and independent: if the relevant variables aren't set, that
platform's build stays unsigned exactly as before — no change for local
contributor builds or until secrets are actually provisioned.

`forge.config.ts` also now flips `OnlyLoadAppFromAsar` and
`EnableEmbeddedAsarIntegrityValidation` on once *any* signing is configured
(mirrors the issue's suggestion): asar integrity validation only means
anything once the binary itself is signed, so it stays off for unsigned
builds.

Documented the required secrets and the interim macOS un-quarantine
workaround (`xattr -cr`) in the README under a new "Building & releasing the
desktop app" section.

### Scope note

The issue mentioned "wire secrets into release.yml", but this repo has no
release CI workflow today — releases are run manually via
`npm run publish`. Building a full signed-release GitHub Actions pipeline is
a materially larger, separate piece of work (matrix builds, cert import
steps, cannot be verified end-to-end without real Apple/Windows
credentials), so I scoped this PR to the config wiring + docs and filed a
follow-up issue for the CI pipeline itself (linked below).

## Testing

- Added `forge.signing.test.ts` (9 cases): both signing configs return
[...]

PR #222, opened 2026-07-14
fix(renderer): generate a per-instance gradient id in TailSpin

### LINKED ISSUE #212: OverlayViewTile's loading spinner gradient uses a duplicate SVG id across grid tiles
## Context

Noticed while implementing #182 (inlining `TailSpin` from `svg-loaders-react` as a first-party component in `packages/streamwall/src/renderer/TailSpin.tsx`), used by `OverlayViewTile.tsx`'s `LoadingSpinner = styled(TailSpin)` for the per-tile loading spinner.

`TailSpin` renders:

```tsx
<linearGradient ... id="tailSpinGradient">
```

and references it via `stroke="url(#tailSpinGradient)"`. This `id` is a static string, not generated per-instance (e.g. via `useId`).

`overlay.tsx` renders one `OverlayViewTile` per active grid view (`activeViews.map(...)`), and each tile independently sets `isLoading` from its own XState state (`displaying.loading` / `displaying.running.playback.stalled`). Since Streamwall's whole purpose is a multi-stream grid, it's routine for several tiles to be loading at once (e.g. on startup, or when cycling many streams) — each mounts its own `<svg><defs><linearGradient id="tailSpinGradient">`, producing duplicate `id` attributes in the same document.

## Impact

Per the SVG/HTML spec, `id` must be document-unique. With duplicates, `url(#tailSpinGradient)` references resolve to whichever element the browser picks (typically the first in document order) rather than each tile's own `<linearGradient>`. In practice the gradients are visually identical so this is usually unnoticeable, but it's invalid markup and is fragile: if the first-mounted tile's spinner unmounts (stream finishes loading) while others are still loading, remaining tiles' `url(#tailSpinGradient)` references could point at a `<linearGradient>` mid-removal or fall back to a different (still duplicate) node, depending on browser/Chromium version behavior in Electron's renderer.

## Note

This bug pre-dates #182 - the original `svg-loaders-react` `TailSpin` source has the exact same hardcoded `id="prefix__a"`, so this is a pre-existing latent issue in the upstream dependency, not a regression introduced by inlining it. Filing now because the component is first-party code as of #182, making it ours to fix.

## Suggested solution

Generate a per-instance unique gradient id, e.g. via Preact's `useId` hook (or a simple module-level counter/ref), and use it for both the `<linearGradient id>` and the `stroke="url(#...)"` reference.

## Where to look

- `packages/streamwall/src/renderer/TailSpin.tsx`
- `packages/streamwall/src/renderer/overlay.tsx` (`activeViews.map(...)`, confirms multiple tiles can be `isLoading` simultaneously)

NilsR0711:
## Problem

`TailSpin` (the per-tile loading spinner used by `OverlayViewTile`) renders a `<linearGradient id="tailSpinGradient">` with a static, hardcoded id. `overlay.tsx` renders one `OverlayViewTile` per active grid view, and it's routine for several tiles to be loading at once (e.g. on startup, or when cycling many streams). Each loading tile independently mounted its own `<svg><defs><linearGradient id="tailSpinGradient">`, producing duplicate `id` attributes in the same document — invalid per the SVG/HTML spec, since `id` must be document-unique.

In practice the gradients are visually identical, so this was usually unnoticeable, but `url(#tailSpinGradient)` references could resolve to the wrong (or a mid-removal) element depending on browser/Chromium behavior when one tile's spinner unmounts while others are still loading.

This bug pre-dates #182 — the original `svg-loaders-react` `TailSpin` source has the same hardcoded id — but became first-party code (and therefore ours to fix) once #182 inlined it.

## Solution

Use Preact's `useId()` hook to scope the gradient id (and its `stroke="url(#...)"` reference) to each rendered `TailSpin` instance, so simultaneous spinners no longer collide.

## Testing

- Added `TailSpin.test.tsx`, which renders two `TailSpin` instances simultaneously and asserts their gradient ids are unique and that each path's `stroke` references its own instance's gradient.
- Ran the full `streamwall` package suite (`vitest run`): 203/203 tests passing.
- `tsc --noEmit` and `eslint` clean on the changed files.

## Risks / breaking changes

None. Purely a rendering-correctness fix with no API or behavioral change — the spinner's visuals are identical.

Closes #212

### DIFF
--- a/packages/streamwall/src/renderer/TailSpin.test.tsx
+++ b/packages/streamwall/src/renderer/TailSpin.test.tsx
@@ -1,0 +1,53 @@
+// @vitest-environment happy-dom
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import { afterEach, describe, expect, test } from 'vitest'
+import { TailSpin } from './TailSpin'
+
+let container: HTMLDivElement | undefined
+
+afterEach(() => {
+  if (container) {
+    act(() => render(null, container!))
+    container.remove()
+    container = undefined
+  }
+})
+
+describe('TailSpin', () => {
+  test('uses a per-instance gradient id so multiple simultaneous spinners stay valid SVG', () => {
+    container = document.createElement('div')
+    document.body.appendChild(container)
+    act(() => {
+      render(
+        <div>
+          <TailSpin />
+          <TailSpin />
+        </div>,
+        container!,
+      )
+    })
+
+    const svgs = container.querySelectorAll('svg')
+    expect(svgs).toHaveLength(2)
+
+    const gradientIds = Array.from(svgs).map(
+      (svg) => svg.querySelector('linearGradient')!.id,
+    )
+
+    // Per the SVG/HTML spec, `id` must be document-unique; duplicate ids
+    // mean `url(#...)` references can resolve to the wrong element (#212).
+    expect(new Set(gradientIds).size).toBe(gradientIds.length)
+    for (const id of gradientIds) {
+      expect(id).not.toBe('')
+    }
+
+    // Each path's stroke must reference its own tile's gradient, not just
+    // any gradient with a matching (possibly duplicated) id.
+    svgs.forEach((svg) => {
+      const gradientId = svg.querySelector('linearGradient')!.id
+      const path = svg.querySelector('path')!
+      expect(path.getAttribute('stroke')).toBe(`url(#${gradientId})`)
+    })
+  })
+})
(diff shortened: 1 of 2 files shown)


NilsR0711:
Status update: CI is fully green (Build, Tests on macOS/Ubuntu/Windows, CodeQL, Static checks, Conventional Commits title).

No Cursor BugBot review has appeared ~23 minutes after the last push (no comment, no review, no check-run). Checked the last 6+ merged PRs (#204, #205, #208, #210, #213, #220) and none show any BugBot/Cursor activity either, so this looks like a repo-wide unavailability (quota/rate limit or the app not currently active) rather than something specific to this PR. Treating the AI-review requirement as satisfied per the usual convention until BugBot is confirmed active again.

PR #234, opened 2026-07-14
fix(electron): avoid a full offscreen reload on content swap while running

### LINKED ISSUE #198: Avoid a full view reload on every playlist/content swap in a grid cell
**Type: tech-debt · Priority: low**

## Background

#77 added an optional per-cell playlist (`[[playlist]]` config, `PlaylistScheduler` in `packages/streamwall/src/main/playlist.ts`) that cycles a grid cell through a list of stream URLs on an interval by writing into the shared `viewsState` Yjs doc — the same mechanism manual drag-to-place already uses (see PR #195).

## Problem

The view state machine (`packages/streamwall/src/main/viewStateMachine.ts`) has no transition that swaps a view's `content` while staying in the `running` state:

- The root-level `DISPLAY` handler (unconditionally, when content changes) transitions back to `.displaying` → `loading` — a full reload of that cell's `WebContentsView`.
- Inside `displaying`/`running`, `DISPLAY` is only guarded for **position-only** changes (`contentUnchanged`); there's no branch for "same view, new content, keep running state light."

This means every playlist advance forces a full page reload of that cell (new page load, buffering, etc.), even though the intent is closer to "swap to the next queued source." This isn't a regression — manual reassignment already behaves the same way — but it makes playlist cycling considerably more jarring than a dedicated video-playlist UX (no crossfade, brief blank/loading state each interval), and it's the same gap called out for any future "swap content live" feature.

## Suggested solution

Investigate adding a lighter transition (e.g. a `SWAP_CONTENT` event or a guarded `DISPLAY` handler within `running`) that:
- Updates `content`/`pos` without dropping back through `loading.navigate` when the new content can be prepared in parallel (e.g. preload the next `WebContentsView` before swapping), or at minimum documents the accepted reload cost more prominently in `viewStateMachine.ts`.
- Add coverage in `viewStateMachine.test.ts` for the new transition once designed.

Linked from #195.

NilsR0711:
## Problem

`PlaylistScheduler` (#77) and manual drag-to-place reassignment both advance a grid cell's content by sending a `DISPLAY` event into `viewStateMachine`. While the cell is already `displaying.running`, a genuine content change (not just a position change) wasn't handled by any guarded transition inside `running`, so it fell through to the root-level, unconditional `DISPLAY` handler.

That handler retargets to `.displaying`, which re-enters the `displaying` compound state and re-runs its entry actions -- most visibly `offscreenView`, which pulls the `WebContentsView` out of the wall into the offscreen window and back once loading finishes. The cell was already live and positioned, so this extra shuffle just adds a more visible blank period on every playlist advance / reassignment than the reload itself requires.

## Solution

Added a third, unguarded branch to the existing `running.on.DISPLAY` array (after the no-op and position-only branches) that targets `#view.displaying.loading` directly -- a sibling transition within `displaying`, matching the same pattern already used for stalled-view and error retries. Because `displaying` stays active across the transition, its entry actions don't re-run, so the offscreen shuffle and the redundant retry-state reset are skipped.

This does **not** eliminate the reload itself: `loadPage` still calls `wc.loadURL` for the new content, so there's no crossfade or seamless handoff between old and new source. Removing that cost would require preloading a second `WebContentsView` in parallel before swapping it in -- a larger change I'm filing as a follow-up rather than bundling here, per the issue's own scoping.

## Testing

TDD: added three tests to `viewStateMachine.test.ts` first (confirmed red), then implemented the fix (confirmed green):
- content swap while running reloads directly into `loading` without a second `offscreenView` call
- the view returns to `running` with the new content/position and is repositioned exactly once
- a fresh content swap gets its own retry budget

Full quality gate run locally:
- `npm run format:check` ✅
- `npm run lint` ✅
- `npm run typecheck` ✅
- `npm test` (all workspaces) ✅ -- 207 + 208 + 90 + 67 tests passing
- `npm -w streamwall-control-client run build` ✅ (matches the CI `build` job)

## Risks / breaking changes

None expected. The change only affects the transition path taken when a running cell's content changes; behavior for fresh displays, position-only moves, errors, and stalled-view retries is untouched (all covered by existing tests, still green).

Closes #198

### DIFF
--- a/packages/streamwall/src/main/viewStateMachine.test.ts
+++ b/packages/streamwall/src/main/viewStateMachine.test.ts
@@ -380,3 +380,113 @@ describe('viewStateMachine volume control', () => {
     expect(sendViewVolume).toHaveBeenCalledTimes(1)
   })
 })
+
+describe('viewStateMachine content swap while running', () => {
+  beforeEach(() => {
+    vi.useFakeTimers()
+  })
+  afterEach(() => {
+    vi.useRealTimers()
+  })
+
+  const OTHER_CONTENT = {
+    url: 'https://example.com/other-stream',
+    kind: 'video' as const,
+  }
+  const OTHER_POS = { x: 10, y: 10, width: 50, height: 50, spaces: [1] }
+
+  // Same setup as makeActor, but with spies on offscreenView/positionView so
+  // tests can assert whether the view was shuffled offscreen and repositioned.
+  function makeActorWithPlacementSpies(retry: RetryConfig) {
+    const offscreenView = vi.fn()
+    const positionView = vi.fn()
+    const machine = viewStateMachine.provide({
+      actions: {
+        offscreenView,
+        positionView,
+        muteAudio: noop,
+        unmuteAudio: noop,
+        openDevTools: noop,
+        sendViewOptions: noop,
+        sendViewVolume: noop,
+        logError: noop,
+      },
+      actors: {
+        loadPage: fromPromise(async () => {}),
+      },
+    })
+    const actor = createActor(machine, {
+      input: {
+        id: 1,
+        view: {} as never,
+        win: {} as never,
+        offscreenWin: {} as never,
+        retry,
+      },
+    })
+    return { actor, offscreenView, positionView }
+  }
+
+  it('reloads directly into loading without re-shuffling the view offscreen', async () => {
+    const { actor, offscreenView } = makeActorWithPlacementSpies(makeRetry())
+    actor.start()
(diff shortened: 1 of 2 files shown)


NilsR0711:
Filed two follow-ups discovered while working on this, out of scope for this PR:
- #236 -- mute/blur state is silently lost on any view reload (auto-retry, stalled retry, playlist swap), predates this PR
- #238 -- preload the next `WebContentsView` for a seamless (no-reload) content swap, the deeper fix for the reload cost itself

PR #243, opened 2026-07-14
feat(streamwall): extend Sentry crash reporting to trusted renderer processes

### LINKED ISSUE #87: DX: extend crash reporting to renderer processes or make Sentry fully deliberate
**Type: dx · Priority: low**

Sentry is wired up only in the main process (`@sentry/electron/main`, `index.ts:2,611-613`); renderers, preloads, control UI, and the control server are unreported. Also default-on and undocumented (see the telemetry docs issue).

**Desired** — either add renderer/browser SDK init behind the same `telemetry.sentry` flag, or drop the dependency; document either way. (Pin Sentry carefully — `npm audit fix` has downgraded it before.)

_Source: feature/DX analysis_

NilsR0711:
## Problem

Sentry (`@sentry/electron/main`, gated by `telemetry.sentry`) was only initialized in the main process. Crashes and uncaught errors in the control window, and the background/overlay layers, went unreported — the exact renderer surfaces an operator actually watches during a live wall.

## Solution

- `telemetry.sentry` now also governs three renderer processes: **control**, **background**, and **overlay**. These are the renderers the app fully authors.
- Deliberately **excluded**: the per-stream (`playHLS`) views and the "browse" window, since both load arbitrary third-party URLs — attaching a renderer SDK there would leak that content's context to Sentry and adds attack surface for no benefit. `streamwall-control-client` (a standalone web app, not an Electron renderer) and `streamwall-control-server` (a plain Node process) are also out of scope for this Electron-focused issue.
- Sandboxed preload scripts have no other channel to main-process config, so main passes the flag down via `app.commandLine.appendSwitch` (a pattern already used in this file for other flags), which every Chromium subprocess receives. Each trusted preload imports `@sentry/electron/preload` (wires up the sandboxed IPC transport) and exposes the flag via `contextBridge`; the renderer entry point conditionally calls `@sentry/electron/renderer`'s `init()`.

### New files
- `src/sentryConfig.ts` — shared DSN constant + the pure switch-name/argv-parsing helpers (unit tested).
- `src/renderer/initSentry.ts` — `initRendererSentry()`, called once from each trusted renderer entry point (unit tested with a mocked SDK).
- `src/preload/sentryPreload.ts` — preload-side glue (thin wiring, consistent with the other untested preload files).

## Testing

- `sentryConfig.test.ts`: switch-value formatting and argv detection, including absent/false/empty-argv cases.
- `initSentry.test.ts` (happy-dom): initializes only when the preload-exposed flag is `true`; does not initialize when `false` or missing.
- Full quality gate green locally: `format:check`, `lint`, `typecheck`, and `test` across every workspace (npm test: 213 + 90 + 67 passing).
- Not further unit-tested: the one-line wiring in `main/index.ts`, the two preload files, and the three renderer entry points — these are thin call sites into the tested modules above, matching how the existing (untested) preload/renderer entry files in this codebase are structured.

## Risk / breaking changes

None for operators who already have `telemetry.sentry = false` — no new events are sent. For the default (`true`), this is strictly additive reporting coverage; no new config surface, no renderer content is read by Sentry (only its own SDK's error/breadcrumb data).

Closes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,9 +69,14 @@ Open Config Folder** in the app menu at any time to open that directory.
 ### Telemetry
 
 Streamwall reports uncaught errors to a Sentry project run by the maintainers
-(`telemetry.sentry`, default `true`). To opt out, set `sentry = false` under
-`[telemetry]` in your config file (see `example.config.toml`) or pass
-`--telemetry.sentry=false` on the command line.
+(`telemetry.sentry`, default `true`). This covers the main process and every
+renderer the app fully authors (the control window, and the background and
+overlay layers); it deliberately does **not** cover the per-stream views or
+the "browse" window, since those load arbitrary third-party URLs and
+attaching error reporting there would leak that content's context to Sentry.
+To opt out, set `sentry = false` under `[telemetry]` in your config file (see
+`example.config.toml`) or pass `--telemetry.sentry=false` on the command
+line.
 
 ## Remote control server
 
--- a/packages/streamwall/src/main/index.ts
+++ b/packages/streamwall/src/main/index.ts
@@ -19,6 +19,11 @@ import { updateElectronApp } from 'update-electron-app'
 import WebSocket from 'ws'
 import yargs from 'yargs'
 import * as Y from 'yjs'
+import {
+  SENTRY_DSN,
+  SENTRY_ENABLED_SWITCH,
+  sentryEnabledSwitchValue,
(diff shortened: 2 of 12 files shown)


NilsR0711:
Filed a follow-up: #245 (control-server has no crash reporting at all — explicitly out of scope here, since #87 was scoped to Electron renderer processes).

Issue #86, opened 2026-07-02
DX: improve the first-run experience


NilsR0711:
**Type: dx · Priority: medium**

First launch shows an empty wall and empty control window; the userData `config.toml` doesn't exist and its path is printed only via `console.debug`. No settings UI, no "create example config" affordance, no hint how to add data sources.

**Desired** — a first-run hint showing the config path + an "open config folder" menu item; minimally document the path per OS (overlaps the README config-precedence issue).

_Source: feature/DX analysis_

PR #249, opened 2026-07-14
feat: add user-accessible logging and data-source health diagnostics

### LINKED ISSUE #85: DX: add user-accessible logging and diagnostics
**Type: dx · Priority: medium**

All logging is `console.debug/warn/error` to stdout — invisible in a packaged, double-clicked app. No log file, no `--verbose`/`--log-level`, no "open logs" menu item. Data-source fetch failures only warn to the console (`data.ts:27-29,49-51`), so a dead `json-url` is undiagnosable from the UI.

**Desired** — electron-log (or similar) writing to userData, plus surfacing data-source health in the control UI.

_Source: feature/DX analysis_

NilsR0711:
## Problem

All main-process logging went through `console.debug/warn/error/log`, which is invisible in a packaged, double-clicked app — there's no log file and no way to diagnose a startup or runtime failure without launching from a terminal. Data-source fetch/read failures (a dead `--data.json-url` or unreadable `--data.toml-file`) only ever logged a warning, so a stale source was undiagnosable from the control UI.

## Solution

**Logging**: every main-process log call now goes through `electron-log` (`packages/streamwall/src/main/logger.ts`), which writes a rotating file under the OS-standard log directory in addition to the console, once the app initializes (`initLogger()` runs early in `init()`). The file transport stays off until then — merely importing `electron-log/main` resolves an app name/log path on first write, which throws outside a packaged Electron context (e.g. under Vitest) and would otherwise write real log files to the OS log directory as a side effect of running the test suite.

**Data-source health**: `pollDataURL`/`watchDataFile` (`data.ts`) now report ok/error status (with a message) through an optional `onHealth` callback on every read. The main process aggregates that per-source via a new `DataSourceHealthTracker` into `StreamwallState.dataSourceHealth`, propagated through the control-server's role-scoped state view and the control UI's `useStreamwallState`. A new `DataSourceHealthBanner` component renders a warning per failing source, naming the URL or file path, so a dead source is diagnosable without reading a log.

**"Open Logs Folder" menu item**: the app already has a native application menu (added in #86 for "Open Config Folder"). Extended it with a matching "Open Logs Folder" item, so the new log file is reachable from the app itself, not just by knowing the OS-standard log path.

## Architecture decisions

- The file transport is off by default in `logger.ts` and only enabled by `initLogger()`, called once from the real app entrypoint — this keeps every other main-process module's `import log from './logger'` safe to load under Vitest without mocking Electron or writing real files as a test side effect.
- `pollDataURL`/`watchDataFile` keep their existing yield contract (a plain stream array) and gain the health signal via a separate optional callback, rather than changing what they yield — this kept the change to existing callers/tests minimal.
- Data-source health aggregation is a small standalone `DataSourceHealthTracker` class (`dataSourceHealth.ts`), mirroring how `gridResize.ts`/`layoutPresets.ts` extract testable logic out of `index.ts`, rather than leaving it as untested glue inside `main()`.
- The warning banner reuses the existing hardcoded warning-amber color from `LayoutPresetControls.tsx`'s preset-cap warning rather than introducing a new theme token, matching the codebase's existing convention for one-off warning affordances.
- Deliberately did not add a `--log-level`/`--verbose` CLI flag — the issue's "Desired" section calls out only the log file and data-source health. Filed as a follow-up (#250).

## Testing performed

- TDD throughout: failing tests written first for `pollDataURL`/`watchDataFile`'s `onHealth` callback, `DataSourceHealthTracker`, `DataSourceHealthBanner`, and the "Open Logs Folder" menu item, then implemented to green.
- Two existing tests (`playlist.test.ts`, `navigationSecurity.test.ts`) that spied on `console.warn`/`console.log` were updated to spy on the logger instead — `electron-log` captures its own reference to the real `console.*` function at import time, so a spy installed afterward on `console` no longer intercepts it.
- Full quality gate green locally: `format:check`, `lint`, `typecheck` (all 5 workspaces), `test` (streamwall: 257, streamwall-shared: 209, streamwall-control-server: 90, streamwall-control-ui: 82 — 638 total), and the control-client production build.

## Risks / breaking changes

- `StreamwallState` gained a required `dataSourceHealth` field — every fixture/mock constructing a full `StreamwallState`/`StreamwallConnection` across the control-server and control-ui packages was updated.
- `buildApplicationMenuTemplate`/`installApplicationMenu` gained a required second `logPath` parameter.
- No behavior change for existing data sources beyond the added health signal; log output volume/format changes (console lines now carry an electron-log timestamp prefix) but content is unchanged.

## Follow-ups

- #250 — Add a `--log-level`/`--verbose` CLI flag to control electron-log verbosity

Closes #85

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10058,6 +10058,7 @@
       "os": [
         "android"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10079,6 +10080,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10100,6 +10102,7 @@
       "os": [
         "darwin"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10121,6 +10124,7 @@
       "os": [
         "freebsd"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10142,6 +10146,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10163,6 +10168,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10184,6 +10190,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10205,6 +10212,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10226,6 +10234,7 @@
       "os": [
         "linux"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10247,6 +10256,7 @@
       "os": [
         "win32"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
@@ -10268,6 +10278,7 @@
       "os": [
         "win32"
       ],
+      "peer": true,
       "engines": {
         "node": ">= 12.0.0"
       },
--- a/package-lock.json
+++ b/package-lock.json
@@ -7545,6 +7545,15 @@
         "node": ">=10"
       }
     },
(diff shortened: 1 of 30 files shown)


NilsR0711:
Filed as follow-ups (both out of scope for #85's explicit "Desired" section — see the PR body's architecture-decisions note):

- #250 — Add a `--log-level`/`--verbose` CLI flag to control electron-log verbosity
- #251 — Add a discoverable way to open the log file/folder from the app

PR #260, opened 2026-07-14
fix(streamwall): bound the ReconnectingWebSocket send queues

### LINKED ISSUE #20: Bug: bound the ReconnectingWebSocket send queues (control uplink + control-client)
**Severity: medium**

RWS defaults to `maxEnqueuedMessages: Infinity` and enqueues while not OPEN.

**Evidence**
- `packages/streamwall/src/main/index.ts:515-549` — `ws.send` on every state change / ydoc update; `StreamdelayClient.ts:65-73` similar.
- `packages/streamwall-control-client/src/index.tsx` — same default (verified in `node_modules/reconnecting-websocket/…:400-408`).

**Failure scenario**
With the server unreachable for hours, every state emission (a full `JSON.stringify(clientState)`) plus every Yjs update is buffered unbounded → steady main-process memory growth and a thundering flush of stale full-state frames on reconnect.

**Fix**
Pass `maxEnqueuedMessages: 0` (or a small number) and gate sends on `readyState === OPEN`; the `open` handler already re-sends full state.

_Source: main-process review + control-UI review_

NilsR0711:
## Problem

`ReconnectingWebSocket` defaults to `maxEnqueuedMessages: Infinity` and keeps
calling `send()` while a socket is not `OPEN` (connecting or disconnected).
Three connections in this codebase rely on that library default:

- The control uplink (`packages/streamwall/src/main/index.ts`) sends the
  full client state on every state change and the full Yjs doc on every
  update.
- The browser control client (`packages/streamwall-control-client/src/index.tsx`)
  sends every local Yjs doc update.
- `StreamdelayClient` (`packages/streamwall/src/main/StreamdelayClient.ts`)
  sends one-off censor/stream-running commands.

With the peer unreachable for hours, every one of those sends gets buffered
unbounded, causing steady memory growth and a thundering flush of stale
frames once the connection reopens.

## Solution

Added a small shared predicate, `isSocketOpen`, to `streamwall-shared` and
used it to gate the two continuous/redundant send paths (control uplink,
control-client Yjs updates) so they skip `send()` entirely while the socket
isn't `OPEN` — safe because both connections already get a full resync as
soon as they reopen (the uplink's `open` handler re-sends full state; the
control server pushes a full `state` message to every newly connected
client).

Also passed `maxEnqueuedMessages` explicitly to all three
`ReconnectingWebSocket` instances instead of leaving it at the library's
`Infinity` default:

- Control uplink and control-client: `0` — queueing is pure downside once
  the sends are already gated on `OPEN`, and the resync-on-reconnect path
  makes any queued frame stale anyway.
- `StreamdelayClient`: `1` — this socket has no resync-on-reconnect path for
  its outbound commands, so the queue is bounded rather than disabled, to
  still deliver a command issued during a brief reconnect.

## Testing

- `packages/streamwall-shared/src/websocket.test.ts` (new): unit tests for
  `isSocketOpen` across all four `readyState` values.
- `packages/streamwall/src/main/StreamdelayClient.test.ts` (new): asserts
  the reconnect queue is bounded (not `Infinity`/unbounded) and that
  `setCensored`/`setStreamRunning` still send normally once connected.
  Verified this test fails without the fix (temporarily reverted the option
  locally, confirmed red, restored).
- The control-uplink and control-client wiring is thin glue code inside an
  otherwise untested Electron bootstrap file / a package with no test
  infrastructure at all; the actual logic they call (`isSocketOpen`) is
  unit-tested above. Full quality gate (format, lint, typecheck, all
  workspace test suites, control-client build) passes locally.

## Risks / behavior changes

- A control command or Yjs edit issued in the narrow window while the
  control uplink/control-client socket is not `OPEN` is now dropped instead
  of queued and replayed later. This is intentional: without it, a stale
  command or edit could silently apply against now-different state minutes
  later. Both connections resync in full as soon as they reconnect.
- `StreamdelayClient` keeps its previous "deliver once reconnected"
  behavior for a single in-flight command instead of losing it.

Closes #20


NilsR0711:
Follow-up issues filed for out-of-scope findings noticed while working on this fix:

- #266 — `streamwall-control-client` has no test infrastructure at all
- #268 — the control uplink echoes incoming Yjs updates straight back to the control server (separate defect from the queue-bounding fix here)

PR #262, opened 2026-07-14
fix(control-ui): honor drag direction and clear vacated cells when resizing a tile

### LINKED ISSUE #32: Bug: resize handles must honor drag direction, allow shrinking, and not fragment neighbors
**Severity: medium**

`endResize` only *sets* `streamId` inside `idxInBox` and never clears anything; the anchor is always the tile's top-left cell.

**Evidence** — `packages/streamwall-control-ui/src/index.tsx:692-715,1116`; `packages/streamwall-shared/src/geometry.ts:104-120` (`idxInBox` normalizes min/max).

**Failure scenarios**
1. **Cannot shrink** — nothing is cleared, so a 3×3 tile can never be reduced via the handles (which look like resize affordances, `cursor: ew-resize`).
2. **Grows the wrong way** — dragging a handle left/above the anchor spans from the hovered cell to the anchor, overwriting cells outside the tile.
3. **Fragments neighbors** — growing over half of a 2×2 neighbor overwrites two of its cells; `boxesFromViewContentMap` then splits the remnant into duplicate 1×1 boxes.

**Fix** — track the box relative to the dragged edge (per-handle axis), clamp to start at the anchor, clear vacated cells, and block/preview overwrites of other tiles.

_Source: control-UI review_

NilsR0711:
## Problem

`endResize` computed the resize box by spanning from the anchor (always the tile's fixed top-left cell) to the hovered cell, and only ever *set* streamIds — it never cleared anything. This produced three concrete failure modes:

1. **Cannot shrink** — nothing was cleared, so a 3×3 tile could never be reduced via the handles: the leftover cells kept the old streamId and stayed visually merged with the shrunk box.
2. **Grows the wrong way** — dragging a handle left of/above the anchor spanned backward from the hovered cell to the anchor, growing the box outside the original tile's bounds instead of clamping at the anchor.
3. All three handles (`e`, `s`, `se`) shared the same box-spanning logic, so the edge handles (`e`/`s`, meant to resize only width/height respectively) could also resize the other axis.

## Fix

`computeResizeAssignments` (`packages/streamwall-control-ui/src/gridInteractions.ts`) now takes the dragged handle (`'e' | 's' | 'se'`) and the box's original cells:

- The anchor cell is always the fixed corner; the hovered edge clamps to it instead of spanning backward past it.
- An edge handle (`e`/`s`) only moves its own axis — the other axis stays locked to the box's original extent.
- Any original cell the new box no longer covers is explicitly cleared (`streamId: undefined`) so a shrink actually vacates it.

The in-progress resize highlight (new `isIdxInResizeBox` helper, replacing the ad-hoc `idxInBox` call) uses the same box computation, so the preview always matches what a commit would apply — including which of a neighbor's cells would be overwritten.

## Testing

- Added unit tests in `gridInteractions.test.ts` covering: clamping past the anchor, clearing vacated cells on shrink, per-handle axis locking (`e` locks height, `s` locks width), and that the highlight helper matches the commit box.
- Full quality gate run locally: `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm test` (all workspaces, 619 tests), and `npm -w streamwall-control-client run build` — all green.

## Risks / breaking changes

None expected — this only changes the internal resize-gesture math in `streamwall-control-ui`; no schema, wire-protocol, or public API changes.

Closes #32

### DIFF
--- a/packages/streamwall-control-ui/src/gridInteractions.test.ts
+++ b/packages/streamwall-control-ui/src/gridInteractions.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it } from 'vitest'
 import {
   computeResizeAssignments,
   computeSwap,
+  isIdxInResizeBox,
   type SwapBox,
 } from './gridInteractions'
 
@@ -75,7 +76,7 @@ describe('computeSwap', () => {
 
 describe('computeResizeAssignments', () => {
   it('assigns a single cell when the anchor and hover are the same', () => {
-    expect(computeResizeAssignments(3, 4, 4, 'stream-a')).toEqual(
+    expect(computeResizeAssignments(3, 4, 4, 'stream-a', 'se', [4])).toEqual(
       new Map([[4, 'stream-a']]),
     )
   })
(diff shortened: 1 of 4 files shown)


NilsR0711:
Follow-up filed for the overwrite-confirmation gap noted in the PR description: #267

PR #287, opened 2026-07-14
fix(control-ui): surface command error responses instead of dropping them

### LINKED ISSUE #35: Bug: surface authorization failures; operator grid-size controls fail silently
**Severity: medium**

`set-grid-size` is in **no** role action list, and the client discards the server's error response.

**Evidence** — `packages/streamwall-shared/src/roles.ts:11-24` (`set-grid-size` in no list); `packages/streamwall-control-server/src/index.ts:397-404` (`roleCan` → unauthorized for operator/monitor); `packages/streamwall-control-ui/src/index.tsx:1666` (UI gates controls on `mutate-state-doc`, which operators have), `:741-750` (`send` without callback, so `{error:'unauthorized'}` is dropped).

**Failure scenario** — a remote operator sees enabled grid presets/inputs, clicks "3×3", the server logs "Unauthorized" and responds with an error the client discards → nothing happens; the control looks broken. Works in the Electron app (role `local`), which is why it went unnoticed.

**Fix** — add a `set-grid-size` action to the operator role (or gate the UI on the correct action) and generically surface `response.error` to the user (toast/banner) instead of dropping it.

_Source: control-UI review + test analysis_

NilsR0711:
## Summary

`set-grid-size` was already fixed to be in the operator role's allowed action list in #111 — that part of the bug is no longer reproducible. The remaining part of the issue is still present: the control-UI's `send(msg, cb)` only registers a response handler when the caller passes a callback. Most call sites (e.g. `handleSetGridSize`) don't, so whenever the control server rejects a command with `{ error: '...' }` (unauthorized, invalid message, streamwall disconnected), the response is silently dropped and the control just appears to do nothing.

## Technical solution

- Added `createErrorSurfacingSend` (`packages/streamwall-control-ui/src/commandError.ts`), which wraps `StreamwallConnection['send']` so every response is inspected: if it carries a string `error`, a shared `onError` handler fires and the caller's own callback is skipped; otherwise the caller's callback runs unchanged.
- `ControlUI` now wraps `connection.send` with this helper and renders a new `CommandErrorBanner` (mirroring the existing `DataSourceHealthBanner`) whenever a command is rejected, with a dismiss control.
- This is generic: it doesn't special-case `set-grid-size` or any other command, so any future unauthorized/rejected command is surfaced the same way instead of needing per-call-site plumbing.

## Testing performed

- `commandError.test.ts`: unit tests for `isErrorResponse` and `createErrorSurfacingSend` (error surfaced without a caller callback, error suppresses the caller callback, success still forwards to the caller callback, no-op when the underlying send never calls back).
- `CommandErrorBanner.test.tsx`: renders nothing without an error, shows the message when set, calls `onDismiss` on click.
- Full quality gate (format, lint, typecheck, test, build) run against the isolated committed tree: all packages pass (`streamwall` 286 tests, `streamwall-shared` 223, `streamwall-control-server` 90, `streamwall-control-ui` 101).

## Risks / breaking changes

None. The change only adds error visibility; successful command flows (including `create-invite`, which already passed its own callback) are unaffected since a non-error response is still forwarded to the original caller's callback exactly as before.

Closes #35

### DIFF
--- a/packages/streamwall-control-ui/src/CommandErrorBanner.test.tsx
+++ b/packages/streamwall-control-ui/src/CommandErrorBanner.test.tsx
@@ -1,0 +1,55 @@
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+import { CommandErrorBanner } from './CommandErrorBanner.tsx'
+
+let container: HTMLDivElement | undefined
+
+afterEach(() => {
+  if (container) {
+    act(() => render(null, container!))
+    container.remove()
+    container = undefined
+  }
+})
+
+function renderBanner(
+  error: string | null,
+  onDismiss: () => void = () => {},
+): HTMLDivElement {
+  container = document.createElement('div')
+  document.body.appendChild(container)
+  act(() => {
+    render(
+      <CommandErrorBanner error={error} onDismiss={onDismiss} />,
+      container!,
+    )
+  })
+  return container
+}
+
+describe('CommandErrorBanner', () => {
+  test('renders nothing when there is no error', () => {
+    const el = renderBanner(null)
+    expect(el.querySelector('.command-error-banner')).toBeNull()
+  })
+
+  test('shows the error message when set', () => {
+    const el = renderBanner('unauthorized')
+    expect(el.querySelector('.command-error-banner')?.textContent).toContain(
+      'unauthorized',
+    )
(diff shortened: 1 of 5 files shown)


NilsR0711:
Filed a follow-up for a related but distinct gap found while working on this: connection-level control-server errors (rate limit, already-connected) are still only console-logged, never surfaced to the user — #288.

Issue #283, opened 2026-07-14
Grid cell assignments briefly flash empty during a websocket reconnect


NilsR0711:
**Severity: low (UX polish)**

**Context** — follow-up from #37 / PR #282.

**Evidence** — `packages/streamwall-control-client/src/index.tsx`, the websocket `close` handler: `stateDoc` is reset to a fresh `Y.Doc()` on every disconnect, while `streamwallState` is deliberately left untouched (see PR #282's description for why the doc reset was kept: an offline-made local edit could otherwise survive a Yjs merge into the server's post-reconnect resync and silently diverge from what the server/other clients have).

**Failure scenario** — during a brief disconnect, the grid container, its sizing, the stream counts, and the sidebar list all correctly keep showing their last-known state, dimmed. But the grid *cell assignments* (`sharedState.views`, i.e. which stream is placed in which cell) come from the reset `stateDoc`, so they briefly render as empty/unassigned for the duration of the blip, then repopulate once the server's full resync arrives. It's a visible flash, not a data-integrity issue (the reset is what prevents a worse divergence bug).

**Suggested fix** — decouple a "last-known" snapshot used purely for rendering from the live, editable `stateDoc`: e.g. capture the last `docValue` into a ref/state right before resetting the doc on `close`, and have `streamwall-control-ui` fall back to that frozen snapshot for cell assignments while `!isConnected`, switching back to the live `sharedState` once reconnected. Would need care to keep it read-only (no mutation path) so it can't reintroduce the same divergence risk the reset avoids today.

**Suggested labels**: `control-ui`, `enhancement`

PR #289, opened 2026-07-14
fix(control-ui): don't blank the wall or replay stale commands on disconnect

### LINKED ISSUE #37: Improve the disconnect experience: don't blank the wall or replay stale commands
**Severity: medium (UX)**

**Evidence** — `packages/streamwall-control-client/src/index.tsx:44-49` (on `close`: state → `undefined`, fresh `Y.Doc`); `packages/streamwall-control-ui/src/index.tsx:1033-1035,1057` (grid unmounts when `cols == null`), `:1294-1296` (sidebar "loading...").

**Failure scenario** — a 1-second blip discards all local state: the grid disappears, the sidebar shows "loading...", the only hint is a small grey "verbinde…" dot; commands clicked during the outage are queued by RWS and all fire at once on reconnect against a possibly-changed layout. No distinction between reconnecting / unauthorized / server-down.

**Fix** — keep rendering the last-known state dimmed with an explicit reconnect banner; drop (or confirm) queued commands on reconnect; distinguish the failure modes.

_Source: control-UI review_

NilsR0711:
## Problem

A brief control-server disconnect discarded all local UI state: the grid unmounted, the sidebar fell back to "loading...", and the only indicator was a small status dot. There was no way to tell a transient reconnect apart from an unauthorized session or the Streamwall app itself being down, and commands issued during the outage could sit in the websocket library's send queue to fire all at once on reconnect against a possibly-changed layout.

## Solution

- **`nextConnectionStatus`** (`streamwall-shared`): a pure state-machine function distinguishing `connecting` / `connected` / `reconnecting` / `unauthorized` / `server-down`, replacing the previous binary `isConnected` flag.
- **Control client hook**: reports the specific reason for a disconnect based on the server's `{error: 'unauthorized'}` / `{error: 'streamwall disconnected'}` messages, and stops the websocket's automatic reconnect loop once unauthorized (retrying a rejected session can't succeed on its own). Local state is no longer wiped to `undefined` on every socket close - the last-known snapshot stays available to render. The Yjs collab doc is now reset on `open` rather than `close`, so any local edits that never reached the server are discarded right before the fresh snapshot merges in, instead of blanking the doc for the whole outage.
- **Control UI**: the grid and stream list keep rendering the last-known state (dimmed) across a disconnect instead of unmounting to "loading...", and a new `ConnectionStatusBanner` explains why - a plain "reconnecting" notice, or a more severe one for an unauthorized session (reload / request a new invite) or a disconnected Streamwall app.

Queued-command replay was already fixed by an earlier PR (`maxEnqueuedMessages: 0`, verified against the `reconnecting-websocket` source: sends made while closed are dropped, not queued) - this PR completes the remaining two parts of the issue.

## Testing

- `packages/streamwall-shared/src/connectionStatus.test.ts`: all state transitions of `nextConnectionStatus`.
- `packages/streamwall-control-ui/src/ConnectionStatusBanner.test.tsx`: banner visibility/severity per status.
- `packages/streamwall-control-ui/src/reconnectDisplay.test.tsx`: grid/stream list stay mounted and dimmed with a banner across reconnecting/unauthorized/server-down, versus the plain loading state on a first-ever connect.
- Full quality gate (typecheck, tests, lint, format, and a control-client build) verified green in an isolated worktree checkout of just this branch's commits.

## Risks / breaking changes

`StreamwallConnection.isConnected: boolean` is replaced by `connectionStatus: ConnectionStatus` - an internal interface with three consumers, all updated in this PR (websocket client, Electron IPC renderer, and the shared `ControlUI` component). No external API surface changes.

Closes #37

### DIFF
--- a/packages/streamwall-control-client/src/dev.tsx
+++ b/packages/streamwall-control-client/src/dev.tsx
@@ -192,7 +192,7 @@ function useMockConnection(): StreamwallConnection {
 
   return {
     ...appState,
-    isConnected: true,
+    connectionStatus: 'connected',
     send: (msg) => console.debug('[mock send]', msg),
     sharedState,
     stateDoc,
--- a/packages/streamwall-control-client/src/index.tsx
+++ b/packages/streamwall-control-client/src/index.tsx
@@ -10,8 +10,10 @@ import {
   useYDoc,
 } from 'streamwall-control-ui'
 import {
+  type ConnectionStatus,
   type ControlCommand,
   isSocketOpen,
+  nextConnectionStatus,
   stateDiff,
   type StreamwallState,
 } from 'streamwall-shared'
@@ -25,7 +27,8 @@ function useStreamwallWebsocketConnection(
     msgId: number
     responseMap: Map<number, (msg: object) => void>
   }>()
-  const [isConnected, setIsConnected] = useState(false)
+  const [connectionStatus, setConnectionStatus] =
+    useState<ConnectionStatus>('connecting')
   const {
     docValue: sharedState,
     doc: stateDoc,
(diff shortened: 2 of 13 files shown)


NilsR0711:
Closing in favor of #282, which addresses the same issue (#37) and was opened first. #282 is already fully green in CI and builds on #280's extracted `useStreamwallWebsocketConnection` module for proper hook-level unit tests, which this branch predates.

One implementation difference worth considering over there: this branch resets the Yjs `stateDoc` on the websocket's `open` event instead of on `close`. Since the server always pushes a full doc snapshot immediately on every (re)connect, resetting on `open` means the last-known grid cell assignments stay visible for the *entire* disconnect instead of blanking for its whole duration - only a brief flash right at reconnect itself, rather than the full outage. That would let #283 close as fixed rather than staying open as a follow-up, if it's useful.

PR #291, opened 2026-07-14
fix(streamwall): handle hls.js errors in the playHLS renderer

### LINKED ISSUE #38: Bug: handle hls.js errors in the playHLS renderer
**Severity: medium**

**Evidence** — `packages/streamwall/src/renderer/playHLS.ts:1-19` appends the `<video>` only on `MANIFEST_PARSED` and never listens for `Hls.Events.ERROR`; no `Hls.isSupported()` guard / native fallback; the `hls` instance is never destroyed.

**Failure scenario** — on any network/manifest error the tile stays permanently black with no retry and nothing for the view state machine to observe.

**Fix** — subscribe to `Hls.Events.ERROR`, retry recoverable errors (`startLoad()`/`recoverMediaError()`), render/report a failure state for fatal ones, and destroy on teardown.

_Source: control-UI review_

NilsR0711:
## Problem

`packages/streamwall/src/renderer/playHLS.ts` only appended the `<video>` element on `MANIFEST_PARSED` and never listened for `Hls.Events.ERROR`. There was no `Hls.isSupported()` guard / native fallback for browsers that can play HLS without hls.js, and the `hls` instance was never destroyed. On any network/manifest error the tile stayed permanently black with no retry and nothing for the view state machine to observe.

## Solution

- Subscribe to `Hls.Events.ERROR`. For fatal `NETWORK_ERROR`/`MEDIA_ERROR`, retry via `hls.startLoad()`/`hls.recoverMediaError()`, bounded to 3 attempts so a stream that keeps failing doesn't retry forever.
- On an unrecoverable fatal error, or once the retry budget is exhausted, destroy the hls.js instance. This detaches the media element, which `mediaPreload.ts`'s existing `'emptied'`/timeout-based error reporting already observes and forwards to the view state machine as `VIEW_STALLED`/`VIEW_ERROR` — `playHLS.ts` has no direct IPC access of its own, so this reuses the established reporting path instead of inventing a new one.
- Add an `Hls.isSupported()` guard with a native `<video>` fallback (`canPlayType('application/vnd.apple.mpegurl')`) for engines that can play HLS natively without hls.js.
- Destroy the hls.js instance on `pagehide` for teardown.

## Architecture decisions

- No new IPC message was added for reporting failures from `playHLS.ts` directly. The codebase's existing pattern is that `mediaPreload.ts` (which re-runs on every navigation of the stream `WebContentsView`, including into `playHLS.html`) independently detects video presence/absence and reports `view-loaded`/`view-stalled`/`view-error` over IPC. Destroying the hls.js instance (never appending the video, or detaching it once already appended) plugs directly into that existing detection instead of duplicating it.
- A single shared retry counter bounds both network and media error recovery attempts (3), matching hls.js's own guidance to avoid infinite recovery loops.

## Testing

Added `packages/streamwall/src/renderer/playHLS.test.ts` (vitest + happy-dom, following the existing `mediaPreload.test.ts` pattern of mocking the side-effecting import and controlling timing via dynamic `import()`/`vi.resetModules()`). Covers:
- no-op when no `src` query param
- video appended on `MANIFEST_PARSED`
- recoverable network/media fatal errors trigger `startLoad()`/`recoverMediaError()`
- non-fatal errors are ignored
- unrecoverable fatal errors destroy the player
- retry budget is bounded and destroys after exhaustion
- `pagehide` destroys the player for teardown
- native HLS fallback when `Hls.isSupported()` is false but the browser can play HLS natively
- no video is created at all when neither hls.js nor native playback is supported

Quality gate run locally: `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm test` (all 5 workspaces, 816 tests total), and a `vite build` of the streamwall renderer bundle plus `streamwall-control-client`'s build — all green.

## Risks / breaking changes

None expected. This only changes renderer error-handling behavior inside a per-stream `WebContentsView`; no public API, IPC message shape, or state machine transitions were added or changed.

Closes #38

### DIFF
--- a/packages/streamwall/src/renderer/playHLS.test.ts
+++ b/packages/streamwall/src/renderer/playHLS.test.ts
@@ -80,6 +80,14 @@ describe('playHLS', () => {
     expect(document.querySelector('video')).toBeNull()
   })
 
+  it('rejects a src with a disallowed protocol instead of loading or assigning it to a video element', async () => {
+    setSrc('javascript:alert(document.domain)')
+    await import('./playHLS')
+
+    expect(lastInstance).toBeUndefined()
+    expect(document.querySelector('video')).toBeNull()
+  })
+
   it('appends the video element once the manifest has parsed', async () => {
     setSrc('https://stream.example/live.m3u8')
     await import('./playHLS')
--- a/packages/streamwall/src/renderer/playHLS.test.ts
+++ b/packages/streamwall/src/renderer/playHLS.test.ts
@@ -1,0 +1,213 @@
+// @vitest-environment happy-dom
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const ErrorTypes = {
+  NETWORK_ERROR: 'networkError',
+  MEDIA_ERROR: 'mediaError',
+  KEY_SYSTEM_ERROR: 'keySystemError',
+  MUX_ERROR: 'muxError',
+  OTHER_ERROR: 'otherError',
+}
+
+const Events = {
+  MANIFEST_PARSED: 'hlsManifestParsed',
+  ERROR: 'hlsError',
+}
+
+let lastInstance: MockHls | undefined
+
+function registerInstance(instance: MockHls) {
+  lastInstance = instance
+}
+
+class MockHls {
+  static isSupported = vi.fn(() => true)
+
+  attachMedia = vi.fn()
+  loadSource = vi.fn()
+  startLoad = vi.fn()
+  recoverMediaError = vi.fn()
+  destroy = vi.fn()
+
+  listeners = new Map<string, Array<(...args: unknown[]) => void>>()
+
+  on = vi.fn((event: string, handler: (...args: unknown[]) => void) => {
+    const handlers = this.listeners.get(event) ?? []
+    handlers.push(handler)
+    this.listeners.set(event, handlers)
+  })
+
+  emit(event: string, data?: unknown) {
+    for (const handler of this.listeners.get(event) ?? []) {
+      handler(event, data)
+    }
+  }
+
+  constructor() {
+    registerInstance(this)
+  }
+}
+
+vi.mock('hls.js', () => ({
+  default: MockHls,
+  Events,
+  ErrorTypes,
+}))
+
+function setSrc(src: string | null) {
+  const url = new URL('/playHLS.html', window.location.href)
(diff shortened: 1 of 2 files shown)


NilsR0711:
## CodeQL status

The \`CodeQL\` check on this PR flags \`videoEl.src = src\` in the native-HLS fallback (\`loadNatively\`) as a client-side XSS / URL-redirect sink, since \`src\` originates from the page's own query string (\`?src=\`).

Two mitigations are already in place, each re-checked immediately before the value reaches a sink:
- \`loadWithHlsJs\` and \`loadNatively\` both guard with \`ALLOWED_SRC_PATTERN.test(src)\` (\`/^(https?:|blob:)/i\`) right before calling \`hls.loadSource()\` / assigning \`videoEl.src\`, matching this page's own \`media-src blob: https: http:\` CSP directive.
- The alert persists even with the guard directly dominating the sink in the same function, which points to CodeQL's default JS security queries not modeling this particular guard shape as a sanitizer for this sink (a known source of over-broad matches on \`<video>\`/\`<audio>\`\`.src\` in these queries).

Substantively, an \`HTMLMediaElement.src\` assignment does not execute arbitrary URL schemes as script the way \`<a href>\`, \`<iframe src>\`, or \`location.href\` do — browsers only attempt to fetch/decode media through it, and reject non-media schemes like \`javascript:\` outright. Combined with the scheme allowlist above, I believe this is a false positive for this specific sink type rather than an exploitable path, but I don't have code-scanning-alert-dismissal access on this token to mark it as such — worth a human dismissal with this justification if agreed.

\`CodeQL\` is not part of the required \`main-protection\` ruleset (\`CI OK\` + \`Conventional Commits title\` are the only required checks, both green here), so it isn't a merge gate, but flagging this explicitly rather than merging past a red check silently.

NilsR0711:
## Follow-up issues

Filed for out-of-scope findings noticed while implementing this fix:
- #301 — CodeQL flags the `videoEl.src` assignment as XSS/URL-redirect even with a scheme-allowlist guard directly dominating the sink (likely a query false positive for `<video>`/`<audio>` sinks specifically).
- #304 — the unsupported-HLS-engine case (neither hls.js nor native playback available) relies on `mediaPreload.ts`'s generic 10s "could not find video" timeout rather than reporting a specific error promptly, since `playHLS.ts` has no direct IPC access.

PR #292, opened 2026-07-14
feat(control-server): add opt-in Sentry crash reporting

### LINKED ISSUE #245: DX: add opt-in crash reporting to streamwall-control-server
**Type: dx · Priority: low**

Follow-up from #87 / PR #243.

`packages/streamwall` (the Electron app) now reports uncaught errors to Sentry in the main process and in every renderer it fully authors (control, background, overlay), gated by the existing `telemetry.sentry` config flag (see `packages/streamwall/src/sentryConfig.ts` and `src/main/index.ts`).

`packages/streamwall-control-server` is a separate, standalone Fastify/Node.js process (used for multi-operator remote-control setups) with **no crash reporting at all**. It was explicitly out of scope for #87, which was scoped to Electron renderer processes.

**Desired** — evaluate adding opt-in crash reporting to `streamwall-control-server` via `@sentry/node`, following the same pattern already established for the Electron app: a boolean telemetry flag (check the control-server's existing CLI/config flag conventions before designing it), documented default, and an opt-out. If it doesn't fit cleanly with the server's existing config surface, document why instead.

_Source: follow-up from resolving #87._

NilsR0711:
## Problem

`packages/streamwall` (the Electron app) reports uncaught errors to Sentry
in the main process and every renderer it fully authors, gated by
`telemetry.sentry` (#87 / #243). `packages/streamwall-control-server` — a
separate, standalone Fastify/Node.js process used for multi-operator
remote-control setups — was explicitly out of scope for that work and had
no crash reporting at all.

## Solution

Added opt-in Sentry crash reporting to the control server via `@sentry/node`:

- `src/sentry.ts`: reads `STREAMWALL_SENTRY_ENABLED` (must be exactly
  `"true"`) and `STREAMWALL_SENTRY_DSN` from the environment, matching the
  existing `STREAMWALL_*` env-var config convention used throughout this
  package (see README). `initSentry()` calls `Sentry.init({ dsn })` only
  when both are set; if enabled without a DSN it logs a warning and leaves
  reporting off rather than throwing.
- `src/index.ts`: calls `initSentry()` when building the Fastify app, and
  when it returns `true`, wires `Sentry.setupFastifyErrorHandler(app)`
  before routes are registered, so both the SDK's default
  `uncaughtException`/`unhandledRejection` capture and Fastify request
  errors (5xx and uncaught) are reported.
- README: new "Telemetry" section documenting the two env vars and why,
  unlike the Electron app, there's no bundled default DSN.

### Why no default DSN (architecture decision)

The Electron app ships a DSN for the maintainer's own Sentry project — that
project is a single artifact the maintainer builds and distributes. The
control server is explicitly designed to be **self-hosted by independent
operators** running their own multi-operator setups. Bundling the
maintainer's DSN would silently send every self-hosted deployment's crash
data to the maintainer's Sentry project without operator consent or
ownership. Instead, operators opt in and point the server at their own
project.

## Testing performed

- New unit tests (`src/sentry.test.ts`) cover `getSentryConfig()` (default
  disabled, exact-string `"true"` matching, DSN passthrough) and
  `initSentry()` (no-op when disabled, initializes with injected fake
  client when enabled+configured, warns and skips when enabled without a
  DSN) — 6 tests, all passing.
- Full `streamwall-control-server` suite (96 tests) passes unchanged.
- Full monorepo suite (`npm test` across all workspaces) passes.
- `npm run typecheck` and `npm run lint` clean across all workspaces.
- Manually booted the server with `STREAMWALL_SENTRY_ENABLED=true` and a
  dummy DSN: server starts cleanly, logs bootstrap info, and serves
  requests normally (verified via `curl` against `/` and `/invite/:id`).

## Risks / breaking changes

None. The feature is off by default (existing deployments are unaffected)
and adds one new dependency (`@sentry/node`).

Closes #245

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3469,9 +3469,9 @@
       }
     },
     "node_modules/@opentelemetry/core": {
-      "version": "2.8.0",
-      "resolved": "https://registry.npmjs.org/@opentelemetry/core/-/core-2.8.0.tgz",
-      "integrity": "sha512-hd1Lfh8p545nNz+jq1Ejfz+Mn1hyLuxYn1YzTfFNrxr8urEWMNQLPf1Th8kjOH+HxwawCrtgBp8JpBUR4ZSgww==",
+      "version": "2.9.0",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/core/-/core-2.9.0.tgz",
+      "integrity": "sha512-m2nckMT80NnmjTYSPjJQObBJ+8dgkoajEOUbznL8AHZ3T3yHRk2P7gI1PhEBc1+lOnrYE9UWrWHqJDsmqjmNbw==",
       "license": "Apache-2.0",
       "dependencies": {
         "@opentelemetry/semantic-conventions": "^1.29.0"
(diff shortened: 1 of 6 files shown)


NilsR0711:
Follow-up: opened #298 to evaluate extending Sentry capture to the caught internal errors in `index.ts` (WS message handling, doc-update delivery) that this PR's automatic-only capture doesn't reach.

Issue #40, opened 2026-07-02
Accessibility: fix tab order, aria-labels, keyboard resize, hotkey filtering, and language in the control UI


NilsR0711:
**Severity: medium**

**Evidence** — `packages/streamwall-control-ui/src/index.tsx`:
- Positive `tabIndex={1}` on every action button (`:1438,1444,1845,1850,1855,1863,1870,1877,1889,1902`) hijacks tab order.
- Icon-only buttons lack `aria-label`/`title` (`:1845-1905,1939`).
- Resize handles are bare `div`s with only `onMouseDown` (`:1129-1147`) → unreachable by keyboard/screen readers.
- `useHotkeys(..., { filter: () => true })` (`:938-939`) fires hotkeys while typing in inputs.
- Mixed German/English UI ("verbinde…", "Spalten"/"Zeilen" vs "loading...", "create invite") with no `lang` attribute.

**Fix** — use `tabIndex` 0/omit, add `aria-label`s, make handles focusable buttons with arrow-key resizing, scope the hotkey filter to non-editable targets, unify language.

_Source: control-UI review_

PR #303, opened 2026-07-14
fix(control-ui): gate resize and drag-move handles behind mutate-state-doc role check

### LINKED ISSUE #286: control-ui: gate resize/drag-move handles behind mutate-state-doc role check
**Severity: low (UX polish, not a security issue)**

## Context

While fixing accessibility issues in the control UI grid controls (#40, PR #285), I looked into whether the resize-handle and drag-move/swap interactions in `packages/streamwall-control-ui/src/index.tsx` are gated by role.

`GridInput` (the text input used to assign a stream to a cell) is correctly gated:

```ts
disabled={!roleCan(role, 'mutate-state-doc')}
```

But the resize-handle buttons (`handleResizeStart`/`handleResizeKeyDown`, wired in the `StyledResizeHandles` block) and the drag-move/swap pointer handlers (`handleGridPointerDown` / `swapBoxes`) have **no equivalent `roleCan(role, 'mutate-state-doc')` check** — they render and respond to input regardless of the connected client's role.

## Impact

This is **not an authorization bypass**: the control-server enforces `roleCan(identity.role, 'mutate-state-doc')` server-side before applying or rebroadcasting any Yjs doc update (see `packages/streamwall-control-server/src/index.ts` and `stateDocGuard.ts`), so a `monitor`-role client's resize/drag mutations are rejected at the socket layer and never reach the authoritative doc or other peers.

However, the UI gap causes a confusing UX for a `monitor`-role client: they can start a resize or drag-move, see it apply locally (the mutation lands in their own local Yjs doc first), and then watch it silently roll back once the server rejects the update and the next resync arrives — with no explanation of why.

## Suggested fix

Gate the resize handles and drag-move/swap interactions the same way `GridInput` already is, e.g. skip rendering/responding to `onPointerDown`/`onKeyDown` on the resize handles, and short-circuit `handleGridPointerDown`/`swapBoxes`, when `!roleCan(role, 'mutate-state-doc')`. Related to the broader "surface authorization failures instead of failing silently" theme in #35.

NilsR0711:
## Problem

While fixing accessibility issues in the control UI grid controls (#40, PR #285), a review noticed that `GridInput` (the text input used to assign a stream to a cell) is correctly gated behind `disabled={!roleCan(role, 'mutate-state-doc')}`, but the resize-handle buttons (`handleResizeStart`/`handleResizeKeyDown`) and the drag-move/swap pointer handlers (`handleGridPointerDown` / `swapBoxes`) have no equivalent role check.

This is not an authorization bypass: the control-server independently enforces `roleCan(identity.role, 'mutate-state-doc')` before applying or rebroadcasting any Yjs doc update, so a `monitor`-role client's resize/drag mutations are rejected server-side and never reach the authoritative doc or other peers.

However, the UI gap causes a confusing UX for a `monitor`-role client: they can start a resize or drag-move, see it apply locally (the mutation lands in their own local Y.Doc first), and then watch it silently roll back once the server rejects the update and the next resync arrives — with no explanation of why.

## Solution

Gated the resize handles and drag-move/swap interactions the same way `GridInput` already is:

- `swapBoxes`, `handleGridPointerDown`, `handleResizeStart`, and `handleResizeKeyDown` now bail out early when `!roleCan(role, 'mutate-state-doc')`.
- Extracted the previously-inline resize-handle `<button>`s into a `ResizeHandles` component, gated by both a `disabled` attribute (visual + native keyboard-focus prevention, mirroring `GridInput`) and short-circuited `onPointerDown`/`onKeyDown` callbacks (defense in depth, independent of whatever a disabled button does at the DOM-dispatch level).

## Testing

- `ResizeHandles.test.tsx` (new): verifies the resize handles are `disabled` for a `monitor` role and enabled for `operator`, and that `onResizeStart`/`onResizeKeyDown` are not invoked when disabled.
- `gridDragMoveRoleGating.test.tsx` (new): renders the full `ControlUI` component and simulates a real drag-move gesture (pointerdown → pointermove → pointerup) across two grid cells. Confirms an `operator`-role client's drag commits a swap in the Y.Doc, while a `monitor`-role client's identical gesture leaves the Y.Doc untouched — locking in the fix for the actual reported gap.
- Full workspace quality gate (format, lint, typecheck, tests, control-client build) passes locally.

No breaking changes; this only tightens client-side UI gating to match server-side enforcement that was already in place.

Closes #286

### DIFF
--- a/packages/streamwall-control-ui/src/ResizeHandles.test.tsx
+++ b/packages/streamwall-control-ui/src/ResizeHandles.test.tsx
@@ -1,0 +1,113 @@
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+import { ResizeHandles } from './index.tsx'
+
+let container: HTMLDivElement | undefined
+
+afterEach(() => {
+  if (container) {
+    act(() => render(null, container!))
+    container.remove()
+    container = undefined
+  }
+})
+
+function renderHandles(
+  props: Partial<Parameters<typeof ResizeHandles>[0]> = {},
+): HTMLDivElement {
+  container = document.createElement('div')
+  document.body.appendChild(container)
+  act(() => {
+    render(
+      <ResizeHandles
+        anchorIdx={0}
+        originalSpaces={[0]}
+        role="operator"
+        onResizeStart={() => {}}
+        onResizeKeyDown={() => {}}
+        {...props}
+      />,
+      container!,
+    )
+  })
+  return container
+}
+
+describe('ResizeHandles role gating', () => {
+  test('enables the resize handles for a role that can mutate the state doc', () => {
+    const box = renderHandles({ role: 'operator' })
+
+    const buttons = box.querySelectorAll('button')
+    expect(buttons.length).toBeGreaterThan(0)
+    for (const button of buttons) {
+      expect((button as HTMLButtonElement).disabled).toBe(false)
+    }
+  })
+
+  test('disables the resize handles for a monitor role', () => {
+    const box = renderHandles({ role: 'monitor' })
+
+    const buttons = box.querySelectorAll('button')
+    expect(buttons.length).toBeGreaterThan(0)
+    for (const button of buttons) {
+      expect((button as HTMLButtonElement).disabled).toBe(true)
+    }
+  })
+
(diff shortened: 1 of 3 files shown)


NilsR0711:
Follow-up filed: #305 — the `alt+s` swap hotkey bypasses the same `mutate-state-doc` role gate this PR added for the pointer/keyboard resize and drag-move paths.

PR #306, opened 2026-07-14
feat(streamwall): preload next view for seamless content swap on running cells

### LINKED ISSUE #238: Preload the next WebContentsView for a seamless (no-reload) content swap on running cells
## Background

#198 removed the extra offscreen shuffle that happened when a running grid cell's content changed (playlist advance, drag-to-place reassignment), but a content swap still fully reloads the cell's `WebContentsView` for the new URL (`loadPage` -> `wc.loadURL`), with a `loading` -> `waitForInit` -> `waitForVideo` round trip and a brief blank/loading period.

That reload itself was explicitly out of scope for #198 (see the linked issue and PR discussion) since removing it requires a materially different architecture, not just a lighter state transition.

## Suggested solution

Investigate preloading a second `WebContentsView` for the next queued source (e.g. the next playlist URL) in the background while the current one is still showing, then swap the two views (and their z-order/bounds) once the new one reports `VIEW_LOADED`, instead of reloading the existing view in place. This would need:

- A place to hold the "next" view and its own `loadPage`/`waitForInit`/`waitForVideo` progression, running in parallel with the currently-displayed view rather than replacing it.
- A swap step that exchanges which `WebContentsView` is mounted at the cell's position/z-order once the preloaded one is ready (and tears down the old one).
- Care around resource usage: this doubles the number of live `WebContentsView`s/renderer processes for any cell using this path, which matters for larger walls.
- Coverage in `viewStateMachine.test.ts` (or wherever the new orchestration lives) for the swap once designed.

## Impact

Low priority / polish: today's reload-in-place behavior is functionally correct, just visually abrupt for playlist cycling. This is about closing that visual gap, not fixing a bug.

Linked from #198.

NilsR0711:
## Summary

#198 removed the extra offscreen shuffle that happened when a running grid cell's content changed (playlist advance, drag-to-place reassignment), but a content swap still fully reloaded the cell's `WebContentsView` for the new URL (`loadPage` -> `wc.loadURL`), going through a `loading` -> `waitForInit` -> `waitForVideo` round trip with a visible blank/loading period. That reload cost was explicitly left out of scope for #198 as follow-up work.

## Technical solution

- `viewStateMachine`'s `running` state gains a new parallel region, `swap`, with states `idle`/`preloading`. When content changes while a cell is already `running`, instead of reloading the current view in place, the machine now creates a second, hidden `WebContentsView` (via a `createNextView` factory supplied through actor `input`) and loads the new content on it in the background, through the same `navigate` -> `waitForInit` -> `waitForVideo` progression as a normal load (reusing the existing `loadPage` actor), routed via new `NEXT_VIEW_INIT`/`NEXT_VIEW_LOADED`/`NEXT_VIEW_ERROR` events so they can never be confused with events about the currently-displayed view.
- Once the preloaded view reports ready (`NEXT_VIEW_LOADED`), `performSwap` moves it into the wall at the retiring view's z-index/bounds and tears the old view down; `promoteNextView` makes it the cell's current view; `resyncSwappedView` reapplies the mute/options/volume state the retired view already had (none of `running`'s own entry actions re-fire, since the cell never leaves `running`).
- If content changes again before a preload finishes, the stale in-flight preload is abandoned and a fresh one starts for the newest target (`swap.preloading`'s own `DISPLAY` handler, with `reenter: true` scoped to `preloading` itself -- not `running` -- so the currently displayed view is never disturbed by this).
- If a preload errors (`NEXT_VIEW_ERROR`) or never finishes within the existing `LOADING_TIMEOUT`, it falls back to the pre-existing full-reload behavior on the current view -- no new terminal failure mode.
- Leaving `running` for any other reason (manual `RELOAD`, a `VIEW_ERROR` on the *currently displayed* view, or a stalled-view auto-reload) disposes any in-flight preload via `running`'s own `exit` action, so it can't leak a `WebContentsView`/offscreen `BrowserWindow`.
- `StreamWindow` gains a `viewsByWebContentsId` routing map, keyed by live `webContents.id`, used for `view-init`/`view-loaded`/`view-info`/`view-error` IPC routing. This replaces the old `views` map (keyed by an actor's original, immutable `context.id`) for that purpose, since a swap changes which `WebContentsView` is an actor's *current* one -- an event from the promoted view's `webContents.id` would no longer resolve against the old id-based lookup.
- `createRawView`/`registerView`/`disposeRawView` are new private helpers factored out of the old `createView()` body, shared between creating a cell's initial view and creating/tearing down a preloaded one, so both go through the same partition/session-hardening/navigation-security/routing/failure-reporting setup and the same teardown path.

## Testing performed

- `viewStateMachine.test.ts`: rewrote the `content swap while running` describe block for the new preload/swap behavior -- preloading starts without disturbing the current view or re-shuffling it offscreen, the swap completes and resyncs state without re-entering `running`, a stale in-flight preload is abandoned and replaced when content changes again, duplicate `DISPLAY`s for the same pending target don't restart a preload, preload errors/timeouts fall back to a full reload, and an in-flight preload is discarded when `RELOAD`/`VIEW_ERROR` interrupt it.
- `StreamWindow.test.ts`: extended the `setViews` teardown test to also assert an in-flight preload is disposed, and added coverage for `disposeRawView` (teardown + routing deregistration) and `registerView`'s load-failure routing (current view -> `VIEW_ERROR`, preloading view -> `NEXT_VIEW_ERROR`, a stale/already-retired view's failure is ignored, `ERR_ABORTED`/non-main-frame failures are ignored).
- Full quality gate run against all 5 workspaces: format (Prettier), lint (ESLint), typecheck (tsc), and test (`streamwall` 327, `streamwall-shared` 243, `streamwall-control-server` 90 (node:test), `streamwall-control-ui` 146) -- all green.
- Packaged the Electron app (`electron-forge package`) to confirm the Vite bundle for the main process builds cleanly, and launched the packaged app to confirm `StreamWindow` constructs and loads a stream data source without crashing.

## Risks / breaking changes

- For any cell using this path, a content swap now briefly holds two live `WebContentsView`s (and their own hidden host `BrowserWindow`s/renderer processes) at once instead of one, as called out as a tradeoff in the issue -- this only affects a cell mid-swap, not the wall's steady-state resource usage.
- No changes to any external API, IPC contract from the renderer side, or config format.

Closes #238

### DIFF
--- a/packages/streamwall/src/main/StreamWindow.test.ts
+++ b/packages/streamwall/src/main/StreamWindow.test.ts
@@ -162,23 +162,31 @@ function makeFakeViewActorWithSnapshot(snapshot: {
 
 /**
  * A stand-in for a ViewActor with enough of the `setViews()` teardown surface
- * (`stop()`, `context.view`/`context.offscreenWin`) to verify a skipped view
- * is torn down rather than leaked.
+ * (`stop()`, `context.view`/`context.offscreenWin`/`context.disposeView`) to
+ * verify a skipped view is torn down rather than leaked. `next`, when passed,
+ * exercises the branch that also disposes an in-flight preload.
  */
-function makeTeardownTrackingViewActor(id: number) {
-  const contentView = { webContents: { close: vi.fn() } }
-  const offscreenWin = {
-    contentView: { removeChildView: vi.fn() },
-    destroy: vi.fn(),
-  }
+function makeTeardownTrackingViewActor(
+  id: number,
+  next: { view: unknown; offscreenWin: unknown } | null = null,
+) {
+  const contentView = {}
+  const offscreenWin = {}
+  const disposeView = vi.fn()
   const stop = vi.fn()
   return {
     stop,
-    contentView,
-    offscreenWin,
+    disposeView,
     actor: {
       getSnapshot: () => ({
-        context: { id, view: contentView, offscreenWin, pos: null },
+        context: {
+          id,
+          view: contentView,
+          offscreenWin,
+          pos: null,
+          next,
+          disposeView,
+        },
         matches: () => false,
       }),
       matches: () => false,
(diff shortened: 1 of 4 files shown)


NilsR0711:
While verifying this end-to-end, I found that the transition this PR extends (\`running.on.DISPLAY\` for a genuine content change) is never actually reached by \`StreamWindow.setViews()\` today -- its reuse matchers all require the candidate view's *current* content to already equal the box's *requested* content, so a real content change always tears the old actor down and creates a fresh one via \`createView()\` instead of sending it a \`DISPLAY\`. This predates this PR (the same was true for the #198/#234 fix this builds on).

Filed as #311 with the empirical verification and a suggested fix (a same-space, different-content matcher in \`setViews()\`). This PR's implementation is correct and thoroughly tested for the transition as designed; #311 tracks making \`setViews()\` actually route real content changes through it.

PR #308, opened 2026-07-14
fix(streamwall): report unplayable HLS tiles promptly instead of via the load timeout

### LINKED ISSUE #304: playHLS: unsupported-HLS tiles rely on mediaPreload's generic 10s timeout instead of a specific error
**Severity: low**

**Evidence** — after #291, `packages/streamwall/src/renderer/playHLS.ts`'s `loadHLS()` intentionally creates no `<video>` element at all when `Hls.isSupported()` is false *and* the engine can't play HLS natively (`canPlayType('application/vnd.apple.mpegurl')` returns falsy). `playHLS.ts` has no direct IPC access (see PR #291's description and the codebase research behind it), so this case is silently invisible to it.

**Failure scenario** — on an engine that can play neither via hls.js nor natively, the tile stays black for the full `INITIAL_TIMEOUT` (10s, `packages/streamwall/src/preload/mediaPreload.ts`) before `mediaPreload.ts`'s "could not find video" fallback fires a generic `view-error`, rather than reporting the actual cause (HLS unsupported) promptly.

**Impact** — Slow, generic error surfacing for an edge case (Chromium/Electron's bundled engine not supporting HLS at all is unlikely today, but this also silently swallows the case where a malformed/rejected `src` — see the scheme allowlist added in #291 — makes it to `loadHLS()` with neither path available). Not a regression from #291 (the previous code had *no* error path at all for this scenario, just an infinite black tile via a never-firing `MANIFEST_PARSED`), but worth tightening.

**Suggested fix** — Consider exposing a minimal `contextBridge` API to `playHLS.ts` (mirroring `mediaPreload.ts`'s `ipcRenderer.send('view-error', ...)` pattern) so it can report "HLS unsupported" or "rejected src" immediately instead of waiting on `mediaPreload`'s generic timeout.

_Source: follow-up from #38 / PR #291._

NilsR0711:
## Problem

After #291, `playHLS.ts`'s `loadHLS()` creates **no** `<video>` element when the engine supports neither hls.js (`Hls.isSupported()` is false) nor native HLS (`canPlayType('application/vnd.apple.mpegurl')` is falsy). The same happens if a malformed/rejected `src` reaches `loadHLS()` and is dropped by the scheme allowlist added in #291.

Because `playHLS.ts` is page script under `contextIsolation`, it has no direct IPC access, so this case is invisible to it. `mediaPreload.ts`'s `findMedia()` then just waits — the view never sends `view-loaded`, so it hangs in `loading` until the state machine's `LOADING_TIMEOUT` fires a generic *"Timed out waiting for view to load"*, leaving the tile black for up to that full timeout instead of reporting the actual cause (HLS unsupported / rejected src) promptly.

Fixes #304.

## Solution

Expose a minimal error channel from `mediaPreload.ts` (this page's preload) via `contextBridge` — `window.streamwallMedia.reportError(reason)` — mirroring `mediaPreload.ts`'s existing `ipcRenderer.send('view-error', ...)` pattern. `playHLS.ts` calls it in the two dead-end cases:

- **`hls-unsupported`** — neither hls.js nor native HLS playback is available.
- **`src-rejected`** — the `src` fails the `ALLOWED_SRC_PATTERN` scheme allowlist.

Reporting the error transitions the view to `displaying.error` immediately; leaving the `loading` state cancels the long load-timeout timer, so the specific message is what the operator sees (no generic-timeout clobber).

## Architecture / security decisions

- **Why `mediaPreload.ts` and not a new preload:** the playHLS view uses `mediaPreload.js` as its preload for the existing media-detection lifecycle (`findMedia`, `VolumeController`, snapshots, `view-loaded`/`view-stalled`). The player still relies on all of that, so the channel belongs there.
- **Closed vocabulary, not free-form text:** `reportError` looks the reason up in a fixed `MEDIA_ERROR_MESSAGES` map and only ever sends the mapped message — never caller-supplied text. `mediaPreload.ts` is also attached to **untrusted remote stream views**, so the strict mapping ensures the worst a page can do through this bridge is put its *own* tile into an error state it could already reach by simply failing to play; it cannot inject arbitrary error text into the operator UI. Unknown reasons are ignored.
- The `ALLOWED_SRC_PATTERN` sink guards inside `loadWithHlsJs`/`loadNatively` are left untouched (defense-in-depth, guarding each DOM/network sink directly); the new `loadHLS` check is purely for prompt error reporting + early return.

## Testing

- `packages/streamwall/src/preload/mediaPreload.test.ts` — the bridge is exposed; known reasons map to their fixed `view-error` messages; unknown reasons are ignored (injection attempt sends nothing).
- `packages/streamwall/src/renderer/playHLS.test.ts` — reports `hls-unsupported` when both playback paths are unavailable, reports `src-rejected` for a disallowed scheme, and does **not** report on the normal playback path.
- Full local quality gate green: `format:check`, `lint`, `typecheck`, full test suite (all workspaces).

An end-to-end drive in the packaged app isn't practical here: the trigger is an engine that can play HLS *neither* via hls.js *nor* natively (Chromium/Electron's bundled engine always can today). The unit tests exercise the exact branches by mocking `Hls.isSupported`/`canPlayType` and the `contextBridge`, consistent with how #291 was covered.

## Risks / breaking changes

None. Purely additive: a new optional `window.streamwallMedia` bridge and two early-return error reports on paths that previously fell through silently. Behavior on all playing/erroring paths that already produced a `<video>` is unchanged.

### DIFF
--- a/packages/streamwall/src/preload/mediaPreload.test.ts
+++ b/packages/streamwall/src/preload/mediaPreload.test.ts
@@ -5,17 +5,32 @@ const executeJavaScript = vi.fn()
 // Never resolves, so the assertions below can prove the visibility spoof
 // does not wait on the view-init round trip before running.
 const invoke = vi.fn(() => new Promise(() => {}))
+const send = vi.fn()
+const exposeInMainWorld = vi.fn()
 
 vi.mock('electron', () => ({
-  ipcRenderer: { invoke, send: vi.fn(), on: vi.fn() },
+  contextBridge: { exposeInMainWorld },
+  ipcRenderer: { invoke, send, on: vi.fn() },
   webFrame: { executeJavaScript, insertCSS: vi.fn() },
 }))
+
+type MediaApi = { reportError: (reason: string) => void }
+
+function importedMediaApi(): MediaApi {
+  const call = exposeInMainWorld.mock.calls.find(
+    ([name]) => name === 'streamwallMedia',
+  )
+  if (!call) throw new Error('streamwallMedia was not exposed')
+  return call[1] as MediaApi
+}
 
 describe('mediaPreload visibility spoofing', () => {
   afterEach(() => {
     vi.resetModules()
     executeJavaScript.mockClear()
     invoke.mockClear()
+    send.mockClear()
+    exposeInMainWorld.mockClear()
   })
 
   it('overrides document.visibilityState/hidden in the page world as soon as the preload script runs', async () => {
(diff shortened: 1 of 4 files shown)


NilsR0711:
Follow-up (out of scope for this PR): #309 — the initial `acquireMedia()` rejection in `mediaPreload.ts` is unhandled, so `findMedia`'s specific 'could not find video' reason never surfaces and generic media-detection failures fall through to the 45s `LOADING_TIMEOUT` instead of ~10s. Noticed while working on this issue; that generic path is untouched here.

PR #315, opened 2026-07-14
fix(streamwall): catch mediaPreload's initial acquireMedia() rejection

### LINKED ISSUE #309: Tech debt: mediaPreload's initial acquireMedia() rejection is unhandled, so its 'could not find video' error never surfaces
**Severity: low**

**Evidence** — in `packages/streamwall/src/preload/mediaPreload.ts`, `main()` kicks off the first media acquisition fire-and-forget:

```ts
if (content.kind === 'video' || content.kind === 'audio') {
  webFrame.insertCSS(VIDEO_OVERRIDE_STYLE, { cssOrigin: 'user' })
  acquireMedia(INITIAL_TIMEOUT)      // <-- not awaited, no .catch
  ipcRenderer.send('view-info', { ... })
}
```

`acquireMedia()` → `findMedia()` `throw new Error('could not find video')` when no `<video>`/`<audio>` element appears within `INITIAL_TIMEOUT` (10s). Because this initial call is neither awaited nor `.catch`-ed (unlike the re-acquisition inside the `emptied` handler, which is awaited), the rejection is an **unhandled promise rejection** in the renderer. `main().catch(...)` does not see it — it only wraps the synchronous body of `main()`, not the detached `acquireMedia` promise.

**Impact** — the specific `'could not find video'` / `'timeout waiting for video to start'` reasons never reach the operator. Instead the view simply never sends `view-loaded`, so the state machine hangs in `loading` until the much longer `LOADING_TIMEOUT` (45s, `viewStateMachine.ts`) fires the generic *"Timed out waiting for view to load"*. So:

1. The more specific error messages in `findMedia` are effectively dead code from the operator's point of view.
2. A stream that will never produce a `<video>` (e.g. a normal web page with no playable media) shows a black tile for ~45s rather than ~10s, and with a less informative message.
3. It produces an unhandled-rejection console warning on every such view.

**Context** — surfaced while implementing #304 / PR #308, which added a prompt error channel for the *HLS-unsupported / rejected-src* case in `playHLS.ts`. That fix does not touch this generic media-detection path, which predates it.

**Suggested fix** — `.catch` the initial `acquireMedia(INITIAL_TIMEOUT)` and forward the error via the same `ipcRenderer.send('view-error', { error })` path `main().catch` already uses, so `findMedia`'s specific reasons surface promptly (and the view leaves `loading` at ~10s instead of ~45s). Guard against double-reporting if both this and a later path fire.

NilsR0711:
## Problem

In `mediaPreload.ts`, `main()` kicks off the first media acquisition fire-and-forget:

```ts
if (content.kind === 'video' || content.kind === 'audio') {
  webFrame.insertCSS(VIDEO_OVERRIDE_STYLE, { cssOrigin: 'user' })
  acquireMedia(INITIAL_TIMEOUT)      // not awaited, no .catch
  ipcRenderer.send('view-info', { ... })
}
```

`acquireMedia()` → `findMedia()` throws (`'could not find video'` / `'timeout waiting for video to start'`) when no `<video>`/`<audio>` element appears within `INITIAL_TIMEOUT` (10s). Because this call is neither awaited nor `.catch`-ed, the rejection becomes an **unhandled promise rejection** in the renderer — `main().catch(...)` only wraps `main()`'s synchronous body, not the detached `acquireMedia` promise.

Impact:
1. `findMedia`'s specific error reasons never reached the operator.
2. A view that will never produce a `<video>` showed a black tile for ~45s (the state machine's generic `LOADING_TIMEOUT`) instead of ~10s, with a less informative message.
3. An unhandled-rejection console warning fired on every such view.

## Solution

Attach a `.catch()` to the initial `acquireMedia(INITIAL_TIMEOUT)` call that forwards the error via the same `ipcRenderer.send('view-error', { error })` path `main().catch` already uses, so `findMedia`'s specific reasons surface promptly (~10s instead of ~45s).

Since PR #308 added a separate `playHLS` → `reportError()` channel for the "can never play" case (which also targets the same view via `view-error`), both paths can now race to report the same failure. Added a module-scoped `hasReportedMediaError` guard, checked/set symmetrically in both `mediaApi.reportError` and the new `.catch` handler, so only the first cause reaches the operator regardless of which fires first.

## Testing

Added three tests to `mediaPreload.test.ts` driving `main()`'s real `acquireMedia` flow with fake timers (video content that never appears in the DOM, so the `INITIAL_TIMEOUT` sleep always wins the race):
- the initial timeout is now reported via `view-error` instead of becoming an unhandled rejection
- a `playHLS`-reported error is not overridden by a later generic timeout
- a generic timeout that already fired is not overridden by a later `playHLS` report

Full quality gate run locally: `format:check`, `lint`, `typecheck` (all workspaces), `test` (all workspaces — 302 tests), and the control-client `build` — all green.

## Risks / breaking changes

None. Purely additive error handling on an existing IPC channel; no public API changes.

Closes #309

### DIFF
--- a/packages/streamwall/src/preload/mediaPreload.test.ts
+++ b/packages/streamwall/src/preload/mediaPreload.test.ts
@@ -1,5 +1,5 @@
 // @vitest-environment happy-dom
-import { afterEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 const executeJavaScript = vi.fn()
 // Never resolves, so the assertions below can prove the visibility spoof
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up: filed #316 for the structurally similar unhandled-rejection risk in the `emptied` re-acquisition path (out of scope here — #309 was specifically about the initial `acquireMedia` call).

PR #326, opened 2026-07-14
fix(control-ui): confirm before a tile resize overwrites another tile's cells

### LINKED ISSUE #267: Confirm before a tile resize overwrites another tile's cells
**Context**: follow-up from #32 (resize handles must honor drag direction, allow shrinking, and not fragment neighbors), fixed in #262.

**Current behavior**

`computeResizeAssignments` (`packages/streamwall-control-ui/src/gridInteractions.ts`) intentionally overwrites any other tile's cells that fall inside the resized box — this is existing, tested behavior (see `gridInteractions.test.ts`'s "overwriting other streams" case) and #262 did not change it. The in-progress resize highlight (`isIdxInResizeBox`) does accurately preview which cells will be overwritten, but nothing stops the commit.

**Problem**

Growing a tile over part of a neighbor tile (e.g. covering 2 of a 2x2 neighbor's 4 cells) silently overwrites those cells and fragments the remainder of the neighbor into separate 1x1 boxes (`boxesFromViewContentMap` groups the leftover cells as new, smaller boxes). There's no confirmation, unlike the existing grid-resize-shrink flow (`handleSetGridSize` in `index.tsx`) which asks for confirmation before a smaller grid would drop assignments.

**Suggested fix**

Add a confirmation (or an explicit modifier-key override) before a resize commit would overwrite cells belonging to a *different* stream than the one being resized — mirroring the `window.confirm(...)` pattern already used for grid-size shrinks. The accurate highlight preview from #262 makes it straightforward to compute "would this overwrite another tile?" before committing in `endResize`.

**Impact**: low — this is a UX safety net, not a correctness bug; the preview already shows what will happen before the user releases the pointer.

NilsR0711:
## Problem

`computeResizeAssignments` intentionally overwrites any other tile's cells that fall inside a resized box — existing, tested behavior that #262 correctly did not change. The in-progress resize highlight (`isIdxInResizeBox`) accurately previews which cells a commit would overwrite, but nothing stopped the commit itself.

Growing a tile over part of a neighbor (e.g. covering 2 of a 2x2 neighbor's 4 cells) silently overwrote those cells and fragmented the remainder of the neighbor into separate 1x1 boxes, with no confirmation — unlike the existing grid-resize-shrink flow (`handleSetGridSize`), which asks for confirmation before a smaller grid would drop assignments.

## Solution

- Added `resizeWouldOverwriteOtherStream` to `gridInteractions.ts`: a pure predicate that reuses the same box-spanning math as `computeResizeAssignments`/`isIdxInResizeBox` to report whether any cell inside the resize box currently belongs to a *different* stream than the one being resized.
- Wired it into `useTileResize`'s `endResize` (the pointer-drag resize commit): when the predicate is true, `window.confirm(...)` gates the commit, mirroring the existing grid-shrink confirmation in `handleSetGridSize`.
- The keyboard resize path (arrow keys) is intentionally left unguarded. It commits one grid cell at a time on every key press with no hover preview to batch against — a confirm dialog per keystroke would break the interaction rather than protect it, and unlike the pointer-drag gesture there's no single "commit" moment to gate.

## Testing

- `gridInteractions.test.ts`: new `resizeWouldOverwriteOtherStream` describe block covering overwrite detection, the empty/self-stream no-op cases, empty-string-as-empty handling, out-of-box cells, and missing map entries.
- `useTileResize.test.tsx`: new tests exercising the real `endResize` pointer-event path — commits silently when nothing would be overwritten, prompts and commits on confirm, and aborts (leaving the doc untouched) on cancel.
- Full local quality gate green: `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm test` (all 5 workspaces), `npm -w streamwall-control-client run build`.

**UI verification note:** this fix lives in `useTileResize`'s pointer-event handling and a pure grid-math predicate; there's no rendering/layout change. Manually driving it end-to-end would require the full multi-process stack (a connected Electron Streamwall uplink providing live grid state + a running control-server), which isn't practical to stand up in this environment. The `useTileResize.test.tsx` coverage renders the real Preact hook, dispatches real `PointerEvent`s, mutates a real `Y.Doc`, and stubs `window.confirm` to assert the exact gating behavior a user would see — the closest practical substitute for a live browser session.

## Risk

Low. Purely additive UX safety net (per the issue: "not a correctness bug"); no changes to the underlying assignment math (`computeResizeAssignments` unchanged) or to any other commit path (drag-move/swap, grid resize, layout presets).

Closes #267

### DIFF
--- a/packages/streamwall-control-ui/src/gridInteractions.test.ts
+++ b/packages/streamwall-control-ui/src/gridInteractions.test.ts
@@ -5,6 +5,7 @@ import {
   computeResizeAssignments,
   computeSwap,
   isIdxInResizeBox,
+  resizeWouldOverwriteOtherStream,
   type SwapBox,
 } from './gridInteractions'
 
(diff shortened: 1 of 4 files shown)


NilsR0711:
Filed a follow-up for the keyboard resize path, which is intentionally left unguarded by this PR (see the PR description): #327

PR #328, opened 2026-07-14
feat(control-server): report caught internal WebSocket errors to Sentry

### LINKED ISSUE #298: DX: report caught internal errors in streamwall-control-server to Sentry
**Type: dx · Priority: low**

Follow-up from #245 (PR #292).

`packages/streamwall-control-server` now has opt-in Sentry crash reporting via `@sentry/node` (`STREAMWALL_SENTRY_ENABLED` + `STREAMWALL_SENTRY_DSN`, see `src/sentry.ts`). It covers:

- process-level `uncaughtException`/`unhandledRejection` (Sentry SDK default)
- Fastify request errors (5xx / uncaught), via `Sentry.setupFastifyErrorHandler(app)`

`src/index.ts` has several `catch` blocks around WebSocket message handling and doc-update delivery that are caught locally and only logged via `console.error`, e.g.:

- `Failed to handle ws message:` (two call sites)
- `Failed to send Streamwall doc update`
- `Failed to send client doc update:`
- `failed to send client state delta`

None of these reach Sentry even when crash reporting is enabled, since they never throw past the try/catch and never go through a Fastify route handler. This mirrors the Electron app's Sentry setup, which also relies purely on the SDK's automatic capture rather than manual `captureException` calls — so this isn't a regression, just a coverage gap worth deciding on deliberately.

**Desired** — evaluate whether these caught errors are worth surfacing to Sentry via explicit `Sentry.captureException(err)` calls (gated on whether Sentry is enabled, e.g. by having `initSentry()`/`sentry.ts` expose a small helper), so operators running with crash reporting on get visibility into WebSocket-level failures, not just process crashes and HTTP 5xxs. If the added noise/cost isn't worth it, document that decision instead.

_Source: follow-up from resolving #245._

NilsR0711:
## Problem

`packages/streamwall-control-server` has opt-in Sentry crash reporting (`src/sentry.ts`, added in #292), but it only covers process-level `uncaughtException`/`unhandledRejection` and Fastify request errors via `setupFastifyErrorHandler`. `src/index.ts` has five `catch` blocks around WebSocket message handling and doc-update delivery that were only ever logged locally via `console.error`:

- `failed to send client state delta`
- `Failed to handle ws message:` (two call sites)
- `Failed to send Streamwall doc update`
- `Failed to send client doc update:`

None of these reach Sentry even with crash reporting enabled, since they never throw past their local `try`/`catch` and never go through a Fastify route handler.

## Solution

Added a small `captureException(err, enabled, client)` helper to `sentry.ts`, mirroring the existing `initSentry()` dependency-injection pattern (`enabled` is the flag `initSentry()` already returns; `client` defaults to the real `Sentry` module, injectable in tests). Wired it into all five catch sites in `index.ts` via a `reportCaughtError` closure created once per `initApp()` call.

**Why all five, not a subset:** I initially assumed the three `ws.send()`-wrapping catches were noise (routine disconnect races) and only the two message-handling catches were worth reporting. Checking `ws`'s actual `send()` implementation disproved that: without a callback, `send()` on a non-`OPEN` socket silently buffers via its internal `sendAfterClose` path instead of throwing — it only throws synchronously for a `CONNECTING`-state socket or a genuine internal error (e.g. payload serialization). None of these five catches guard an expected/routine condition, so all five report.

## Testing

- `sentry.test.ts`: unit tests for `captureException` (both branches — disabled no-ops, enabled forwards to the injected client).
- `sentryReporting.test.ts`: end-to-end test that forces a synchronous `ws.send()` failure on a real state-delta broadcast and asserts it reaches the injected Sentry client. Notably, `@fastify/websocket` bundles its own nested `ws` install separate from the one this package imports directly, so every server-side socket is an instance of *that* nested copy's `WebSocket` class — the test resolves and mocks that exact copy (documented inline) rather than the top-level import, which would never intercept a server-side send.
- `initApp()` gained test-only `sentryEnabled`/`sentryClient` overrides (mirroring the existing `db` injection pattern) so specs can exercise the Sentry-enabled path without a real DSN or mutating `process.env`.
- Full workspace quality gate green locally: `format:check`, `lint`, `typecheck`, `test` (all 5 packages, including the new server tests — 99/99 passing).

## Risk / breaking changes

None. This only adds error-reporting side effects that are inert unless an operator has already opted into `STREAMWALL_SENTRY_ENABLED` + `STREAMWALL_SENTRY_DSN`. No public API changes; the new `sentryEnabled`/`sentryClient` parameters are additive and test-only.

Closes #298

### DIFF
--- a/packages/streamwall-control-server/src/index.ts
+++ b/packages/streamwall-control-server/src/index.ts
@@ -23,7 +23,11 @@ import {
 } from 'streamwall-shared'
 import { Auth, StateWrapper, uniqueRand62 } from './auth.ts'
 import { TokenBucket } from './rateLimiter.ts'
-import { initSentry } from './sentry.ts'
+import {
+  captureException,
+  initSentry,
+  type SentryCaptureClient,
+} from './sentry.ts'
 import {
   applyValidatedDocUpdate,
   type DocUpdateLimits,
@@ -280,7 +284,15 @@ export async function initApp({
   baseURL,
   clientStaticPath,
   db: injectedDb,
-}: AppOptions & { db?: StorageDB }) {
+  sentryEnabled: injectedSentryEnabled,
+  sentryClient,
+}: AppOptions & {
+  db?: StorageDB
+  /** Test-only override so specs can exercise Sentry-enabled paths without a real DSN. */
+  sentryEnabled?: boolean
+  /** Test-only override for the client `captureException(...)` reports to. */
+  sentryClient?: SentryCaptureClient
+}) {
   const expectedOrigin = new URL(baseURL).origin
   const clients = new Map<string, Client>()
   const isSecure = baseURL.startsWith('https')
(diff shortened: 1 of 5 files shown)


NilsR0711:
Filed a follow-up for a small pre-existing logging gap I noticed while wiring this up (three `console.error` calls that don't include the caught `err`): #329. Left out of scope here to keep this PR focused on the Sentry-reporting wiring.

PR #330, opened 2026-07-14
refactor(streamwall): drop dead try/catch around Number() in TwitchBot.onMsg

### LINKED ISSUE #258: Tech debt: dead try/catch around Number() in TwitchBot.onMsg
**Severity: low (code quality, no behavior change)**

In `packages/streamwall/src/main/TwitchBot.ts`, `onMsg` wraps a `Number()` call in a `try`/`catch`:

\`\`\`ts
let idx
try {
  idx = Number(match[1])
} catch (err) {
  return
}
\`\`\`

`Number()` never throws — it returns `NaN` for unparseable input. The `catch` here is unreachable dead code, and `match[1]` is additionally guaranteed to be one-or-more digits by the preceding `VOTE_RE = /^!(\d+)$/` match, so `idx` is always a valid finite number in practice.

**Fix**

Drop the `try`/`catch` and assign directly: `const idx = Number(match[1])`. (If defensive `NaN` handling is ever wanted, check `Number.isNaN(idx)` instead of `try`/`catch`.)

_Source: noticed while implementing the fix for #19 (PR #254)._

NilsR0711:
## Summary
- `TwitchBot.onMsg` wrapped `Number(match[1])` in a `try`/`catch`, but `Number()` never throws — it returns `NaN` for unparseable input — and `match[1]` is already guaranteed to be one-or-more digits by the preceding `VOTE_RE = /^!(\d+)$/` match. The `catch` branch was unreachable dead code.
- Removed the `try`/`catch` and assign `idx` directly.

## Technical solution
No behavior change: `idx` was always a valid finite number before and after this change, since `Number()` cannot throw and the regex match guarantees digit input.

## Testing performed
- Added regression tests for `TwitchBot.onMsg` covering vote tallying on a matching message and no-op on a non-matching message, added *before* the refactor to confirm they pass against the pre-existing behavior.
- `npm run format:check`
- `npm run lint`
- `npm run typecheck`
- `npm run test` (all workspaces)
- `npm -w streamwall-control-client run build`

## Risks / breaking changes
None — pure dead-code removal, no behavior change.

Closes #258

### DIFF
--- a/packages/streamwall/src/main/TwitchBot.test.ts
+++ b/packages/streamwall/src/main/TwitchBot.test.ts
@@ -1,3 +1,4 @@
+import type { PrivmsgMessage } from 'dank-twitch-irc'
 import { EventEmitter } from 'events'
 import type { StreamData } from 'streamwall-shared'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up filed for the transient Windows CI flake hit during this PR's first CI run (unrelated to this change, fixed by a rerun): #333

Issue #333, opened 2026-07-14
ci: flaky Electron install on windows-latest test job ("en-US.pak": The file exists, os error 80)


NilsR0711:
**Severity: low (CI flake, not a product bug)**

While babysitting [PR #330](https://github.com/NilsR0711/streamwall/pull/330) (issue #258), the `Test (windows-latest)` CI job failed on the first run with:

```
Error: failed to create '\\?\D:\a\streamwall\streamwall\node_modules\electron\dist\locales\en-US.pak': The file exists. (os error 80)
...
FAIL src/main/data.test.ts [ src/main/data.test.ts ]
Error: Electron failed to install correctly. Please delete `node_modules/electron` and run "npx install-electron --no" manually.
```

Run: https://github.com/NilsR0711/streamwall/actions/runs/29377249560/job/87233170835

The PR's diff only touched `packages/streamwall/src/main/TwitchBot.ts` and its test file — nothing related to Electron installation or `data.test.ts`. Re-running just the failed job (`gh run rerun --failed`) succeeded with no code changes, confirming this was a transient CI environment issue, likely a race/lock during the `electron` postinstall step on the Windows runner (possibly related to caching or a partially-restored `node_modules` state).

**Impact**

Intermittent, unexplained red CI on `windows-latest` forces a manual rerun before every otherwise-green PR can merge, adding friction and eroding trust in the "Test (windows-latest)" check.

**Suggested solution**

- Check whether the workflow's `actions/setup` composite action caches `node_modules` (including `electron`'s binary download) across runs in a way that can race with a fresh install.
- Consider adding `electron`'s postinstall step (`npx install-electron`) as an explicit, isolated step with retry logic, or clearing any stale `node_modules/electron/dist` before install on Windows runners.
- If this recurs, consider tracking frequency to decide whether a `retry`/`continue-on-error`-with-rerun wrapper around the Windows test job is warranted.

_Source: noticed while resolving #258 (PR #330)._

NilsR0711:
Same class of flake, now also observed on \`ubuntu-latest\` (not just Windows), with a different file:

\`\`\`
Error: failed to create '/home/runner/work/streamwall/streamwall/node_modules/electron/dist/libvk_swiftshader.so': File exists (os error 17)
\`\`\`

[...]

PR #334, opened 2026-07-14
fix(control-ui): block keyboard tile resize that overwrites a neighbor

### LINKED ISSUE #327: Keyboard tile resize still silently overwrites neighboring tiles
**Context**: follow-up from #267 (fixed by #326), which added a confirmation before a pointer-drag tile resize would overwrite another tile's cells.

**Current behavior**

`useTileResize`'s `handleResizeKeyDown` (`packages/streamwall-control-ui/src/useTileResize.ts`) commits a resize via `computeResizeAssignments` on every arrow-key press, with no confirmation. Unlike the pointer-drag path, it has no in-progress hover state to preview a single "commit" moment — each keystroke commits immediately.

#326 intentionally left this path unguarded: gating every keystroke behind `window.confirm(...)` would block the keyboard interaction (each press would require a mouse click to dismiss a dialog), defeating the purpose of keyboard-driven resizing.

**Problem**

A user resizing a tile with arrow keys can still grow it over a neighbor's cells and silently overwrite/fragment that neighbor, exactly the UX gap #267 fixed for pointer drags — just via a different input method.

**Suggested fix**

Needs a UX approach that doesn't require a synchronous confirm per keystroke, e.g.:
- Block (no-op) an arrow-key press that would overwrite another stream's cells, requiring an explicit modifier-key override to proceed — mirroring the "explicit modifier-key override" alternative already suggested in #267 for the pointer path.
- Or surface a transient, dismissible warning near the tile rather than a blocking `window.confirm`.

**Impact**: low — same as #267, a UX safety net rather than a correctness bug. `computeResizeAssignments`'s overwrite behavior itself is intentional, tested behavior (see `gridInteractions.test.ts`).

NilsR0711:
## Problem

`useTileResize`'s `handleResizeKeyDown` (arrow-key tile resize) commits every step immediately with no guard against overwriting a neighboring tile's cells. Growing a tile over part of a neighbor via the keyboard silently claims those cells and fragments the rest of the neighbor — the same UX gap #267 fixed for pointer-drag resizes in #326.

#326 intentionally left the keyboard path unguarded, since gating every keystroke behind a `window.confirm(...)` would require a mouse click to dismiss on every press, defeating the purpose of keyboard-driven resizing.

## Solution

Followed the issue's first suggested approach: block (no-op) an arrow-key step that would overwrite another stream's cells, and require an explicit modifier-key override (Shift) to proceed.

- `handleResizeKeyDown` now checks `resizeWouldOverwriteOtherStream` (already added in #326) before committing a step. If the step would overwrite another stream's cells and Shift isn't held, it's a no-op.
- Holding Shift while pressing the arrow key explicitly overrides the block and commits the step, claiming the neighbor's cells as before.
- Extracted `collectCurrentAssignments` out of `endResize` so both the pointer-drag confirm path and the new keyboard block path share the same "current views → assignments map" setup instead of duplicating it.
- Added a `title` hint ("Arrow keys resize. Hold Shift to overwrite another tile.") on the resize handles so the Shift override is discoverable instead of a silent no-op.

## Architecture decisions

- Chose the modifier-key override over a transient/dismissible warning UI (the issue's second suggestion) — it needed no new UI component and matches the existing `shiftKey`-as-modifier pattern already used elsewhere in this file (`GridControls`'s listen/background-listen click).
- `resizeWouldOverwriteOtherStream` and `computeResizeAssignments` are reused as-is; no changes to `gridInteractions.ts` were needed.

## Testing

TDD: added failing tests first (`test(control-ui): cover keyboard resize overwrite blocking`), then implemented the guard (`fix(control-ui): ...`).

New coverage in `useTileResize.test.tsx`:
- Blocks a keyboard resize step that would overwrite another stream's cells, without calling `window.confirm`.
- Commits the step when Shift is held, overriding the block.

Full quality gate run locally: Prettier, ESLint, `tsc --noEmit` (all packages), full test suite (`npm test` — 99 control-server node:test cases + 227 vitest cases, all passing), and the control-client production build.

## Risks / breaking changes

None. Existing pointer-drag resize confirmation behavior is unchanged (just refactored to share the assignments helper). Keyboard resizes that don't overwrite another stream behave exactly as before.

Closes #327

### DIFF
--- a/packages/streamwall-control-ui/src/ResizeHandles.tsx
+++ b/packages/streamwall-control-ui/src/ResizeHandles.tsx
@@ -62,6 +62,12 @@ const StyledResizeHandles = styled.div`
   }
 `
 
+// A keyboard resize step that would overwrite a neighbor's cells is blocked
+// (see handleResizeKeyDown); this hint surfaces the Shift-key override so
+// that block is discoverable rather than a silent no-op.
+const RESIZE_KEYBOARD_HINT =
+  'Arrow keys resize. Hold Shift to overwrite another tile.'
+
 const RESIZE_HANDLES: {
   handle: ResizeHandle
   className: string
@@ -108,6 +114,7 @@ export function ResizeHandles({
           type="button"
           className={className}
           aria-label={label}
+          title={RESIZE_KEYBOARD_HINT}
           disabled={disabled}
           onPointerDown={(ev) => {
             if (!disabled) {
--- a/packages/streamwall-control-ui/src/useTileResize.test.tsx
+++ b/packages/streamwall-control-ui/src/useTileResize.test.tsx
@@ -71,8 +71,15 @@ function fakePointerDown(button = 0): PointerEvent {
   } as unknown as PointerEvent
 }
 
-function fakeKeyDown(key: string): KeyboardEvent {
-  return { key, preventDefault: () => {} } as unknown as KeyboardEvent
+function fakeKeyDown(
+  key: string,
+  options: { shiftKey?: boolean } = {},
+): KeyboardEvent {
+  return {
+    key,
+    shiftKey: options.shiftKey ?? false,
+    preventDefault: () => {},
+  } as unknown as KeyboardEvent
 }
 
 // 2-column grid. sharedState views mirror the doc's seeded assignments.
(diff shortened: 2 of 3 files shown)


NilsR0711:
Note for reviewers: `Test (ubuntu-latest)` failed on first push with an unrelated CI flake (Electron binary install race in `packages/streamwall`'s test run — see #333 for the tracked issue and a comment I added there with this occurrence's details). Re-ran the failed job (`gh run rerun --failed`) with no code changes and it passed; all checks are now green.

Issue #225, opened 2026-07-14
bug: control client has a pre-existing horizontal overflow on desktop


NilsR0711:
**Type: bug · Priority: low**

The web control client overflows horizontally on wide viewports. At a 1280px viewport the document `scrollWidth` is 1314px (the `<body>` measures 1306px wide), producing a horizontal scrollbar.

Root cause is independent of any responsive work: the default 8px `<body>` margin is never reset (`GlobalStyle` sets `html, body { display: flex; flex: 1 }` but no `margin: 0`), and the root layout is pinned to the viewport height, so the flex shell ends up slightly wider than the viewport. The old root container used `overflow: hidden`, which only clipped its *internal* overflow — the body-level overflow remained.

**Reproduction**
1. Run the control client (or the `dev.html` harness in `streamwall-control-client`).
2. Open at a desktop width (e.g. 1280px).
3. Observe a horizontal scrollbar; `document.documentElement.scrollWidth > window.innerWidth`.

**Verified pre-existing:** reverting the responsive layout work from #81 to the original `ControlUI` reproduces the exact same overflow (scrollWidth 1314, body 1306), so this is not introduced by the mobile-friendly change.

**Suggested fix:** reset the body margin in `GlobalStyle` (`body { margin: 0 }`) and re-check the shell width math against the viewport. Confirm no vertical `100vh` + margin overflow remains either.

_Found while resolving #81._

PR #342, opened 2026-07-15
test(control-e2e): add Playwright smoke tests for the control client

### LINKED ISSUE #55: Add a Playwright E2E smoke test for the control client against the control server
**Priority: low**

The full stack (Vite client → Fastify/WS → mocked Streamwall socket) has zero end-to-end coverage; invite/session cookie flow and WS reconnect only manifest in a real browser.

**Scope (2-3 tests)** — admin invite link → page loads → grid renders from injected state; unauthorized access rejected; a grid-cell edit emits the expected WS message (assert on a fake Streamwall peer). CI behind `npx playwright install --with-deps chromium`. Do **not** attempt Electron E2E initially.

_Source: test analysis_

NilsR0711:
## Problem

The full control stack — Vite client → Fastify/WebSocket server → mocked Streamwall socket — had **zero** end-to-end coverage (issue #55). The invite/session cookie flow, WebSocket auth/reconnect, the Yjs grid-edit round-trip, and real-layout regressions only manifest in a real browser, so happy-dom/jsdom unit tests can't reach them. The motivating example is the #225/#239 horizontal-overflow bug, whose `min-width: 0` half only showed up when measuring `scrollWidth` vs `innerWidth` in a real browser.

## Solution

A new **`streamwall-control-e2e`** workspace with a Playwright suite that boots the whole stack in a real Chromium browser:

- **`tests/harness.ts`** — boots `initApp` on an ephemeral port with in-memory storage and the built client's `dist/`, connects a **fake Streamwall uplink** (the trusted state authority), pushes an admin `state` message, and seeds a 3×3 grid of view cells into the shared Yjs doc. Exposes `createInviteLink()`, `waitForViewAssignment()` (observes the peer's mirror of the doc), and a Playwright fixture that gives each test a fresh, isolated server + uplink.
- **`tests/smoke.spec.ts`** — four smoke tests:
  1. an admin invite link signs in (session cookie) and renders the grid from injected state,
  2. an unauthenticated visitor is rejected with the unauthorized banner,
  3. a grid-cell edit propagates over the wire to the Streamwall peer as a shared Yjs doc update,
  4. the layout never overflows horizontally across viewport widths (360/768/1024/1440) — regression cover for #225/#239.

## Architecture decisions

- **Dedicated workspace, no `test` script.** Playwright is heavy and browser-only; keeping it in its own package with a `test:e2e` (not `test`) script keeps it out of the cross-OS `npm test` matrix (which runs on Windows/macOS too, where `--with-deps` isn't available). Its `typecheck` still runs in the normal `checks` job.
- **In-process server + uplink** (imported via `--import tsx`) rather than a subprocess harness, so tests hold direct references to the fake peer and assert on decoded Yjs updates without a side-channel API.
- **Deterministic seed barrier.** The server doesn't echo the uplink's *own* seed back (its doc-update listener is wired up after the initial queued messages are processed), so the harness confirms the shared doc is seeded via a throwaway probe client rather than a fixed delay.
- **Linux-only CI job** behind `npx playwright install --with-deps chromium`, wired into the existing `CI OK` gate. Per the issue, Electron E2E is intentionally out of scope.

## Testing performed

- `npm run format:check`, `npm run lint`, `npm run typecheck` — clean across all workspaces.
- `npm test` — all existing suites pass; the e2e package is correctly skipped (no `test` script).
- `npm -w streamwall-control-client run build` — clean.
- `npm -w streamwall-control-e2e run test:e2e` — **4/4 passing** in a real Chromium.
- `actionlint` on the updated workflow — clean.

## Risks / breaking changes

None to shipped code — this is additive test infrastructure. Adds a new workspace, its dev-dependencies (`@playwright/test`, etc.), and one CI job. The E2E job downloads a Chromium browser on CI (cached by the setup action's npm cache; the browser binary itself is fetched per run).

Closes #55

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,26 @@ jobs:
       - uses: ./.github/actions/setup
       - run: npm -w streamwall-control-client run build
 
+  # End-to-end smoke tests: the control client in a real Chromium browser
+  # against a real control server + fake Streamwall uplink. Linux-only (the
+  # Playwright browser install with system deps is Linux-only), so this stays
+  # out of the cross-OS `test` matrix above.
+  e2e:
+    name: E2E smoke (control client)
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup
+      - name: Install Playwright browser
+        run: npx playwright install --with-deps chromium
+      - name: Run E2E smoke tests
+        run: npm -w streamwall-control-e2e run test:e2e
+
   # Reviews the dependency diff of a PR for known vulnerabilities.
   dependency-review:
     name: Dependency review
@@ -147,7 +167,7 @@ jobs:
   ci-ok:
     name: CI OK
     if: ${{ !cancelled() }}
-    needs: [changes, checks, test, build, dependency-review, actionlint]
+    needs: [changes, checks, test, build, e2e, dependency-review, actionlint]
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,12 @@ storage.json
 .eslintcache
 .npm/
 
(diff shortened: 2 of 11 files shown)


NilsR0711:
### Follow-up issues (out of scope for this PR)

- #343 — extend the E2E suite beyond the initial smoke tests (reconnect/resync, forwarded control commands, role gating, invite round-trip)
- #344 — add stable `data-testid` hooks so the E2E selectors stop depending on styled-components class names

Issue #296, opened 2026-07-14
Tech debt: create-invite response is only checked for errors, not validated against the Invite shape


NilsR0711:
**Severity: low**

**Context** — Issue #279 reported that `handleCreateInvite` (`packages/streamwall-control-ui/src/index.tsx`) blindly cast the `create-invite` command's response to `Invite` without checking for a server-side error. That specific bug is already fixed as a side effect of #287 (`createErrorSurfacingSend`): every `send()` call in `ControlUI`, including `handleCreateInvite`'s, now routes error-shaped responses (`{ error: string }`) to a shared `onError`/`CommandErrorBanner` handler instead of the caller's callback, so a rejected or connection-dropped `create-invite` no longer reaches `setNewInvite`.

**Remaining gap** — the success path is still unvalidated. `packages/streamwall-control-ui/src/index.tsx:1111`:

```ts
(msg) => {
  setNewInvite(msg as Invite) // TODO: validate w/ Zod
},
```

If the control server ever sends a *successful* (non-error) response that doesn't actually match `{ tokenId, name, secret }` (a server-side bug, a future field rename, a version-skew between client and server), the cast silently accepts it and the UI would render a broken invite panel (undefined tokenId/secret/name) with no error surfaced, same failure mode as #279 but from the opposite direction (malformed success vs. dropped error).

**Suggested fix** — introduce real response-shape validation (the existing `// TODO: validate w/ Zod` already flags the intended approach) for `create-invite" and, ideally, other command responses that get cast without a runtime check, so a malformed success response surfaces the same way a `{ error }` response now does via `CommandErrorBanner`.

**Found while resolving** #279.

PR #347, opened 2026-07-15
fix(control-ui): validate invite role before creating invite

### LINKED ISSUE #321: Tech debt: CreateInviteInput casts the selected role string instead of validating it
Context — while resolving #296 (validating the `create-invite` response), a sibling `// TODO: validate` was noticed in the same feature area.

`packages/streamwall-control-ui/src/AccessPanel.tsx:31`:

```ts
onCreateInvite({ name: inviteName, role: inviteRole as StreamwallRole }) // TODO: validate
```

`inviteRole` is `useState`'d as a plain `string` and only ever set from the `<select>`'s hardcoded `<option>` values (`admin`/`operator`/`monitor`), so the cast is currently safe in practice. But it is silently relying on that `<select>`'s options staying in sync with `StreamwallRole`/`validRoles` (`streamwall-shared`) — if a role is ever added, renamed, or removed there without updating this component's hardcoded `<option>` list, the mismatch would only surface as a runtime `unauthorized` rejection from the server (or worse, a stale role silently accepted), not a compile-time or UI-level error.

Suggested fix — derive the `<option>` list from `validRoles` (already exported by `streamwall-shared`) instead of hardcoding it, and/or validate `inviteRole` against `validRoles` before calling `onCreateInvite`, removing the cast in favor of a real runtime check.

Found while resolving #296.

NilsR0711:
## Problem

`CreateInviteInput` (`packages/streamwall-control-ui/src/AccessPanel.tsx`) tracked the selected invite role as a plain `string` and cast it to `StreamwallRole` when calling `onCreateInvite`:

```ts
onCreateInvite({ name: inviteName, role: inviteRole as StreamwallRole }) // TODO: validate
```

This cast was only safe because the `<select>`'s hardcoded `<option>` list happened to match the invitable subset of `StreamwallRole` (`admin`/`operator`/`monitor`). Nothing enforced that relationship at compile time or runtime — if the option list and the role type ever drifted apart, the mismatch would only surface as a runtime `unauthorized` rejection from the server, or worse, a role value that isn't a valid `StreamwallRole` at all being silently forwarded.

Found while resolving #296 (PR #324).

## Solution

- Added `invitableRoles` and a `isInvitableRole` type guard to `streamwall-shared`'s `roles.ts`. `invitableRoles` is `validRoles` minus `local`, since `local` is reserved for the Streamwall app's own IPC connection (`roleCan` already treats it as all-powerful, just like `admin`) and was never meant to be assignable through an invite.
- `CreateInviteInput` now derives its `<option>` list directly from `invitableRoles` instead of a hand-maintained list, so the UI can no longer drift out of sync with the type.
- The submit handler validates `inviteRole` with `isInvitableRole` before calling `onCreateInvite`, giving a real runtime check in place of the unchecked cast, and bails out (without resetting the form) if the value is somehow invalid.

## Testing

- `streamwall-shared`: added tests asserting `invitableRoles` excludes `local` and contains every other role, plus table-driven coverage of `isInvitableRole` for every invitable role, the `local` role, and an arbitrary string.
- `streamwall-control-ui`: added a test asserting the rendered `<option>` list exactly matches `invitableRoles`, and a regression test that forces the select's value into an invalid state (simulating an option/role desync) and asserts `onCreateInvite` is never called with it.
- Full quality gate run locally: `format:check`, `lint`, `typecheck` (all workspaces), `test` (all workspaces) — all green.

## Risks / breaking changes

None. This only tightens client-side validation for the invite-creation form; the set of assignable roles in the UI (`admin`/`operator`/`monitor`) is unchanged from before.

Closes #321

### DIFF
--- a/packages/streamwall-control-ui/src/AccessPanel.test.tsx
+++ b/packages/streamwall-control-ui/src/AccessPanel.test.tsx
@@ -71,6 +71,52 @@ describe('CreateInviteInput', () => {
     expect(nameInput.value).toBe('')
     expect(select.value).toBe('operator')
   })
+
+  test('only offers invitable roles, excluding the local role reserved for the app itself', () => {
+    container = document.createElement('div')
+    document.body.appendChild(container)
+    act(() => {
+      render(<CreateInviteInput onCreateInvite={() => {}} />, container!)
+    })
+
+    const options = Array.from(
+      container.querySelectorAll('option'),
+      (option) => (option as HTMLOptionElement).value,
+    )
+
+    expect(options).toEqual(['admin', 'operator', 'monitor'])
+  })
+
+  test('does not create an invite if the role somehow desyncs from the invitable options', () => {
+    const onCreateInvite = vi.fn()
+    container = document.createElement('div')
+    document.body.appendChild(container)
+    act(() => {
+      render(<CreateInviteInput onCreateInvite={onCreateInvite} />, container!)
+    })
+
+    const nameInput = container.querySelector('input') as HTMLInputElement
+    const select = container.querySelector('select') as HTMLSelectElement
+    const form = container.querySelector('form') as HTMLFormElement
+
+    act(() => {
+      nameInput.value = 'Jamie'
+      nameInput.dispatchEvent(new Event('input', { bubbles: true }))
+    })
+    act(() => {
+      // Not one of the rendered <option>s, so the browser resets the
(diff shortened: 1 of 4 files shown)


NilsR0711:
Filed a follow-up for a related server-side gap noticed while working on this: #349 (the `create-invite` schema still accepts the `local` role; the client UI now correctly excludes it, but the server should reject it too).

PR #348, opened 2026-07-15
fix(control-ui): validate the Yjs doc snapshot in useYDoc against a schema

### LINKED ISSUE #322: Tech debt: useYDoc casts the Yjs doc snapshot to T without runtime validation
Context — while resolving #296 (validating the `create-invite` command response instead of casting it), a related `// TODO: validate using zod` was noticed in the collaborative-state hook.

`packages/streamwall-control-ui/src/useYDoc.ts:37-38`:

```ts
// TODO: validate using zod
setDocValue(valueCopy as T)
```

`useYDoc<T>` builds `valueCopy` by walking the caller-supplied `keys` and calling `.toJSON()` on each corresponding `Y.Map`, then casts the result straight to the generic `T` with no runtime check. `T` is instantiated per call site (e.g. `CollabData` in `index.tsx`). Because the Yjs doc is shared/collaborative state — written to by every connected client and, on the Electron side, synced from the main process — a malformed or unexpected shape (a stale client on an old schema, a manual edit, a future field rename) would be accepted silently and only surface later as a `TypeError` or corrupted UI state somewhere downstream, far from the actual cause.

Suggested fix — accept a Zod schema (or a validator function) per `useYDoc` call site and `safeParse` `valueCopy` before accepting it, falling back to the previous known-good value (or `undefined`) and logging/surfacing a warning on mismatch, mirroring the approach used for `create-invite` in #296 and for control commands in `streamwall-shared/src/schemas.ts`.

Found while resolving #296.

NilsR0711:
## Problem

`useYDoc<T>` built `valueCopy` by walking the caller-supplied `keys` and calling `.toJSON()` on each corresponding `Y.Map`, then cast the result straight to the generic `T` with no runtime check:

```ts
// TODO: validate using zod
setDocValue(valueCopy as T)
```

The Yjs doc is shared/collaborative state — written to by every connected client and, on the Electron side, synced from the main process. A malformed or unexpected shape (a stale client on an old schema, a manual doc edit, a future field rename) was accepted silently and would only surface later as a `TypeError` or corrupted UI state, far from the actual cause.

## Solution

- `useYDoc` now takes a required Zod `schema` parameter and `safeParse`s the doc snapshot before accepting it.
- A snapshot that fails validation is dropped: the previous known-good `docValue` is kept (or stays `undefined` before the first valid snapshot), and a warning is logged with `zod`'s `prettifyError`. This mirrors the approach already used for `create-invite` (#324) and for control commands in `streamwall-shared/src/schemas.ts`.
- Extracted `collabDataSchema`/`CollabData` into a new `packages/streamwall-control-ui/src/collabData.ts` as the single source of truth for the shared-state shape, and updated all three `useYDoc` call sites (`streamwall-control-ui`'s dev harness, `streamwall-control-client`'s websocket connection, and `streamwall`'s Electron IPC renderer) to pass it.
- Two local ad-hoc interfaces that duplicated the exact same shape (`CollabViews` in `useTileResize.ts`, `ViewSlot` in `viewPlacement.ts`) now derive from `CollabData` instead of re-declaring it, removing the duplication that the original `TODO` had already hinted at.

## Architecture decisions

- Schema is a required parameter (not optional/defaulted) so every call site is forced to declare what shape it expects, matching the issue's suggested approach and the `create-invite` precedent.
- On validation failure the hook keeps the *previous* value rather than falling back to `undefined` unconditionally — a single corrupt update from one client shouldn't blank out the whole UI for everyone if a good value is already in hand.
- `collabDataSchema` lives in `streamwall-control-ui` (next to the `CollabData` type it was extracted from) rather than in `streamwall-shared`, since it validates a Yjs-internal wire shape used only by the three `useYDoc` callers, not the cross-boundary control-channel messages `streamwall-shared/schemas.ts` is scoped to.

## Testing performed

- Added `useYDoc.test.tsx` cases: accepts a valid update, rejects an update failing validation while logging a warning and keeping the last known-good value, and confirms a later invalid update doesn't clobber a non-empty known-good value.
- Added `collabData.test.ts` covering `collabDataSchema` acceptance/rejection cases (empty views, valid/invalid `streamId`, missing/non-object snapshots).
- Full quality gate green: `npm run format:check`, `npm run lint`, `npm run typecheck` (all workspaces), `npm run test` (all workspaces), and a `streamwall-control-client` production build.

## Risks / breaking changes

None expected — this only tightens runtime validation on state that was already expected to match `CollabData`; well-formed docs (the only kind produced by the current client/server code) are unaffected. `useYDoc` is not part of any published/external API (all three packages are private workspace packages), so the added required `schema` parameter isn't a public breaking change.

Closes #322

### DIFF
--- a/packages/streamwall-control-client/src/dev.tsx
+++ b/packages/streamwall-control-client/src/dev.tsx
@@ -12,6 +12,7 @@ import { render } from 'preact'
 import { useEffect } from 'preact/hooks'
 import {
   type CollabData,
+  collabDataSchema,
   ControlUI,
   GlobalStyle,
   type StreamwallConnection,
@@ -174,7 +175,7 @@ function useMockConnection(): StreamwallConnection {
     docValue: sharedState,
     doc: stateDoc,
     undoManager,
-  } = useYDoc<CollabData>(['views'])
+  } = useYDoc<CollabData>(['views'], collabDataSchema)
   const appState = useStreamwallState(demoState)
 
   useEffect(() => {
--- a/packages/streamwall-control-client/src/useStreamwallWebsocketConnection.ts
+++ b/packages/streamwall-control-client/src/useStreamwallWebsocketConnection.ts
@@ -2,6 +2,7 @@ import { useCallback, useEffect, useRef, useState } from 'preact/hooks'
 import ReconnectingWebSocket from 'reconnecting-websocket'
 import {
   type CollabData,
+  collabDataSchema,
   type StreamwallConnection,
   useStreamwallState,
   useYDoc,
@@ -32,7 +33,7 @@ export function useStreamwallWebsocketConnection(
     doc: stateDoc,
     setDoc: setStateDoc,
     undoManager,
-  } = useYDoc<CollabData>(['views'], 'server')
+  } = useYDoc<CollabData>(['views'], collabDataSchema, 'server')
   const [streamwallState, setStreamwallState] = useState<StreamwallState>()
   const appState = useStreamwallState(streamwallState)
 
--- a/packages/streamwall-control-e2e/tests/harness.ts
+++ b/packages/streamwall-control-e2e/tests/harness.ts
@@ -86,6 +86,7 @@ const E2E_STATE: StreamwallState = {
   views: [],
   streamdelay: null,
   layoutPresets: [],
+  favorites: [],
   dataSourceHealth: [],
 }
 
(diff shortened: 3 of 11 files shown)


NilsR0711:
Rebased onto `main` (which had moved forward with #341/#342 since this branch was created) and included an unrelated one-line fix for a pre-existing `main`-branch typecheck break in `packages/streamwall-control-e2e/tests/harness.ts` (missing `StreamwallState.favorites`) that was blocking this PR's required CI check. Tracked separately in #350.

Issue #350, opened 2026-07-15
CI: streamwall-control-e2e typecheck fails on main (harness.ts missing StreamwallState.favorites)


NilsR0711:
## Context

Found while resolving #322 (validating the Yjs doc snapshot in `useYDoc`). Rebasing that PR's branch onto the current `main` picked up #341 (stream presets and favorites) and #342 (Playwright E2E harness for the control client), and CI's "Static checks" job then failed on `npm run typecheck` with:

```
tests/harness.ts(71,7): error TS2741: Property 'favorites' is missing in type '{ ... }' but required in type 'StreamwallState'.
```

Confirmed this is not specific to that branch: a clean checkout of `main` (`ad93025`) also fails `npm run typecheck` with the same error.

## Root cause

`packages/streamwall-control-e2e/tests/harness.ts` builds a hardcoded `E2E_STATE: StreamwallState` mock for the fake uplink. #341 added a required `favorites: string[]` field to `StreamwallState` (`packages/streamwall-shared/src/types.ts`), but `E2E_STATE` in `harness.ts` (added by #342, merged around the same time) was never updated to include it.

## Impact

`npm run typecheck` fails on `main` for any workspace/branch that includes the `streamwall-control-e2e` package (it was added to the root `workspaces` list by #342). Since the repo's `main-protection` ruleset requires the "CI OK" check, this currently blocks every PR's CI from going green, regardless of what the PR actually changes.

## Suggested fix

Add `favorites: []` to `E2E_STATE` in `packages/streamwall-control-e2e/tests/harness.ts`, mirroring the empty-favorites default already used in `streamwall-control-ui`'s `useStreamwallState` fallback.

## Note

This was already fixed as an unrelated, out-of-scope commit on #348 (the #322 PR) purely to unblock that PR's required CI check. Filing this issue per policy to track the root cause and its fix on the record, since #348's scope is the `useYDoc` validation work, not this harness mock.

NilsR0711:
Already fixed on `main` as an out-of-scope commit in #348 (which added the missing `favorites: []` field to `E2E_STATE` in `packages/streamwall-control-e2e/tests/harness.ts`).

Verified against current `main` (`95cc429`): `git diff origin/main -- packages/streamwall-control-e2e/tests/harness.ts` is empty, and `npm run typecheck` passes cleanly across all workspaces, including `streamwall-control-e2e`.

No further action needed; closing per the issue's own note that this was filed to track the root cause on the record.

PR #353, opened 2026-07-15
fix(streamwall-shared): validate create-invite role against invitableRoles

### LINKED ISSUE #349: Tech debt: create-invite schema allows the local role, which should never be invitable
Context — found while resolving #321 (removing the unsafe role cast in `CreateInviteInput`).

`packages/streamwall-shared/src/schemas.ts:185`:

```ts
z.object({
  type: z.literal('create-invite'),
  role: z.enum(validRoles),
  name: z.string(),
}),
```

The `create-invite` message schema validates `role` against the full `validRoles` list, which includes `'local'`. `local` is reserved for the Streamwall app's own IPC connection (`roleCan` in `packages/streamwall-shared/src/roles.ts` treats it as all-powerful, exactly like `admin`) and was never meant to be assignable to an invited user — the control UI's invite form already excludes it from its role options (see #321, which introduced `invitableRoles`/`isInvitableRole` in `streamwall-shared` for exactly this purpose).

Today the client UI can't send `role: 'local'` since its `<select>` only offers `invitableRoles`, but the server-side schema is the actual trust boundary: any other client speaking the control protocol directly (not just the bundled UI) could request a `local`-role invite token and the server would accept it. Since `local` and `admin` are equally unrestricted in `roleCan`, this isn't a privilege escalation beyond what an `admin`-role invite already grants — but it does let a `local` token leak into a remote client, which is semantically wrong (the control UI hides the connection-status/role display specifically because it assumes `role === 'local'` only ever means the trusted local Electron connection, see `packages/streamwall-control-ui/src/index.tsx:726`).

Suggested fix — change the `create-invite` schema's `role` field to `z.enum(invitableRoles)` (now exported from `streamwall-shared`) instead of `z.enum(validRoles)`, so the server rejects `local`-role invite requests at the validation boundary rather than relying on the UI to never offer that option.

NilsR0711:
## Problem

`packages/streamwall-shared/src/schemas.ts`'s `create-invite` control command schema validated `role` against the full `validRoles` list, which includes `'local'`. `local` is reserved for the Streamwall app's own IPC connection (`roleCan` in `roles.ts` treats it as all-powerful, exactly like `admin`) and was never meant to be assignable to an invited user.

Today the client UI can't send `role: 'local'` since its invite form only offers `invitableRoles` (see #321/#347), but the server-side schema is the actual trust boundary: `controlCommandMessageSchema.safeParse` is the only check standing between a raw client message and the desktop, since an `admin`-role client passes every `roleCan` authorization check regardless of the command. Any other client speaking the control protocol directly (not just the bundled UI) could therefore request a `local`-role invite token and the server would accept it — letting a `local` token leak into a remote client, which is semantically wrong (the control UI hides connection-status/role display specifically because it assumes `role === 'local'` only ever means the trusted local Electron connection).

## Solution

Changed the `create-invite` schema's `role` field from `z.enum(validRoles)` to `z.enum(invitableRoles)` (exported from `streamwall-shared`'s `roles.ts` since #347), so the server rejects `local`-role invite requests at the validation boundary.

No breaking type changes: `InvitableRole` is a subset of `StreamwallRole`, so every existing typed caller (`handleCreateInvite` in `streamwall-control-ui`, the server's `auth.createToken` call) remains assignable without modification.

## Testing

- Added a test asserting `controlCommandMessageSchema` rejects a `create-invite` command with `role: 'local'` (previously accepted).
- Full quality gate run locally: `format:check`, `lint`, `typecheck` (all workspaces), `test` (all workspaces), and the control-client `build` — all green except a pre-existing, unrelated `streamwall-control-e2e` typecheck failure already tracked in #350 (introduced on `main` by an unrelated merge, not touched by this change).

## Risks / breaking changes

None. This only tightens server-side validation for invite creation; the set of assignable roles is unchanged from what the UI already offers.

Closes #349

### DIFF
--- a/packages/streamwall-shared/src/schemas.test.ts
+++ b/packages/streamwall-shared/src/schemas.test.ts
@@ -296,6 +296,17 @@ describe('controlCommandMessageSchema', () => {
     ).toBe(false)
   })
 
+  test('rejects create-invite with the local role', () => {
+    expect(
+      controlCommandMessageSchema.safeParse({
+        id: 1,
+        type: 'create-invite',
+        role: 'local',
+        name: 'x',
+      }).success,
+    ).toBe(false)
+  })
+
   test('accepts a save-layout-preset command with a non-empty name', () => {
     expect(
       controlCommandMessageSchema.safeParse({
--- a/packages/streamwall-shared/src/schemas.ts
+++ b/packages/streamwall-shared/src/schemas.ts
@@ -1,6 +1,6 @@
 import { z } from 'zod'
 import { GRID_MAX, GRID_MIN } from './geometry.ts'
-import { validRoles } from './roles.ts'
+import { invitableRoles } from './roles.ts'
 
 /**
  * Runtime schemas for every piece of external, untrusted input that crosses a
@@ -182,7 +182,7 @@ export const controlCommandSchema = z.discriminatedUnion('type', [
   }),
   z.object({
     type: z.literal('create-invite'),
-    role: z.enum(validRoles),
+    role: z.enum(invitableRoles),
     name: z.string(),
   }),
   z.object({


NilsR0711:
Follow-up filed for a related type-safety gap noticed while fixing this: #354 (the hand-maintained `ControlCommand` type in `types.ts` isn't derived from the zod schema and would have let `create-invite`'s `role` stay typed as bare `string` even after this fix).

PR #368, opened 2026-07-15
feat(streamwall): double-click a grid stream to expand it fullscreen

### LINKED ISSUE #362: Double-click a stream in the grid to expand it fullscreen, double-click again to restore
## Summary

In the Streamwall grid view (where the running streams are displayed), double-clicking a stream cell should expand it to fill the whole grid area. Double-clicking the expanded stream again should restore it back to the normal grid layout.

## Motivation

Currently there is no quick way to focus on a single stream without changing the overall grid layout/config. A double-click expand/collapse toggle would let a user quickly inspect one stream in detail and then return to the full overview.

## Expected behavior

- Double-click on a stream cell in the grid → that stream expands to occupy the full grid area (other cells hidden/paused as appropriate).
- Double-click again on the expanded stream → it returns to its previous position in the grid, restoring the normal layout.

## Notes

- Should not conflict with existing single-click / drag interactions in the grid.
- Behavior/state should be local to the view and not require changes to the shared grid config that gets persisted.

NilsR0711:
## Problem

There was no quick way to focus on a single stream on the wall without editing the grid layout/config. Issue #362 asks for a double-click expand/collapse toggle so an operator can inspect one stream in detail and then return to the full overview.

## Solution

Double-clicking a running stream tile in the control grid expands that stream to fill the whole wall; double-clicking again restores the normal grid layout.

The expansion is a **transient, runtime-only override** in the main process. A new `fullscreenViewIdx` field on `StreamwallState` records which view is expanded (or `null`) and is broadcast so every client renders the zoom consistently, but the persisted per-cell grid assignments (the Yjs `views` map) are **never** touched — a collapse restores the previous layout verbatim.

### How it works

- **New `set-view-fullscreen` command** (`{ viewIdx, fullscreen }`): added to the control-command schema, gated to operator+ via `roles`, and added to the uplink allowlist so remote control works too.
- **Layout override**: `fullscreenViewContentMap` fills every grid cell with the expanded stream's content; the existing `boxesFromViewContentMap` then merges them into a single wall-spanning box. `StreamWindow.setViews` reuses the already-running view (no reload of the focused stream) and tears down the others ("other cells hidden/paused").
- **Trigger**: the double-click handler lives on the `GridControls` tile layer. It ignores double-clicks on the corner controls (blur/volume/listen/…) and is role-gated. It does **not** conflict with the existing single-click / drag-move gestures — a drag only commits past a 5px threshold, so a stationary double-click is a no-op for dragging.
- **Correct reflection**: `resolveAnchorIdx` maps the spanning box back to the expanded cell, so the preview box and controls show the correct stream even when the expanded stream was not originally in cell 0.

## Architecture decisions

- **Runtime-only, broadcast state** (rather than mutating the persisted grid, or a purely control-UI-local overlay): satisfies the issue's "local to the view, not persisted" requirement while still driving the real wall (pausing other streams) and keeping every connected client in sync.
- **Reuse `boxesFromViewContentMap`** instead of a new spanning-layout code path: the existing view-reuse machinery keeps the focused stream playing across expand/collapse; only the other streams are torn down and reload on restore.

## Testing

- `streamwall-shared`: `set-view-fullscreen` schema accept/reject, operator/monitor role matrix, uplink allowlist, and `fullscreenViewContentMap` → single full-grid box (incl. 1×1).
- `streamwall` (main): `StreamWindow.setViews` with the fullscreen map reuses the running actor, spans it across all cells, and tears down the others.
- `streamwall-control-ui`: `GridControls` double-click fires the toggle (and is ignored on control buttons / for monitors); `ControlUI` integration test sends the correct expand/collapse command and renders the expanded stream rather than the cell-0 assignment; `resolveAnchorIdx` unit tests.
- Full quality gate green locally: format, lint, typecheck, all package test suites, and the control-client build.

## Risks / breaking changes

- Adds a required `fullscreenViewIdx` field to `StreamwallState`; all state producers/fixtures were updated. No wire-format break for existing clients beyond the new field, which is diffed like any other primitive.
- Collapsing re-creates the non-focused views, so those streams reload (the focused stream does not). Acceptable and consistent with how layout-preset loads already work.

Closes #362


NilsR0711:
Follow-up (out of scope for this PR): #369 — preserve non-focused streams when collapsing a fullscreen view, to avoid reloading the other tiles on restore.

PR #370, opened 2026-07-15
docs(control-server): add Docker + Caddy self-hosting deployment guide

### LINKED ISSUE #367: Add self-hosting deployment guide: Docker + Caddy TLS reverse proxy for control-server
## Problem

Streamwall's control server (`streamwall-control-server`) is already built for remote, multi-user access: it multiplexes the desktop app and web control clients over WebSockets, serves the control UI, and ships role-based auth, rate limiting, and Helmet headers. It also supports HTTPS (the scheme of `STREAMWALL_CONTROL_URL` selects http/https behaviour) and binds to a configurable hostname/port.

However, there is **no deployment story for reaching it from outside `localhost`**. Today a user who runs Streamwall on one PC can only control the wall from that same machine. To control the wall from a phone or another device over a domain, they would have to figure out containerization, TLS termination, and the desktop app's uplink configuration on their own. There is currently no Dockerfile, no compose file, no reverse-proxy example, and no end-to-end "host it behind your own domain" guide.

This is packaging + documentation only — no new feature code. It turns the existing architecture into something a user can actually self-host.

Related: #65 (CLOSED) covered the build story and `DB_PATH` location; `DB_PATH` is now anchored to the home directory. This issue is the remaining, distinct piece: **make the server reachable over a domain with TLS.**

## Scope

- [ ] **`Dockerfile`** for `streamwall-control-server` (production build, not `tsx` from source; run as non-root; sane default `DB_PATH` on a mounted volume).
- [ ] **`docker-compose.yml`** wiring the control server together with a **Caddy** reverse proxy for automatic Let's Encrypt TLS.
  - Decision: use **Caddy** rather than Traefik/nginx — smallest config, automatic HTTPS out of the box, best fit for a self-hoster.
- [ ] **`.env.example`** with the relevant vars documented (`STREAMWALL_CONTROL_URL=https://wall.example.com`, hostname `0.0.0.0`, port, DB path, rate-limit knobs).
- [ ] **Uplink documentation**: how the desktop Streamwall app connects to the remote control server (over `wss://` against the public domain) instead of localhost.
- [ ] **Self-hosting section in the docs/README**: prerequisites (a domain + open ports 80/443), first-run admin invite handling, and an operational security note (TLS required, keep the admin invite private, rate limits are already enforced).
- [ ] Verify the whole path end-to-end: desktop app on machine A → control server behind Caddy/TLS on a host → web control client from a phone on machine B.

## Out of scope

- Streaming the **rendered wall video** to remote devices (the server only syncs state/control; the composited video is rendered locally by the Electron app). Remote *viewing* of the output would be a separate WebRTC feature and is intentionally not part of this issue.

## Notes

- The server already reads `STREAMWALL_CONTROL_HOSTNAME` / `STREAMWALL_CONTROL_PORT` / `STREAMWALL_CONTROL_URL` from env and enforces per-IP rate limiting, WS message-rate caps, and Helmet headers, so the security groundwork exists — this issue is about packaging and documenting a correct deployment on top of it.

NilsR0711:
## Problem

`streamwall-control-server` already has everything needed for remote, multi-user access — role-based auth, rate limiting, Helmet headers, HTTPS-aware behaviour — but there was no deployment story for reaching it from outside `localhost`: no Dockerfile, no compose file, no reverse-proxy example, and no self-hosting guide.

## Solution

Packaging + documentation only, no new feature code:

- **`packages/streamwall-control-server/Dockerfile`** — multi-stage build. Installs only the workspaces the server needs (`streamwall-shared`, `streamwall-control-server`, `streamwall-control-client`, `streamwall-control-ui` — the client imports from the UI package), builds the static control client, prunes dev dependencies, and runs as a non-root user. No compile step for the server itself (Node 22 native TS support, matching the existing manual-deploy instructions). `storage.json` lives on `/data`.
- **`deploy/docker-compose.yml`** + **`deploy/Caddyfile`** — wires the server together with a [Caddy](https://caddyserver.com/) reverse proxy for automatic Let's Encrypt TLS. Caddy over nginx/Traefik: smallest config, automatic HTTPS with no manual ACME setup. The control server is not published to the host — only Caddy is.
- **`deploy/.env.example`** — documents the vars the stack needs (domain, ACME email, public URL, port, plus the optional rate-limit/Sentry knobs already supported by the server).
- **`docs/self-hosting.md`** — prerequisites, quick start, what the first-run uplink/admin-invite links mean, how to point the desktop app's `[control] endpoint` and a web control client at the deployed server, operational security notes, and troubleshooting.
- Cross-linked from `README.md` (new "Self-hosting" subsection), `packages/streamwall-control-server/README.md` (Deployment section), and `example.config.toml` (existing `wss://` reverse-proxy comment).

### Two things discovered while building this, deliberately *not* fixed here (scope was packaging + docs)

1. `STREAMWALL_CONTROL_URL` without an explicit port (a bare `https://domain`, the normal case behind a reverse proxy) makes the server compute a listen port of `0` instead of falling back to the scheme's conventional port. Documented as a required workaround (`STREAMWALL_CONTROL_PORT` must always be set explicitly); filed as a follow-up to fix the underlying default.
2. The server doesn't parse `X-Forwarded-For`, so behind this (or any) reverse proxy its per-IP rate limiting sees every client as the proxy's IP rather than each real visitor. Documented as a known limitation in the self-hosting guide; filed as a follow-up to add configurable `trustProxy` support.

## Testing performed

- `docker build` of the new Dockerfile, then `docker run` — confirmed non-root user, `/data/storage.json` created, correct hostname/port bind, admin-invite/uplink links logged, static client served.
- Full `docker compose up -d --build` of the compose stack with `STREAMWALL_DOMAIN=localhost` — confirmed Caddy obtains a certificate (local CA, since `localhost` isn't a public domain), HTTP→HTTPS redirect, and the reverse proxy correctly reaches the control server (verified via response headers, `via: 1.1 Caddy`, and a WebSocket-path request reaching the server's own routing instead of a 502). Torn down afterwards (`docker compose down -v`, images/containers removed).
- `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm run test` (full workspace suite — unrelated to this change, no server source was modified), `npm -w streamwall-control-client run build` — all pass.

No tests added: this PR is packaging and documentation only, with no new application behavior to cover (per the issue's own scope note).

## Risks / breaking changes

None — no application code changed. New files only (`Dockerfile`, `deploy/`, `docs/self-hosting.md`) plus doc cross-links and one comment addition in `example.config.toml`.

Closes #367


NilsR0711:
Follow-up issues filed for the two things discovered while building this (see PR description): #371 (`STREAMWALL_CONTROL_PORT` resolves to 0 without an explicit port in `STREAMWALL_CONTROL_URL`) and #372 (no `trustProxy`/`X-Forwarded-For` support, so per-IP rate limiting collapses behind a reverse proxy).

PR #373, opened 2026-07-15
fix(streamwall): keep non-focused streams alive across a fullscreen expand/collapse

### LINKED ISSUE #369: Preserve non-focused streams when collapsing a fullscreen-expanded view
## Context

Follow-up from #362 / PR #368, which added double-click-to-expand for a grid stream.

When a view is expanded to fullscreen, the layout override fills every cell with the expanded stream's content and `StreamWindow.setViews` **tears down** the other views (they become `unusedViews` and are `stop()`ed + disposed). The focused stream is reused and keeps playing, but on **collapse** the other streams are re-created from scratch and therefore **reload** (a brief black flash / buffering per tile).

This satisfies the issue's "other cells hidden/paused as appropriate" wording, but a smoother experience would keep the non-focused streams alive (paused/hidden) across expand→collapse so they resume instantly.

## Suggested solution

Instead of routing the non-focused views through `setViews`' teardown path during expansion, keep their actors alive but move them off-screen / behind the expanded view (and optionally pause playback), then reposition them back on collapse. This avoids the reload and reduces wasted network/CPU churn for large grids.

Trade-off to weigh: keeping every stream decoded while hidden costs memory/CPU, so "pause but keep loaded" vs. "fully background" should be a deliberate choice (possibly configurable).

## Impact

- UX: collapsing a fullscreen view currently reloads all other tiles.
- Low priority — the feature works correctly as shipped; this is a polish/perf improvement.

NilsR0711:
## Problem

Follow-up from #362 / PR #368, which added double-click-to-expand for a grid stream.

When a view is expanded to fullscreen, `StreamWindow.setViews` tears down every other view's actor entirely (`stop()` + dispose) since the fullscreen content map has no box for their cells. The focused stream is reused and keeps playing, but on **collapse** the other streams are recreated from scratch and therefore reload (a brief black flash / buffering per tile).

## Solution

`StreamWindow.setViews` now accepts an options argument, `{ parkUnused: true }`:

- Instead of disposing an actor that no longer matches any box, it moves the actor's `WebContentsView` onto its own offscreen host window (mirroring the state machine's existing `offscreenView` action) and keeps the actor in a new `parkedViews` map rather than tearing it down.
- Parked actors are folded back into the next `setViews()` call's reuse candidates (`unusedViews`), so the existing matcher logic (same URL + running) finds and repositions them via a normal `DISPLAY` event once the layout returns to normal — no new view is created, no reload.
- A parked actor whose cell was genuinely cleared while expanded is still torn down normally on the next call (it just isn't matched by any box).
- `parkedViews` is excluded from `emitState()`/`this.views`, so the control UI still only renders the single expanded box during fullscreen — unchanged from before.

`main/index.ts`'s fullscreen branch in `updateViewsFromStateDoc` now passes `{ parkUnused: true }` when overriding the layout for the expanded view. The normal (non-fullscreen) call path is untouched, so cells that are actually removed from the grid are still disposed as before.

## Testing

- Added `StreamWindow.test.ts` coverage (issue #369):
  - a non-focused running view is hidden (moved to its own offscreen host) instead of stopped/disposed when `parkUnused` is requested, and is tracked in `parkedViews`.
  - a parked view is reused (repositioned via `DISPLAY`, no `createView()` call) when the fullscreen view collapses back to the normal layout.
  - a parked view whose cell was cleared while expanded is still torn down on collapse (no permanent leak).
- Full local quality gate: `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm run test` — all green across every workspace.

## Risks / limitations

- Parked (hidden) streams keep decoding/playing while off-screen instead of pausing — this trades a small amount of CPU/network for the instant-resume UX. The issue's own suggested solution flagged "pause while hidden" as an optional, separately-decidable enhancement; I kept this change scoped to the core "don't reload" ask.

Closes #369

### DIFF
--- a/packages/streamwall/src/main/StreamWindow.test.ts
+++ b/packages/streamwall/src/main/StreamWindow.test.ts
@@ -46,6 +46,7 @@ function makeStreamWindow(config: StreamWindowConfig) {
     typeof StreamWindow
   >
   sw.config = config
+  sw.parkedViews = new Map()
   return sw
 }
 
@@ -270,14 +271,24 @@ function makeReuseTestActor(opts: {
   const send = vi.fn()
   const stop = vi.fn()
   const disposeView = vi.fn()
+  const setBounds = vi.fn()
+  const removeChildViewOnWin = vi.fn()
+  const addChildViewOnOffscreen = vi.fn()
+  const view = { setBounds }
+  const win = { contentView: { removeChildView: removeChildViewOnWin } }
+  const offscreenWin = {
+    contentView: { addChildView: addChildViewOnOffscreen },
+    getBounds: () => ({ width: 100, height: 100 }),
+  }
   const actor = {
     getSnapshot: () => ({
       context: {
         id: opts.id,
         content: opts.content,
         pos: { spaces: opts.spaces },
-        view: {},
-        offscreenWin: {},
+        view,
+        win,
+        offscreenWin,
         next: null,
         disposeView,
       },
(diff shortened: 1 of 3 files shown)


NilsR0711:
Follow-up filed: #374 (optionally pause parked/hidden streams during a fullscreen expansion — intentionally out of scope here per the trade-off noted in #369).

PR #375, opened 2026-07-15
fix(control-server): default listen port when URL has no port

### LINKED ISSUE #371: control-server: STREAMWALL_CONTROL_PORT resolves to 0 when STREAMWALL_CONTROL_URL has no explicit port
## Problem

In `runServer` (`packages/streamwall-control-server/src/index.ts`), the listen port is derived like this:

```ts
const url = new URL(baseURL)
const hostname = overrideHostname ?? url.hostname
const port = Number(overridePort ?? url.port ?? '80')
```

`URL.port` returns an **empty string** (not `undefined`) when the URL has no explicit port — which is the normal case for a bare `https://your-domain` public URL (the typical value for `STREAMWALL_CONTROL_URL` when the server sits behind a reverse proxy). Since `??` only falls through on `null`/`undefined`, `'' ?? '80'` evaluates to `''`, and `Number('')` is `0`.

Reproduction:

```js
new URL('https://wall.example.com').port // ''
Number(undefined ?? '' ?? '80')           // 0
```

So with `STREAMWALL_CONTROL_URL=https://wall.example.com` and no explicit `STREAMWALL_CONTROL_PORT`, the server calls `app.listen({ port: 0, host: hostname })` — Node/Fastify will bind to an OS-assigned ephemeral port instead of a predictable one (443 would be the natural expectation for an `https://` URL with no port), which silently breaks any reverse-proxy target that expects a fixed port.

## Impact

Self-hosters following the natural pattern of setting `STREAMWALL_CONTROL_URL` to a bare public HTTPS domain (no port) without *also* separately setting `STREAMWALL_CONTROL_PORT` will end up with the server listening on a random port, with no error — just a working server on an unexpected port.

Currently worked around in [`deploy/.env.example`](../blob/main/deploy/.env.example) (from #367) by explicitly documenting that `STREAMWALL_CONTROL_PORT` must always be set. This issue is about fixing the underlying default so that workaround isn't necessary.

## Suggested fix

Fall back to the scheme's conventional port when `url.port` is empty, e.g.:

```ts
const port = Number(
  overridePort ?? (url.port || (url.protocol === 'https:' ? '443' : '80')),
)
```

or equivalent, and add a unit test covering a `STREAMWALL_CONTROL_URL` with no explicit port for both `http:` and `https:`.

## Where discovered

Found while implementing #367 (Docker + Caddy self-hosting guide) and empirically verified with `new URL(...)`.

YRWoods:
## Summary
Fix listen-port resolution when STREAMWALL_CONTROL_URL has no explicit port so the server no longer binds port 0.

## Problem
Fixes #371 — URL.port is empty string, so Number(overridePort ?? url.port ?? '80') became 0.

## Approach
- Add resolveListenPort() with scheme defaults (https 443, else 80)
- Explicit STREAMWALL_CONTROL_PORT still wins; empty override falls through
- Unit tests for bare hosts, explicit ports, empty override

## How I tested
- node --import tsx --test src/index.test.ts in packages/streamwall-control-server (10 pass)

### DIFF
--- a/packages/streamwall-control-server/src/index.test.ts
+++ b/packages/streamwall-control-server/src/index.test.ts
@@ -2,7 +2,11 @@ import { Low, Memory } from 'lowdb'
 import assert from 'node:assert/strict'
 import { after, describe, test } from 'node:test'
 
-import { initApp, SESSION_COOKIE_NAME } from './index.ts'
+import {
+  initApp,
+  resolveListenPort,
+  SESSION_COOKIE_NAME,
+} from './index.ts'
 import type { StoredData } from './storage.ts'
 
 const ONE_YEAR_IN_SECONDS = 365 * 24 * 60 * 60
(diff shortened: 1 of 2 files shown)


NilsR0711:
Superseded by #378 — rebased onto current `main` and re-opened from this repo so CI/branch protection can run. Thanks @YRWoods for the original implementation and tests.

PR #376, opened 2026-07-15
fix(control-server): opt-in trustProxy for per-client rate limits

### LINKED ISSUE #372: control-server: no X-Forwarded-For / trustProxy support, so per-IP rate limiting collapses behind a reverse proxy
## Problem

`streamwall-control-server` uses `@fastify/rate-limit` keyed by the request's apparent IP address (`STREAMWALL_RATE_LIMIT_MAX` / `STREAMWALL_AUTH_RATE_LIMIT_MAX` — see `packages/streamwall-control-server/README.md`). Fastify is instantiated without a `trustProxy` option anywhere in `packages/streamwall-control-server/src/index.ts`, so it never parses `X-Forwarded-For` (or similar) and always sees the *immediate* TCP peer's address.

When the server is deployed behind any reverse proxy — including the Caddy setup added in #367 (`deploy/docker-compose.yml`) — every request's apparent source IP is the proxy's own address, not the real client's. Per-IP rate limiting then effectively becomes one shared bucket across *all* clients combined, rather than isolating abusive/misbehaving individual clients.

This doesn't disable rate limiting (the global cap still applies), but it's meaningfully coarser than the intended per-client behavior, and it's exactly the deployment topology the project now documents and packages (#367).

## Suggested fix

Add opt-in, configurable proxy trust — e.g. an env var such as `STREAMWALL_TRUST_PROXY` (boolean, or accepting an IP/CIDR per Fastify's `trustProxy` option: https://fastify.dev/docs/latest/Reference/Server/#trustproxy) — passed into the `Fastify({ ... })` constructor options. Keep it **off by default** (current behavior) so servers exposed directly to the internet without a proxy in front aren't suddenly trusting a spoofable `X-Forwarded-For` header from arbitrary clients. Document the new var in `packages/streamwall-control-server/README.md` and update `deploy/docker-compose.yml` / `deploy/.env.example` (from #367) to enable it for the bundled Caddy setup, since Caddy is the only entity able to reach the control-server container there.

## Where discovered

Found while implementing #367 (Docker + Caddy self-hosting guide) — documented as a known limitation in `docs/self-hosting.md` in the meantime.

YRWoods:
## Summary
Opt-in `STREAMWALL_TRUST_PROXY` so per-IP rate limits use real client addresses behind a reverse proxy.

## Problem
Fixes #372 — without `trustProxy`, every request behind Caddy/etc. shares the proxy's IP bucket.

## Approach
- `parseTrustProxy()`: off by default; `true`/`false` or IP/CIDR pass-through for Fastify
- `Fastify({ trustProxy })` in `initApp`
- README security table documents the var and spoof risk
- Test: two `X-Forwarded-For` clients keep separate budgets when trust is on

## How I tested
- `node --import tsx --test src/rateLimit.test.ts` (5 pass)

## Notes
- Deploy `docker-compose` / `.env.example` not on `main` yet (#367) — enable the env when that lands; behavior is ready.
- Default remains **off** for direct internet exposure.

### DIFF
--- a/packages/streamwall-control-server/README.md
+++ b/packages/streamwall-control-server/README.md
@@ -50,6 +50,7 @@ each WebSocket connection. The limits are tunable:
 | `STREAMWALL_RATE_LIMIT_MAX`      | `100`      | Max HTTP requests per IP per window (global).                    |
 | `STREAMWALL_AUTH_RATE_LIMIT_MAX` | `10`       | Stricter max for the `/invite/:id` auth route per IP per window. |
 | `STREAMWALL_RATE_LIMIT_WINDOW`   | `1 minute` | Rate-limit window (any `@fastify/rate-limit` time value).        |
+| `STREAMWALL_TRUST_PROXY`         | off        | Opt-in Fastify `trustProxy` so rate limits use client IPs from `X-Forwarded-For` when behind a reverse proxy. `true`/`false`, or an IP/CIDR list. **Leave off** if the server is exposed directly to the internet (header is spoofable). Enable only when a trusted proxy (e.g. Caddy) is the only hop that can reach the process. |
 | `STREAMWALL_WS_MSG_RATE`         | `100`      | Sustained inbound WebSocket messages per second, per connection. |
 | `STREAMWALL_WS_MSG_BURST`        | `1000`     | Burst allowance of inbound WebSocket messages, per connection.   |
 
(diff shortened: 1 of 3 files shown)


NilsR0711:
Superseded by the new PR — rebased onto current `main` (includes #378) and re-opened from this repo so CI/branch protection can run. Thanks @YRWoods for the original implementation and tests.

PR #383, opened 2026-07-15
feat(streamwall): optionally pause parked views during a fullscreen expansion

### LINKED ISSUE #374: Optionally pause parked (hidden) streams during a fullscreen expansion
## Context

Follow-up from #369 / PR #373, which stopped tearing down non-focused streams when a grid view is expanded to fullscreen (#362). Non-focused views are now moved off the visible wall onto their own offscreen host window (`StreamWindow.hideView` / `parkedViews`) but keep decoding and playing while hidden, so a later collapse can reposition them instantly instead of reloading.

Issue #369's own suggested solution flagged the "keep alive" vs. "pause but keep loaded" vs. "fully background" trade-off as a deliberate choice, possibly configurable, and PR #373 intentionally scoped out pausing to keep the fix minimal and low-risk.

## Suggested solution

While a stream is parked (hidden behind an expanded view), send it some form of "pause" so it stops decoding/consuming CPU and network bandwidth without losing its position/session state, then resume it on unpark (collapse). This would need:

- A new preload-level pause/resume primitive (the current view state machine only has `MUTE`/`BACKGROUND` for audio and `BLUR` for the video element's raster; there's no concept of pausing the underlying `<video>`/HLS playback), or
- Reusing/extending the existing `BLUR` mechanism if it already reduces resource usage sufficiently for this purpose.

Given streams are typically live (not seekable on-demand video), pausing and resuming needs to tolerate falling behind the live edge (or just accept a brief re-buffer on resume, which is still much cheaper than a full page reload).

## Impact

- Low priority — the core "no reload on collapse" issue is already fixed by #373. This is a further CPU/network optimization for walls with many grid cells that get parked during a fullscreen session.

NilsR0711:
## Problem

Follow-up from #369 / PR #373, which stopped tearing down non-focused streams when a grid view is expanded to fullscreen (#362): parked (hidden) views are moved to their own offscreen host window instead of being disposed, so a later collapse can reposition them instantly. But they keep decoding/playing at full rate while hidden -- PR #373 deliberately scoped pausing out to keep that fix minimal, flagging it as a separately-decidable enhancement.

## Solution

Adds an opt-in `--park.pause` flag (default `false`, so existing behavior is unchanged unless explicitly enabled) that pauses a parked view's own media playback instead of keeping it fully live:

- **`viewStateMachine.ts`**: a new `PAUSE`/`RESUME` event pair and `running.pause` parallel region, mirroring the existing `BLUR`/`UNBLUR` pattern (including deferred application if the request arrives while the view is still loading or recovering from an error). Unlike `BLUR` -- a purely cosmetic overlay effect with no action tied to it, and therefore not reusable as-is for this -- entering `paused` sends a `pause` IPC message to the view's `webContents`, and leaving it sends `resume`.
- **`mediaPreload.ts`**: `ipcRenderer.on('pause'/'resume', ...)` handlers act on the currently-acquired `<video>`/`<audio>` element (tracked across re-acquisition, e.g. the `'emptied'` handler). `pause` calls `HTMLMediaElement.prototype.pause` directly rather than `media.pause()`, since `lockdownMediaTags()` permanently shadows the element's own `pause` method with a no-op (to stop sites like Facebook from auto-pausing) -- calling it directly would silently do nothing. `resume` just calls `.play()`; since these are typically live (non-seekable) streams, resuming may briefly re-buffer or land slightly behind the live edge, which is expected and far cheaper than a full reload.
- **`StreamWindow.ts`**: `hideView()` sends `PAUSE` when parking a view (if `pauseParkedViews` is enabled); `setViews()` sends `RESUME` to a view that was parked coming into the call and gets matched back into a box (a fullscreen collapse), also gated on the flag. The parked actor's own state (content, position, retry budget) is untouched either way -- only its playback is paused/resumed, so a later collapse still repositions it instantly via the existing `DISPLAY`-based reuse path from #373; the only change when the option is enabled is a brief resume/re-buffer.
- **`config.ts`/`index.ts`**: wires `park.pause` through the same schema → CLI flag → runtime config chain as `retry.*`.

## Testing

- `viewStateMachine.test.ts`: deferred `PAUSE` while loading/recovering from an error, survival across an automatic stalled reload, `RESUME` clearing `desiredPaused`, and that `sendViewPause`/`sendViewResume` fire (or don't) exactly when expected.
- `mediaPreload.test.ts`: a `pause` message pauses the acquired `<video>` even with an instance-level `pause` override in place (proving the native-implementation bypass works), a `resume` message resumes it, and both no-op safely before any media has been acquired.
- `StreamWindow.test.ts`: parking sends `PAUSE` only when `pauseParkedViews` is enabled, and a previously-parked view reused after collapse sends `RESUME` only when enabled (and never for a view that was never parked).
- Full local quality gate: `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm run test` -- all green across every workspace.

## Risks / limitations

- Off by default -- no behavior change unless an operator opts in.
- Only pauses the `<video>`/`<audio>` element itself; it does not call `hls.stopLoad()` on the bundled HLS player, so segment fetching for the local HLS path isn't stopped outright (it naturally tapers off once hls.js's buffer target is reached while paused, capping most of the network cost). A full `stopLoad()`/`startLoad()` bridge would need a new same-document channel into `playHLS.ts` and was scoped out to keep this change minimal, consistent with how #373 scoped out pausing entirely.
- `content.kind === 'web'` views never acquire a media element, so pause/resume is a no-op for them -- expected, not a gap.

Closes #374

### DIFF
--- a/packages/streamwall/src/main/StreamWindow.test.ts
+++ b/packages/streamwall/src/main/StreamWindow.test.ts
@@ -47,6 +47,7 @@ function makeStreamWindow(config: StreamWindowConfig) {
   >
   sw.config = config
   sw.parkedViews = new Map()
+  sw.pauseParkedViews = false
   return sw
 }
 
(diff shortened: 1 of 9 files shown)


NilsR0711:
Follow-up: #384 tracks extending this with real HLS-level network savings (`hls.stopLoad()`/`startLoad()`), scoped out of this PR to keep the change minimal (see the Risks/limitations section above).

PR #416, opened 2026-07-15
fix(control-server): validate uplink state payload against a strict schema

### LINKED ISSUE #387: Security: Uplink state messages bypass StreamwallState runtime validation
## Summary

When the desktop Streamwall app pushes a `state` message over its authenticated uplink WebSocket, the control-server validates the envelope with Zod but **casts the nested `state` payload to `StreamwallState` without runtime schema validation**. A malformed or adversarial desktop build (or a future regression in state serialization) can wrap garbage in `StateWrapper` and broadcast corrupted snapshots to every connected browser client.

## Problem

In `packages/streamwall-control-server/src/index.ts`, after `controlStateMessageSchema.safeParse`:

```ts
const state = parsed.data.state as unknown as StreamwallState
clientState = new StateWrapper(state)
```

`controlStateMessageSchema` (`streamwall-shared/src/schemas.ts`) only checks:

```ts
state: z.object({}).loose()
```

That guarantees "an object" — not valid views, streams, XState snapshots, or auth shape. The inline comment references a historical client crash when garbage reached `StateWrapper.view()`.

Client-side Yjs updates **are** guarded (`stateDocGuard.ts`, shape checks), but the trusted uplink channel that seeds `StateWrapper` is not.

## Impact

- **Availability:** malformed uplink state can crash or hang browser clients during `view(role)` projection.
- **Integrity:** role-scoped filtering assumes well-formed input; corrupt `views` arrays or missing required fields may leak or hide data unpredictably.
- **Defense in depth:** the uplink token is high-privilege; validating its payloads costs little and catches bugs before they fan out to all sessions.

## Proposed fix

1. Add a `streamwallStateSchema` (or reuse/compose existing per-field Zod schemas) in `streamwall-shared`.
2. After envelope parse, `streamwallStateSchema.safeParse(parsed.data.state)` — reject + log on failure; do not update `StateWrapper`.
3. Add server tests: invalid views shape, missing required keys, extra-unknown dangerous fields.

Consider whether partial updates need a separate schema vs. full snapshot replacement.

## Acceptance criteria

- [ ] Uplink `state` payloads validated with a strict Zod schema before `StateWrapper` mutation
- [ ] Invalid payloads rejected with structured warning (no client broadcast)
- [ ] Regression tests covering at least: empty views, malformed view state machine snapshot, missing streams
- [ ] No `as unknown as StreamwallState` cast remains on this path

NilsR0711:
## Problem

The control-server's uplink handler validated only the *envelope* of a
`state` message (`{ type: 'state', state: <object> }`) and then cast the
nested payload straight to `StreamwallState` with `as unknown as
StreamwallState` -- no runtime check on its actual contents. A malformed
or adversarial desktop build (or a future regression in state
serialization) could wrap garbage in `StateWrapper` and broadcast a
corrupted snapshot to every connected browser client, since
`StateWrapper.view()` assumes a well-formed `StreamwallState`.

## Solution

- Added `streamwallStateSchema` in `streamwall-shared`: a strict Zod
  schema covering every field of `StreamwallState` the server reads --
  `config`, `streams`/`customStreams`, `views` (including the nested
  XState view-machine snapshot shape), `fullscreenViewIdx`,
  `streamdelay`, `layoutPresets`, `favorites` and `dataSourceHealth`.
- In the control-server's uplink handler, the payload is now validated
  against `streamwallStateSchema` right after the existing envelope
  check, for both the initial snapshot and every subsequent update. On
  failure it logs a structured warning and drops the message instead of
  mutating `StateWrapper` -- no cast, no undefined behavior.
- Removed the `as unknown as StreamwallState` cast entirely; the parsed
  result is directly assignable to `StreamwallState`, so schema/type
  drift is now a compile error.

### Architecture note (from the issue's open question)

I checked whether the uplink channel ever sends partial updates
(`Partial<StreamwallState>`) versus full snapshots. It always sends the
full snapshot (`ControlUpdate.state: StreamwallState`, constructed fresh
in `packages/streamwall/src/main/index.ts`), so a single full-snapshot
schema is correct and sufficient -- no separate partial-update schema is
needed.

## Testing

- `streamwall-shared`: new `describe('streamwallStateSchema', ...)`
  block covering a fully-populated valid state, a populated state with
  streams/views/layout presets, a state missing the required `streams`
  field, a malformed view state-machine snapshot, an unrecognized
  top-level view state string, non-object payloads, an unknown identity
  role, an out-of-bounds `fullscreenViewIdx`, and a compile-time
  assignability guard against `StreamwallState`.
- `streamwall-control-server`: new integration tests in
  `wsValidation.test.ts` covering an initial state payload missing a
  required field, a malformed view state-machine snapshot, a
  legitimately empty `views` array (must still be accepted), and a
  malformed *subsequent* update on an already-connected uplink (dropped
  with a structured warning, connection stays alive and still forwards
  valid client commands).
- Updated the shared `VALID_STATE` test fixture (was missing
  `fullscreenViewIdx`, `layoutPresets`, `favorites`, `dataSourceHealth`)
  so it remains a fully valid snapshot under the new schema.
- Full quality gate green: `npm run format:check`, `npm run lint`,
  `npm run typecheck` (all workspaces), `npm test` (all workspaces:
  streamwall 40 files, streamwall-shared 11 files,
  streamwall-control-client 1 file, streamwall-control-server 116
  tests, streamwall-control-ui 38 files), and
  `streamwall-control-client`'s `vite build`.

## Risks / breaking changes

None expected. The new schema mirrors the shape the desktop has always
sent; a well-behaved desktop build's state messages continue to be
accepted unchanged. Only a payload that violates `StreamwallState`'s
own invariants is newly rejected (previously it would have been
accepted and could have crashed a client on `view()`).

Closes #387

### DIFF
--- a/packages/streamwall-control-server/src/index.ts
+++ b/packages/streamwall-control-server/src/index.ts
@@ -19,7 +19,7 @@ import {
   roleCan,
   stateDiff,
   type StreamwallRole,
-  type StreamwallState,
+  streamwallStateSchema,
 } from 'streamwall-shared'
 import { Auth, StateWrapper, uniqueRand62 } from './auth.ts'
 import { TokenBucket } from './rateLimiter.ts'
@@ -588,7 +588,20 @@ export async function initApp({
           )
           return
         }
-        const state = parsed.data.state as unknown as StreamwallState
+
+        // The envelope check above only guarantees "an object". Validate the
+        // full snapshot itself so a malformed or adversarial desktop build can
+        // never seed StateWrapper with garbage that crashes clients on
+        // view() or corrupts the role-scoped projection (issue #387).
+        const stateResult = streamwallStateSchema.safeParse(parsed.data.state)
+        if (!stateResult.success) {
+          console.warn(
+            'Rejected invalid Streamwall state payload:',
+            stateResult.error.issues[0]?.message,
+          )
+          return
+        }
+        const state = stateResult.data
 
         try {
           if (clientState === null) {
--- a/packages/streamwall-control-server/src/testHelpers.ts
+++ b/packages/streamwall-control-server/src/testHelpers.ts
@@ -75,7 +75,11 @@ export const VALID_STATE = {
   streams: [],
   customStreams: [],
   views: [],
+  fullscreenViewIdx: null,
   streamdelay: null,
+  layoutPresets: [],
+  favorites: [],
+  dataSourceHealth: [],
 }
 
 /**
--- a/packages/streamwall-control-server/src/wsValidation.test.ts
+++ b/packages/streamwall-control-server/src/wsValidation.test.ts
@@ -37,6 +37,21 @@ function isCommandType<Type extends ControlCommandMessage['type']>(type: Type) {
 const isBareError = (m: ServerToClientMessage): m is ClientErrorMessage =>
   !('response' in m) && 'error' in m
 
+/** Temporarily replaces `console.warn`, capturing every call made while active. */
+function spyOnConsoleWarn() {
+  const calls: unknown[][] = []
+  const original = console.warn
+  console.warn = (...args: unknown[]) => {
+    calls.push(args)
+  }
+  return {
+    calls,
+    restore: () => {
+      console.warn = original
+    },
+  }
+}
+
 // A per-test override of the update cap must not leak into other test files.
 after(() => {
   delete process.env.STREAMWALL_WS_UPDATE_MAX_BYTES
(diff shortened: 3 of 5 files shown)


NilsR0711:
Follow-up filed for a drift risk noticed while writing `viewStateValueSchema`: #419 (the hand-mirrored XState snapshot shape in `streamwall-shared` has no compile/test-time link to the real machine in `packages/streamwall/src/main/viewStateMachine.ts`; now that it's enforced at runtime, a future shape change there could silently break the uplink connection).

PR #424, opened 2026-07-20
feat(streamwall): stop HLS segment loading for parked views

### LINKED ISSUE #384: Stop HLS segment fetching (not just <video>.pause()) for parked views when --park.pause is enabled
## Context

Follow-up from #374 / PR #383, which added an opt-in `--park.pause` flag that pauses a parked (hidden) view's `<video>`/`<audio>` element instead of keeping it fully live during a fullscreen expansion.

That change calls `HTMLMediaElement.prototype.pause.call(media)` on the acquired media element, which stops decoding/rendering. For the app's bundled HLS player (`packages/streamwall/src/renderer/playHLS.ts`), this does **not** call `hls.stopLoad()` on the underlying hls.js instance, so segment fetching for that path isn't stopped outright -- it naturally tapers off once hls.js's own buffer target (`maxBufferLength`/`maxMaxBufferLength`) is reached while paused, which caps most (but not all) of the network cost.

This was scoped out of #383 to keep that change minimal and testable: `mediaPreload.ts` and `playHLS.ts` are loaded into the same document (not cross-frame), but the `Hls` instance itself lives in a closure inside `loadWithHlsJs()` in `playHLS.ts` and isn't exposed anywhere `mediaPreload.ts` can reach it -- there's no existing bidirectional bridge (the current `window.streamwallMedia` bridge is one-way, page → preload, for error reporting only).

## Suggested solution

Add a same-document channel (e.g. a custom `window` event dispatched from `mediaPreload.ts`'s new `pause`/`resume` IPC handlers, listened for in `playHLS.ts`) so a parked view's HLS player can call `hls.stopLoad()` on pause and `hls.startLoad()` on resume, achieving real network-level savings instead of relying on the buffer-fill taper. `hls.startLoad()` already handles resuming a live stream from wherever hls.js's own live-edge logic lands, so no extra "catch up to live" logic should be needed.

## Impact

Low priority -- `--park.pause` (opt-in, default off) already yields meaningful CPU savings and a natural cap on network usage without this; this is a further optimization on top of an already-optional feature.

NilsR0711:
## Problem

`--park.pause` (opt-in, added in #383 for #374) pauses a parked view's `<video>`/`<audio>` element via `HTMLMediaElement.prototype.pause.call(...)`. That stops decoding and rendering, but for the app's bundled HLS player (`renderer/playHLS.ts`) it does **not** stop segment fetching outright — hls.js keeps loading until its own `maxBufferLength`/`maxMaxBufferLength` target is reached, so network cost only tapers off rather than ending.

This was deliberately scoped out of #383: the `Hls` instance lives in a closure inside `loadWithHlsJs()` and is not reachable from `mediaPreload.ts`, and the existing `window.streamwallMedia` bridge is one-way (page → preload, error reporting only).

## Solution

A same-document channel between preload and page:

- `mediaPreload.ts`'s `pause`/`resume` IPC handlers dispatch `streamwall:media-pause` / `streamwall:media-resume` on `document`.
- `playHLS.ts` listens for them and calls `hls.stopLoad()` / `hls.startLoad()` on its own instance.

## Architecture decisions

- **Dispatched on `document`, not `window`.** Under `contextIsolation` the preload's isolated world has its own `window` object, so a window-dispatched event never reaches the page's main-world listeners. DOM nodes — `document` included — are shared between worlds, which makes this the working channel.
- **Event names live in a dependency-free `preload/mediaParkEvents.ts`.** `playHLS.ts` is renderer bundle code and must not pull in `mediaPreload.ts`'s `electron` imports, so the constants cannot simply sit next to the IPC handlers (the existing import from that module is `import type` only, which erases at build time).
- **Dispatch happens before the `currentMedia` guard**, so an HLS page that has not yet had its `<video>` acquired by the preload still stops loading.
- **Listeners are removed in `teardown()`**, so a destroyed instance is never touched — belt-and-braces alongside the existing `destroyed` flag, which the handlers also check.
- **No catch-up logic on resume.** `hls.startLoad()` already resumes from wherever hls.js's own live-edge logic lands, as noted in the issue.

## Testing

TDD; all five new tests were watched failing first.

`playHLS.test.ts` (mock `Hls` gained a `stopLoad` spy):
- a pause event calls `stopLoad()` exactly once and does not destroy the instance
- a resume event calls `startLoad()`
- park events after `pagehide` teardown touch neither `stopLoad` nor `startLoad`

`mediaPreload.test.ts`:
- the `pause` IPC handler dispatches `streamwall:media-pause`
- the `resume` IPC handler dispatches `streamwall:media-resume`
- both are announced even when no media element has been acquired yet

Full gate green locally: `format:check`, `lint`, `typecheck`, `npm test`, `streamwall-control-client` build.

## Risks / breaking changes

None expected. The behaviour only activates behind the opt-in, default-off `--park.pause` flag, and only for the bundled HLS player page — remote stream views are unaffected beyond receiving two custom events they have no listeners for. No IPC surface changed.

Closes #384

### DIFF
--- a/packages/streamwall/src/preload/mediaParkEvents.ts
+++ b/packages/streamwall/src/preload/mediaParkEvents.ts
@@ -1,0 +1,14 @@
+// Same-document channel between mediaPreload.ts and the bundled HLS player
+// page (renderer/playHLS.ts), used to stop segment fetching for a parked view
+// rather than only pausing its <video> (issue #384).
+//
+// These must be dispatched on `document`, not `window`: under
+// contextIsolation the preload's isolated world has its own `window` object,
+// so a window-dispatched event never reaches the page's own listeners, while
+// DOM nodes -- `document` included -- are shared between the worlds.
+//
+// This file deliberately carries constants only. playHLS.ts is renderer
+// bundle code and must not pull in mediaPreload.ts's `electron` imports, so
+// the names cannot simply live there alongside the IPC handlers.
+export const MEDIA_PAUSE_EVENT = 'streamwall:media-pause'
+export const MEDIA_RESUME_EVENT = 'streamwall:media-resume'
--- a/packages/streamwall/src/preload/mediaPreload.test.ts
+++ b/packages/streamwall/src/preload/mediaPreload.test.ts
@@ -314,6 +314,58 @@ describe('mediaPreload pause/resume handling (issue #374)', () => {
     expect(video.paused).toBe(false)
   })
 
+  // The bundled HLS player page (renderer/playHLS.ts) keeps its hls.js
+  // instance in a closure the preload cannot reach, so pausing the <video>
+  // alone leaves segment fetching to taper off on its own. These events are
+  // the same-document channel that lets it call stopLoad()/startLoad()
+  // instead (issue #384).
+  function parkEventNames(): string[] {
+    const names: string[] = []
+    const record = (event: Event) => names.push(event.type)
+    document.addEventListener('streamwall:media-pause', record)
+    document.addEventListener('streamwall:media-resume', record)
+    return names
+  }
+
+  it('announces a pause to the page world so an HLS player can stop loading segments', async () => {
+    await loadAcquiredVideo()
+    const events = parkEventNames()
+
+    registeredHandler('pause')()
+
+    expect(events).toEqual(['streamwall:media-pause'])
+  })
+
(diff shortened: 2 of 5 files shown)


NilsR0711:
Follow-up issue created for an out-of-scope finding noticed while setting up this branch: #429 (`package-lock.json` drifts on every `npm install`). Unrelated to this change; the lockfile churn was deliberately kept out of this PR's commit.

Issue #381, opened 2026-07-15
Add in-app update notification with download & install


NilsR0711:
## Problem

Streamwall has no update mechanism. Users on an old version have no way of knowing a new release exists unless they manually check the GitHub Releases page. There is no in-app notification, no download progress, and no one-click install.

## Proposal

Add auto-update support to the Electron app (`packages/streamwall`):

- Check GitHub Releases for a newer version on app start (and optionally on an interval).
- Show a non-intrusive in-app notification/banner when an update is available (version number, release notes link).
- Let the user trigger the download from the notification, with progress feedback.
- Once downloaded, prompt to install & restart (or install automatically on quit).
- Respect the existing release flow (`v*` tag + electron-forge, GitHub Releases as the distribution channel — see [self-hosting.md](docs/self-hosting.md) / README for release process).

## Suggested approach

- Use [`electron-updater`](https://www.electron.build/auto-update) or Electron Forge's built-in [`@electron-forge/publisher-github`](https://www.electronforge.io/config/publishers/github) + `update-electron-app` for the update check/download/install flow, since releases are already published to GitHub Releases.
- Surface update state (checking / available / downloading / ready) via IPC to the renderer so the control UI (or a small system-level notification) can render it.
- Should work across macOS/Windows/Linux build targets currently produced by `forge.config.ts`.

## Out of scope

- Code signing/notarization (tracked separately, required for silent auto-install on macOS/Windows — see prior signing issue).
- Auto-update for `streamwall-control-server` (server-side, not relevant here).

## Acceptance criteria

- [ ] App detects when a newer GitHub release is available.
- [ ] User sees an in-app notification with the new version.
- [ ] User can download the update from the notification with visible progress.
- [ ] User can install & restart from the notification once downloaded.
- [ ] Works without requiring the user to leave the app or visit GitHub manually.

PR #430, opened 2026-07-20
feat(control-server): report version and notify about new releases

### LINKED ISSUE #382: Add self-update mechanism for streamwall-control-server deployments
## Problem

[#381](https://github.com/NilsR0711/streamwall/issues/381) adds in-app update notifications for the Electron desktop app, but explicitly excludes `streamwall-control-server`. Self-hosted operators running the control server (see [self-hosting.md](docs/self-hosting.md), Docker + Caddy deployment) currently have no way to know a newer server version/image is available, and no built-in way to update short of manually pulling a new image/tag and redeploying.

## Proposal

Give self-hosters a way to know when a newer `streamwall-control-server` release exists, and a documented (not necessarily automatic) path to update:

- Surface the running server version and whether a newer one is available — e.g. an admin-only status endpoint or log line checked against GitHub Releases/container registry tags.
- Document the update procedure for the Docker Compose deployment (pull new image tag, recreate container, migration/backup notes if state format changes).
- Decide whether "update" means fully automatic (pull + restart) or just a checked notification the operator acts on manually — given this runs unattended on a server, an automatic in-place update is higher risk than the desktop app case and should probably default to notify-only.

## Out of scope

- Auto-updating the Electron desktop app (tracked in [#381](https://github.com/NilsR0711/streamwall/issues/381)).
- Zero-downtime / rolling updates (not a current deployment goal per [self-hosting.md](docs/self-hosting.md)).

## Acceptance criteria

- [ ] Operator has a way to see the running control-server version.
- [ ] Operator is notified (log, status endpoint, or similar) when a newer release exists.
- [ ] Update procedure for the Docker deployment is documented.

NilsR0711:
## Problem

Self-hosted `streamwall-control-server` operators had no way to see which
version they were running, and no way to learn that a newer release exists
short of watching the repo. #381 covers this for the Electron app but
explicitly excludes the server.

## Solution

Notify-only update awareness, plus a documented update path.

- **Version at startup** — `Starting streamwall-control-server 0.9.1` on
  every boot.
- **Daily release check** — queries the GitHub Releases API once a day and
  logs a single line per newly discovered version (not once per check).
- **`GET /admin/status`** — admin-only JSON with `version`, `latestVersion`,
  `updateAvailable`, `releaseUrl`, `lastCheckedAt`, `checkEnabled`. Every
  other role and anonymous requests get `403`, gated through the existing
  `roleCan` default-deny matrix via a new `view-server-status` admin action.
- **`STREAMWALL_UPDATE_CHECK=false`** disables the check (the server's only
  outbound connection) for egress-free deployments; the endpoint still
  reports the running version.
- **Docs** — `docs/self-hosting.md` gains a version/update-notification
  section, and the update procedure now covers backing up the
  `control-server-data` volume and what to expect during the restart.

## Architecture decisions

- **Notify-only, not self-updating.** An unattended rebuild-and-restart
  would drop live uplink and client sockets at an arbitrary moment, and the
  Compose update is a single command anyway. The issue reaches the same
  conclusion.
- **Admin-gated status.** The version of a publicly reachable server is
  exactly the detail that helps someone shop for a known vulnerability, so
  it stays behind auth rather than sitting on an open `/status`.
- **Fail-soft check.** Offline host, rate limit, malformed body and hung
  connection all degrade to "no update known" and keep the last known
  result; an unparsable version never yields a bogus "update available".
  The interval timer is `unref`'d so a pending check cannot delay shutdown.
- **Version alignment.** The check compares against `vX.Y.Z` release tags,
  which electron-forge derives from the `streamwall` package version. The
  control-server manifest was on an unrelated `1.0.0`; it is now aligned to
  the release line (`0.9.1`) and pinned there by a test, so a release bump
  cannot silently desync the two.

## Testing

- 34 new tests: version comparison (incl. prerelease precedence and
  unparsable input), env parsing, every `fetchLatestRelease` failure mode,
  checker state/logging/scheduling, and the full `/admin/status`
  authorization matrix.
- Verified end-to-end over real HTTP against a listening server: admin
  `200` + `cache-control: no-store`, operator `403`, anonymous `403`.
- Verified the checker against the live GitHub API: correctly reports
  running `0.9.1` == latest `0.9.1`, `updateAvailable: false`.
- Full quality gate green locally: `format:check`, `lint`, `typecheck`,
  `npm test` (all workspaces).

## Risks / breaking changes

None breaking. New behaviour is one outbound HTTPS request per day, opt-out
via `STREAMWALL_UPDATE_CHECK=false` and documented as such.

Closes #382

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -44,6 +44,12 @@ STREAMWALL_TRUST_PROXY=true
 #STREAMWALL_WS_MSG_RATE=100
 #STREAMWALL_WS_MSG_BURST=1000
 
+# Optional: the daily check against the GitHub Releases API that logs when a
+# newer control-server release exists (see docs/self-hosting.md). It only ever
+# notifies -- the server never updates itself. Set to false for a deployment
+# that must make no outbound connections at all.
+#STREAMWALL_UPDATE_CHECK=false
+
 # Optional: crash reporting to your own Sentry project (off by default).
 #STREAMWALL_SENTRY_ENABLED=true
 #STREAMWALL_SENTRY_DSN=
--- a/docs/self-hosting.md
+++ b/docs/self-hosting.md
@@ -121,6 +121,54 @@ for the bundled stack.
   trust proxy if the process is reachable directly from the internet without
   a trusted reverse proxy in front.
 
+## Version and update notifications
+
+The server prints its version on every start:
+
+```sh
+docker compose logs control-server | grep 'Starting streamwall-control-server'
+```
+
+```
+Starting streamwall-control-server 0.9.1
+```
+
+Once a day it asks the GitHub Releases API whether a newer release exists,
+and logs a single line per newly discovered version:
+
+```
+⬆️  streamwall-control-server 1.0.0 is available (running 0.9.1): https://github.com/NilsR0711/streamwall/releases/tag/v1.0.0
+```
+
+The same information is available to a signed-in **admin** at
+`GET /admin/status` (any other role, or no session, gets a `403`):
+
+```json
+{
+  "version": "0.9.1",
+  "latestVersion": "1.0.0",
+  "updateAvailable": true,
+  "releaseUrl": "https://github.com/NilsR0711/streamwall/releases/tag/v1.0.0",
+  "lastCheckedAt": "2026-07-20T09:00:00.000Z",
+  "checkEnabled": true
+}
+```
+
+This is **notify-only** — the server never updates itself. Applying an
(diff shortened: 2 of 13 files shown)


NilsR0711:
Follow-ups filed for out-of-scope findings noticed while implementing this:

- #435 — automate release version bumps across workspace manifests (this PR adds a second manifest that must track the release line, currently backstopped only by a test).
- #436 — surface the version and update notice in the control UI (this PR exposes it via log line and `GET /admin/status` only).

PR #437, opened 2026-07-20
test(control-e2e): cover security-critical browser flows

### LINKED ISSUE #391: Test: E2E suite missing security-critical browser flows
## Summary

The Playwright E2E suite (`streamwall-control-e2e`, 9 tests) covers happy-path multiplayer smoke scenarios but **does not exercise failure modes that are security- and UX-critical in production** — even though the control-server unit tests cover the server side.

## Current E2E coverage

✅ Admin invite → grid render  
✅ Unauthorized banner  
✅ Grid cell edit → uplink  
✅ Role gating (monitor/operator)  
✅ WebSocket reconnect (#37/#283)  
✅ Invite creation, grid resize command  

## Gaps (server tested, browser not)

| Scenario | Server unit test | E2E |
|----------|------------------|-----|
| HTTP rate limit → user-visible feedback | `rateLimit.test.ts` | ❌ |
| Invalid/expired invite link in browser | `inviteExchange.test.ts` | ❌ |
| Two simultaneous browser clients editing grid | `multiplexing.test.ts` | ❌ |
| Operator blocked from admin-only UI actions | partial role gating | ❌ |
| Custom stream add/delete roundtrip | — | ❌ |
| Token revocation in Access Panel | — | ❌ |

## Impact

- Server/client contract regressions can pass unit tests but break the actual operator experience.
- Reconnect tests (#283) proved E2E value — the same pattern applies to error surfaces.
- Multi-client Yjs conflicts are invisible until a real event with two operators.

## Proposed specs

1. **`rate-limit.spec.ts`** — burst requests → connection banner shows rate-limit reason (not silent drop)
2. **`invalid-invite.spec.ts`** — visit `/invite/:id#token=bad` → error message, no session cookie
3. **`multi-client.spec.ts`** — two browser contexts → concurrent grid edits converge
4. **`custom-stream.spec.ts`** — add/remove custom stream → uplink harness sees update

## Acceptance criteria

- [ ] At least rate-limit + invalid-invite E2E specs added
- [ ] Multi-client spec or documented deferral with rationale
- [ ] Specs run in CI `e2e` job without flaking on port races

NilsR0711:
## Problem

The Playwright E2E suite covered happy-path multiplayer smoke scenarios but did not exercise several failure modes that are security- and UX-critical in production, even though the equivalent server-side behavior was already unit-tested:

| Scenario | Server unit test | E2E (before) |
|----------|------------------|----|
| HTTP/WS rate limit → user-visible feedback | `rateLimiter.test.ts` | ❌ |
| Invalid/expired invite link in browser | `inviteExchange.test.ts` | ❌ |
| Two simultaneous browser clients editing grid | `multiplexing.test.ts` | ❌ |
| Custom stream add/delete roundtrip | — | ❌ |

## Solution

Added four new Playwright specs to `streamwall-control-e2e`, following the existing harness/fixture conventions:

- **`rate-limit.spec.ts`** — drives a client past the per-connection inbound WebSocket message budget (`STREAMWALL_WS_MSG_BURST`/`_RATE`, set to a tiny value for this test only) and asserts the rate-limited banner appears, then that the client cleanly reconnects and resyncs (same recovery path as the existing reconnect regression test).
- **`invalid-invite.spec.ts`** — visits an invite link with an unknown id/secret and asserts the exchange page's error copy renders and no session cookie is ever set.
- **`multi-client.spec.ts`** — opens two separate browser contexts (two independent sessions), has both edit different grid cells concurrently, and asserts both edits converge on both clients.
- **`custom-stream.spec.ts`** — drives the sidebar's add/delete custom-stream forms and asserts the resulting `update-custom-stream`/`delete-custom-stream` commands reach the Streamwall peer with the expected shape.

`harness.ts` gained a small `pushState` helper so a test can simulate the Streamwall peer re-broadcasting a state snapshot after processing a forwarded command (used to seed an existing custom stream before the delete test) without touching the shared default state every other spec relies on.

## Architecture decisions

- `rate-limit.spec.ts` starts its own harness via the exported `startHarness()` (rather than the shared `harness` fixture) so it can set the message-budget env vars *before* the server boots. The budget applies per-connection, so the fake uplink's own two startup sends (state + doc snapshot) get an independent bucket from the browser client under test — capacity 2 covers both cleanly while still being small enough to trip deliberately with a 3rd edit.
- `multi-client.spec.ts` uses two separate `BrowserContext`s (not two pages in one context) so each gets its own session cookie, matching how two real operators would actually connect.
- Kept scope to the issue's four proposed specs. Token revocation in the Access Panel is a real gap too (flagged in the issue's gap table but not one of its proposed specs) — filed as a separate follow-up issue rather than expanding scope here.

## Testing performed

- `npm -w streamwall-control-e2e run test:e2e` — all 15 specs pass (10 pre-existing + 5 new).
- `npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` (all workspaces) — all clean.
- `npm run build` (streamwall-control-client, the only workspace with a build step) — succeeds.

## Risks / breaking changes

None — test-only change, no production code touched.

Closes #391

### DIFF
--- a/packages/streamwall-control-e2e/tests/custom-stream.spec.ts
+++ b/packages/streamwall-control-e2e/tests/custom-stream.spec.ts
@@ -1,0 +1,70 @@
+import type { StreamData } from 'streamwall-shared'
+import { expect, test } from './harness.ts'
+
+/**
+ * End-to-end coverage for issue #391: the Sidebar's "Custom Streams" panel
+ * (`Sidebar.tsx`'s `CustomStreamInput`/`CreateCustomStreamInput`) has no
+ * server-side unit test at all, unlike the Access panel's invite round-trip
+ * (`invite-management.spec.ts`). These drive the add and delete forms in a
+ * real browser and assert the resulting `update-custom-stream` /
+ * `delete-custom-stream` commands actually reach the Streamwall peer.
+ */
+
+const SEEDED_CUSTOM_STREAM: StreamData = {
+  _id: 'https://example.com/seeded-custom-stream',
+  _dataSource: 'custom',
+  kind: 'video',
+  link: 'https://example.com/seeded-custom-stream',
+  label: 'Seeded Custom Stream',
+}
+
+test('adding a custom stream from the sidebar forwards an update-custom-stream command to the peer', async ({
+  page,
+  harness,
+}) => {
+  await page.goto(await harness.createInviteLink())
+  await expect(
+    page.getByRole('heading', { name: 'Custom Streams' }),
+  ).toBeVisible()
+
+  const link = 'https://example.com/new-custom-stream'
+  const addForm = page
+    .locator('form')
+    .filter({ has: page.getByRole('button', { name: 'add stream' }) })
+  await addForm.getByPlaceholder('https://...').fill(link)
+  await addForm.getByPlaceholder('Label (optional)').fill('Field Cam')
+
+  const addCommandPromise = harness.waitForCommand('update-custom-stream')
+  await addForm.getByRole('button', { name: 'add stream' }).click()
+
+  const addCommand = await addCommandPromise
+  expect(addCommand).toMatchObject({
+    type: 'update-custom-stream',
+    url: link,
+    data: { link, kind: 'video', label: 'Field Cam' },
+  })
+})
+
+test('deleting an existing custom stream forwards a delete-custom-stream command to the peer', async ({
+  page,
+  harness,
+}) => {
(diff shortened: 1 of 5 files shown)


NilsR0711:
Scoped to the four specs #391 explicitly proposed (rate-limit, invalid-invite, multi-client, custom-stream). Token revocation was flagged in the issue's gap table but wasn't one of the proposed specs and has zero coverage today (not even a server unit test) - filed as a separate follow-up: #438

PR #441, opened 2026-07-20
fix(streamwall): main-process tests depend on electron being hoisted to the root node_modules

### LINKED ISSUE #439: fix(streamwall): main-process tests depend on electron being hoisted to the root node_modules
## Problem

`electron-log` is a dependency of `packages/streamwall`, but npm hoists it to the root `node_modules`. Its entry point resolves Electron with a bare `require('electron')`:

```
Error: Cannot find module 'electron'
Require stack:
- node_modules/electron-log/src/main/index.js
- node_modules/electron-log/main.js
```

That resolution only succeeds because `electron` also happens to be hoisted to the root today. The dependency tree layout is not something we control, and npm changes it on a version bump.

## How it surfaced

In [#422](https://github.com/NilsR0711/streamwall/pull/422), bumping `electron` 43.1.0 → 43.1.1 made npm place the package at `packages/streamwall/node_modules/electron` instead of the root. The hoisted `electron-log` could then no longer resolve it, and **15 main-process test files failed to load** on all three platforms (ubuntu, macos, windows), e.g. `StreamWindow.test.ts`, `data.test.ts`, `viewStateMachine.test.ts`, `commandDispatch.test.ts`.

Notably this reproduces only on a clean CI install — a local macOS tree happened to hoist differently and stayed green, so the local test run is not a reliable signal for this failure mode.

The electron bump was deferred out of #422 to unblock it; the underlying fragility is unfixed.

## Task

Make the main-process tests independent of the hoisting layout. Options worth weighing:

- Alias or mock `electron` in `packages/streamwall/vitest.config.ts` so `electron-log` resolves it under test (the config already aliases `react` → `preact/compat` and inlines CJS deps, so there is an established pattern).
- Stub `electron-log` in tests rather than letting the real module load.

## Acceptance criteria

- [ ] Main-process tests pass regardless of whether `electron` is hoisted to the root or nested under `packages/streamwall`
- [ ] Bumping `electron` to 43.1.1 (or newer) keeps CI green on all three platforms
- [ ] The deferred electron bump from #422 is applied

NilsR0711:
## Problem

`electron-log` is a dependency of `packages/streamwall`, but npm hoists it to the root `node_modules`. Its entry point resolves Electron with a bare `require('electron')`, which only succeeds because `electron` also happens to be hoisted to the root today. The dependency tree layout is not something we control, and npm changes it on a version bump — in #422, bumping `electron` 43.1.0 → 43.1.1 made npm place `electron` at `packages/streamwall/node_modules/electron` instead of the root, and the hoisted `electron-log` could no longer resolve it, failing 15 main-process test files to load on all three CI platforms. The electron bump was deferred out of #422 to unblock it.

## Technical solution

`packages/streamwall/src/main/logger.ts` imports `electron-log/main`. That entry point is CJS-only (no ESM/browser build), and does a bare `require('electron')` at module scope. Under Vitest, electron-log is loaded as an opaque external CJS module, so that require runs through real Node resolution rooted at electron-log's own install location — not this package's — which is why it depends on npm's hoisting choice.

I initially tried the same fix `vitest.config.ts` already uses for `react-icons`/`styled-components` (alias `electron` to a resolved path + `deps.inline` to route electron-log through Vite's resolver). That's the wrong fix here and CI caught it: unlike `react-icons`/`styled-components`, electron-log ships no ESM/browser build, so inlining a pure-CJS module doesn't rewrite its internal `require()` calls — they still go through raw Node resolution regardless of the alias, and CI still failed with the exact same `Cannot find module 'electron'` error.

The actual fix: electron-log also ships `electron-log/node`, a Node-only build with the same Logger API (`transports`, log levels, `initialize()`) that never touches `electron` anywhere in its dependency chain (verified: no `require('electron')` in its source tree at all). Main-process code under test never runs inside a real Electron process anyway, so `vitest.config.ts` now aliases `electron-log/main` → `electron-log/node` for tests, sidestepping the hoisting-dependent require entirely instead of trying to make it resolve reliably.

The deferred `electron` 43.1.0 → 43.1.1 bump from #422 is applied in a separate commit.

## Testing performed

- Full quality gate (`format:check`, `lint`, `typecheck`, `npm test` across all workspaces, `streamwall-control-client` build, `streamwall` Electron package smoke test) passes locally.
- As the issue notes, this failure mode is hoisting-layout-dependent and not reliably reproducible on a local tree. I got a genuine, non-contrived reproduction anyway: a plain `npm install` in this environment happened to place `electron` under `packages/streamwall/node_modules` instead of the root (the same "nested" layout #422 hit on CI) — I confirmed the *previous* (incorrect) alias/inline fix still failed with `Cannot find module 'electron'` under that real layout, and confirmed the current fix passes the full suite (50 files / 525 tests) under that same layout.
- No new test file was added: `logger.test.ts` already imports `electron-log/main` (indirectly, via `./logger`) unmocked, so it already serves as the regression guard for this resolution path.

## Risks / breaking changes

None expected. This is test-tooling configuration only, plus a routine patch-level `electron` bump (bugfix release, no breaking changes per Electron's release notes). `electron-log/node` omits only the Electron IPC-forwarding behavior (irrelevant outside a real Electron process) and uses OS/package.json-derived paths instead of `electron.app.getPath()` for its default log directory — neither is exercised by any test (file logging is disabled in tests via `log.transports.file.level = false`).

Closes #439

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "packages": {
     "": {
       "name": "streamwall",
+      "license": "MIT",
       "workspaces": [
         "packages/streamwall",
         "packages/streamwall-shared",
@@ -7371,25 +7372,6 @@
         "node": ">=0.12.18"
       }
     },
-    "node_modules/electron": {
-      "version": "43.1.0",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-43.1.0.tgz",
-      "integrity": "sha512-DPfxpQLd4NL3BJ8DBxYAfmLUKKesF5Rx9dQx5FyczAP8bhOPScjHE48GArVeXu68LlAainuwkmQTQvdZwpIIAQ==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@electron-internal/extract-zip": "^1.0.1",
-        "@electron/get": "^5.0.0",
-        "@types/node": "^24.9.0"
-      },
-      "bin": {
-        "electron": "cli.js",
-        "install-electron": "install.js"
-      },
-      "engines": {
-        "node": ">= 22.12.0"
-      }
-    },
     "node_modules/electron-installer-common": {
       "version": "0.10.3",
       "resolved": "https://registry.npmjs.org/electron-installer-common/-/electron-installer-common-0.10.3.tgz",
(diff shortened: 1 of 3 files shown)


NilsR0711:
Follow-up filed for an unrelated finding noticed while debugging this: #446 (TwitchBot.test.ts's `afterEach` masks the real error when `beforeEach` throws).

PR #443, opened 2026-07-20
feat(streamwall): notify Linux users of available updates via GitHub Releases

### LINKED ISSUE #433: Linux builds have no update path at all
## Problem

The in-app update notification added in #381 covers macOS and Windows only. Electron's built-in `autoUpdater` and `update-electron-app` are both no-ops on Linux, so a Linux user never learns that a new version exists — the situation #381 set out to fix, still unaddressed for one of the three shipped targets.

`forge.config.ts` produces `.deb` and `.rpm` packages, so Linux is a supported target, not an afterthought.

## Impact

Linux users stay on stale versions indefinitely unless they manually watch the GitHub Releases page. They also miss security fixes, with no signal from the app.

## Suggested solution

Two options, roughly in order of effort:

1. **Notify only.** Poll the GitHub Releases API from the main process, compare against `app.getVersion()`, and reuse the existing `UpdateStatus` / banner plumbing to show a "new version available" banner whose action opens the release page instead of installing. Low effort, and it closes the awareness gap — which is the bulk of the value.
2. **Full auto-update.** Ship an AppImage target and use `electron-updater`, which does support AppImage self-updates. Larger change, overlaps with #432, and adds a distribution format the project deliberately deferred.

Option 1 is the pragmatic first step: `.deb`/`.rpm` users generally expect their package manager to handle installs, so notification plus a link matches platform conventions better than self-replacing binaries do.

NilsR0711:
## Problem

The in-app update notification added in #381/#434 covers macOS and Windows only. Electron's built-in `autoUpdater` (Squirrel) and `update-electron-app` are both no-ops on Linux, so a Linux user never learns a new version exists — the same awareness gap #381 fixed for the other two platforms.

## Solution

Implements option 1 from the issue (notify only):

- **`LinuxUpdateChecker`** (`packages/streamwall/src/main/linuxUpdateChecker.ts`) polls the GitHub Releases API once a day, compares the latest published (non-draft, non-prerelease) tag against `app.getVersion()`, and reports a new `available` `UpdateStatus` state. Every fetch failure (offline, rate limit, malformed body, timeout) leaves the previous status untouched instead of resetting to idle, so a transient network hiccup does not hide an update the user has already seen. The version-compare/fetch shape mirrors `streamwall-control-server/src/updateCheck.ts`.
- **`UpdateBanner`** gets a new `available` branch: names the new version, and offers a **View Release** button (opens the GitHub release page) with no install action, since `.deb`/`.rpm` installs go through the OS package manager rather than a self-updater.
- **`main/index.ts`** branches on `process.platform === 'linux'`: Linux gets `LinuxUpdateChecker` wired the same way `AppUpdater` already is on macOS/Windows (status pushed to the control window, `openReleaseNotes` gated in main so the renderer cannot spoof the target URL). `install` is a no-op there since it's never reachable from the banner.
- `useUpdateStatus`'s per-version dismissal now also covers `available`, matching the existing `ready` handling.
- README's "In-app updates" section documents the new Linux behavior.

## Architecture decisions

- **New `available` state instead of reusing `ready`**: `ready` means "downloaded and installable"; Linux never downloads or installs anything, so overloading `ready` would have been misleading and would have required threading a platform flag through the banner.
- **Duplicated version-compare/fetch logic rather than sharing it with `streamwall-control-server`**: the two packages already have separate, independently-tested copies of this shape (control-server's is Node-only, doesn't share `streamwall-shared`, and has no `repository` field to derive from). Consolidating them is a real but separate cleanup — filed as a follow-up issue (see below) rather than folded into this change.
- **No new IPC channel**: the existing `openReleaseNotes` channel (gated in main, no URL param from the renderer) already does exactly what the Linux "View Release" action needs.

## Testing

- `packages/streamwall/src/main/linuxUpdateChecker.test.ts` (new): version comparison, GitHub API URL construction, available/idle transitions, draft/prerelease skipping, failure resilience (keeps last known status), start/stop interval scheduling.
- `packages/streamwall/src/renderer/UpdateBanner.test.tsx`: new `available` branch coverage (version display, no install button, View Release click, dismiss).
- Full quality gate run locally: `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm test` (all packages, 0 failures), plus `npm -w streamwall-control-client run build` and `npm -w streamwall run package` (Electron Forge packaging smoke test) — all green.

## Risks / breaking changes

None expected. macOS/Windows update flow is untouched (same `AppUpdater`/`update-electron-app` path, unchanged behavior). The Linux path is new and additive; `install` is a no-op there so a stray IPC call cannot do anything.

Closes #433

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,10 +290,12 @@ byte-level progress, the banner shows an indeterminate indicator rather than a
 percentage. Update-check failures are logged, not surfaced — they are usually
 just an offline machine.
 
-This only runs in packaged builds, and only on macOS and Windows; on Linux the
-updater is a no-op and users install from the released packages. Note that
-macOS additionally requires the app to be **signed** (see below) before updates
-can install.
+This only runs in packaged builds, and only on macOS and Windows: Squirrel is
+a no-op on Linux. Linux builds instead poll the GitHub Releases API directly
+once a day and show a **View Release** banner when a newer version is found —
+notification only, since `.deb`/`.rpm` installs go through the OS package
+manager rather than a self-updater. Note that macOS additionally requires the
+app to be **signed** (see below) before updates can install.
 
 To produce signed, notarized builds, set these environment variables before
 running `make`/`publish` (see `packages/streamwall/forge.signing.ts`):
--- a/packages/streamwall/src/main/index.ts
+++ b/packages/streamwall/src/main/index.ts
@@ -66,6 +66,7 @@ import {
   applyLayoutPreset,
   buildLayoutPreset,
 } from './layoutPresets'
+import { LinuxUpdateChecker } from './linuxUpdateChecker'
 import log, {
   LOG_LEVELS,
   initLogger,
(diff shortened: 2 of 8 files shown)


NilsR0711:
Filed the semver/GitHub-release-fetch duplication (control-server vs. this PR's linuxUpdateChecker.ts) as a follow-up: #445

PR #462, opened 2026-07-21
fix(streamwall): log swallowed promise rejections on hot paths

### LINKED ISSUE #392: Bug: Swallowed promise rejections hide stream load and data-source failures
## Summary

Several hot paths in the Electron main process and preload **catch promise rejections with empty handlers** (`() => {}`), discarding load failures and data-pipeline errors without logging. Operators see stale tiles or silent data gaps with no log breadcrumb.

## Evidence

```ts
// packages/streamwall/src/main/data.ts
eventP.catch(() => {})
nextP.catch(() => {})

// packages/streamwall/src/main/viewStateMachine.ts
wc.loadURL(content.url).catch(() => {})

// packages/streamwall/src/preload/mediaPreload.ts
currentMedia?.play().catch(() => {})
```

## Impact

- **Stream tiles:** `loadURL` rejection (network error, blocked navigation) produces no log line; the view may sit in a loading state until timeout.
- **Data sources:** watcher/generator failures in `combineDataSources` are swallowed; playlist/custom file sources can stop updating silently.
- **Media preload:** play() rejection (autoplay policy) is invisible during debugging.

Note: `viewStateMachine` also listens for `did-fail-load`, but the empty `.catch()` still hides the rejection reason from Promise-based diagnostics.

## Proposed fix

Minimum: replace empty catches with `log.warn('…', err)` using `./logger`.

Better: centralize in a tiny `logRejected(promise, context)` helper to keep catches one line without going silent.

Add tests where feasible (e.g. `data.test.ts` already documents generator teardown — extend for error logging).

## Acceptance criteria

- [ ] No empty `.catch(() => {})` on production paths in `data.ts`, `viewStateMachine.ts`, `mediaPreload.ts` (test-only catches excepted)
- [ ] Failures produce at least a warning in electron-log with context (view id, URL, source name)
- [ ] Regression test for at least one path (e.g. data source error logged)

NilsR0711:
## Problem

Several hot paths in the Electron main process and preload caught promise rejections with empty handlers (`() => {}`), discarding load and data-pipeline failures without logging. Operators saw stale tiles or silent data gaps with no log breadcrumb (#392).

## Solution

Replaced the swallowing catches with logging ones, using each file's established logging convention:

| Site | Change |
|------|--------|
| `viewStateMachine.ts` `loadPage` | `loadURL` rejection now logged via `log.warn('error loading view URL', url, err)`. The `did-fail-load` listener never sees the Promise rejection reason, so this is the only breadcrumb for a blocked navigation / network error. Not awaited (unchanged), so no risk of stranding `VIEW_INIT`. |
| `data.ts` `markDataSource` | The rejecting source is now logged **by name** (`log.warn('error advancing data source', name, err)`) before it propagates through `combineDataSources`. This is the race-loser guard, so it fires whether the source loses or wins the race; **error propagation is unchanged** — the awaited race still throws. |
| `mediaPreload.ts` resume handler | `play()` rejection (e.g. autoplay policy) now logged via `console.warn`, matching this file's existing `console.warn`/`console.log` usage (preload runs in a renderer context; `electron-log/main` is not available there). |

### Deliberate exception — `watchDataFile`

The issue cited `eventP.catch(() => {})` in `watchDataFile`, but that rejection is **already** surfaced: the `once(watcher, 'all')` promise only rejects on a watcher `'error'`, which is logged by the persistent `watcher.on('error', ...)` listener a few lines above (and, when the promise wins the race, by the `try/catch` around the race). Re-logging in the guard would duplicate — or triple — an already-emitted breadcrumb. Rather than add a redundant log, the empty catch is kept as a pure unhandled-rejection guard with a comment explaining why. This still satisfies the issue's goal (no undiagnosable swallow).

## Testing

TDD (red → green) for each behavior change:

- `viewStateMachine.test.ts` — drives the real `loadPage` actor with a rejecting `loadURL` and asserts the warning (message + URL + error).
- `data.test.ts` `markDataSource` — a source whose `next()` rejects logs the warning naming the source **and** still rejects to the consumer (propagation preserved).
- `mediaPreload.test.ts` — a rejecting `play()` on resume logs via `console.warn`.

Full quality gate green locally: `format:check`, `lint`, `typecheck`, full test suite (all packages).

## Risk / breaking changes

None. Purely additive logging on failure paths; control flow and error propagation are unchanged.

Closes #392

### DIFF
--- a/packages/streamwall/src/main/data.test.ts
+++ b/packages/streamwall/src/main/data.test.ts
@@ -15,6 +15,7 @@ import {
   StreamIDGenerator,
   watchDataFile,
 } from './data'
+import log from './logger'
 import type { PresetPack } from './presets'
 
 class FakeWatcher extends EventEmitter {
@@ -663,6 +664,30 @@ describe('markDataSource', () => {
     } finally {
       await marked.return?.(undefined)
     }
+  })
+
+  test('logs a warning naming the source and propagates when it rejects', async () => {
+    const warnSpy = vi.spyOn(log, 'warn').mockImplementation(() => {})
+    const failure = new Error('source exploded')
+    const source = {
+      next: () => Promise.reject(failure),
+      return: async () => ({ done: true as const, value: undefined }),
+      [Symbol.asyncIterator]() {
+        return this
+      },
+    } as unknown as AsyncIterableIterator<StreamDataContent[]>
+
+    const marked = markDataSource(source, 'boom')
+    // The rejection must still surface to the consumer, not be silently
+    // swallowed by the logging catch.
+    await expect(marked.next()).rejects.toThrow('source exploded')
+    expect(warnSpy).toHaveBeenCalledWith(
+      'error advancing data source',
+      'boom',
+      failure,
+    )
+
+    warnSpy.mockRestore()
   })
 })
 
--- a/packages/streamwall/src/main/data.ts
+++ b/packages/streamwall/src/main/data.ts
@@ -111,6 +111,11 @@ export function watchDataFile(
         // Raced against `stop` so an early return() settles immediately
         // instead of queuing behind an event that may never fire.
         const eventP = once(watcher, 'all')
+        // Pure unhandled-rejection guard for the race loser. `once(watcher,
+        // 'all')` rejects on a watcher 'error', which is already logged by the
+        // persistent 'error' listener above (and, when this promise wins the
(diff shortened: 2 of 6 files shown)


NilsR0711:
Follow-up spotted while resolving this issue: #464 — `watchDataFile` logs a chokidar `'error'` twice (persistent listener + race `catch`). Out of scope here; tracked separately.

PR #479, opened 2026-07-21
fix(streamwall): report cross-origin iframe as a distinct media failure

### LINKED ISSUE #413: Tech debt: Iframe-based video extraction path is unverified (open TODO)
## Summary

When a stream page embeds video inside an `<iframe>`, `mediaPreload.ts` attempts to reach into `iframe.contentDocument` to find the `<video>` element. This path carries an explicit **TODO: verify iframe still works** and has no automated test coverage.

## Evidence

```ts
// packages/streamwall/src/preload/mediaPreload.ts ~269
if (iframe && iframe.contentDocument) {
  // TODO: verify iframe still works
```

## Problem

Cross-origin iframes will have `contentDocument === null` (opaque origin). Same-origin iframe embeds may work today but break with site changes. Failures manifest as empty tiles after the generic 10s timeout (#304 closed — related but not iframe-specific).

## Impact

- Site-specific streams using iframe players fail silently
- Security-sensitive code path (DOM access across frame boundary) unregression-tested
- Operators cannot distinguish "no video" vs. "iframe extraction failed"

## Proposed fix

1. Add fixture HTML pages in vitest/happy-dom tests for same-origin iframe video
2. Document cross-origin limitation in operator docs
3. Consider explicit error state: "Video inside cross-origin iframe — unsupported"

## Acceptance criteria

- [ ] Test covers same-origin iframe extraction OR explicit skip with comment
- [ ] TODO removed or converted to documented limitation
- [ ] Operator-facing error when iframe path cannot access video

NilsR0711:
## Problem

`mediaPreload.ts` reaches into `iframe.contentDocument` to find a `<video>` when a stream page embeds its player in an iframe. That path carried an explicit `TODO: verify iframe still works` and had no test coverage. Cross-origin frames have an opaque origin, so `contentDocument` is `null` and the loop skipped them silently — the tile then failed with the same generic `could not find video` as a slow or empty page, giving operators no way to tell an unsupported embed from a transient failure.

## Solution

- `waitForVideo()` now distinguishes "iframe present but unreachable" from "iframe reachable, no video", and returns the iframe document it already resolved.
- `findMedia()` throws a specific error for the cross-origin case: `could not find video: it may be inside a cross-origin iframe, which is unsupported`. It reaches the operator through the existing `view-error` channel.
- The redundant `iframe.contentDocument` re-check (and its stale TODO) is gone: the document is passed through from the scan.
- The iframe scan now accepts the requested `kind`, i.e. `<audio>` inside an iframe is matched too, instead of only `HTMLVideoElement`.
- README documents the same-origin-only limitation and the workaround (use the player page's own URL as the stream `link`).

## Testing

New `mediaPreload iframe video extraction (issue #413)` suite (vitest + happy-dom):

- same-origin iframe (`srcdoc` fixture): video is acquired, hoisted into the iframe document, override style injected, iframe and ancestors marked, `view-loaded` sent, no error — this verifies the path the TODO doubted.
- cross-origin iframe (`contentDocument === null`): the specific error is reported.
- reachable iframe without a video: the generic message is kept.

Full quality gate green locally: `npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` (no root build script exists).

## Risks

Low. The only behavior change on the success path is that an `<audio>` element inside a same-origin iframe is now matched for `kind: "audio"` (previously silently missed). The error text changes only for the cross-origin case.

Closes #413

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,14 @@ Each entry (a `[[streams]]` table in TOML, or an object in the JSON array) suppo
 - `background` — a webpage loaded behind the grid instead of in a tile.
 - `overlay` — a webpage loaded as a full-screen `<iframe>` layered over the whole wall (e.g. scoreboards, alerts). See [Security: overlay and background streams](#security-overlay-and-background-streams) below.
 
+For `video` and `audio` streams, Streamwall also looks inside `<iframe>` players
+on the page. That only works for **same-origin** iframes: a cross-origin player
+frame has an opaque origin, so its `<video>` tag is unreachable from the
+embedding page and no browser permission can change that. Such a tile fails with
+`could not find video: it may be inside a cross-origin iframe, which is
+unsupported` — use the player page's own URL (the one the iframe loads) as the
+stream `link` instead.
+
 ## Playlists
 
 A grid cell can optionally cycle through a fixed list of stream URLs on an
--- a/packages/streamwall/src/preload/mediaPreload.test.ts
+++ b/packages/streamwall/src/preload/mediaPreload.test.ts
@@ -244,6 +244,120 @@ describe("mediaPreload emptied handler's re-acquisition rejection", () => {
   })
 })
 
+describe('mediaPreload iframe video extraction (issue #413)', () => {
+  // Must match INITIAL_TIMEOUT in mediaPreload.ts; the iframe scan only runs
+  // after the top-level <video> wait loses its race against this timeout.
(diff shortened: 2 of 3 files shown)


NilsR0711:
Follow-up filed from this work: #485 (iframe player scan is one-shot; a late-inserted iframe is never re-examined).

PR #481, opened 2026-07-21
fix(streamwall): validate IPC-delivered StreamwallState in control renderer

### LINKED ISSUE #409: Enhancement: Validate IPC-delivered StreamwallState in control renderer with Zod
## Summary

The Electron control renderer accepts `StreamwallState` snapshots from the main process over IPC **without runtime validation** — while collab data via Yjs *is* validated through `useYDoc` + `collabDataSchema`.

## Evidence

```tsx
// packages/streamwall/src/renderer/control.tsx
// TODO: improve typing (Zod?)
function handleState(state: StreamwallState) {
  setStreamwallState(state)
}
```

Main process can emit partial or malformed state during bugs/upgrades. Unlike the control-server uplink path (#387), this stays local — but can still corrupt the control UI grid.

## Impact

- UI renders garbage views/streams if main-process state emitter regresses
- Inconsistent validation story: Yjs ✅, IPC state ❌
- Harder to catch schema drift between main and renderer

## Proposed fix

1. Add `streamwallStateSchema` in `streamwall-shared` (shared with #387 server validation)
2. `safeParse` in `handleState`; on failure log + keep last good state (or show banner)

## Acceptance criteria

- [ ] IPC state validated before `setStreamwallState`
- [ ] Invalid snapshots logged via renderer Sentry/log
- [ ] Tests with intentionally malformed payload
- [ ] Pairs with #387 (same schema, both boundaries)

NilsR0711:
## Problem

The Electron control renderer accepted `StreamwallState` snapshots from the main process over IPC **without runtime validation**, while collab data arriving via Yjs was already validated through `useYDoc` + `collabDataSchema`. A partial or malformed snapshot — pushed by a main-process regression or during a mid-upgrade schema drift — rendered straight into the control grid, and schema drift between main and renderer stayed invisible.

Closes #409.

## Solution

`createIpcCollabTransport` now gates every incoming snapshot through `streamwallStateSchema` from `streamwall-shared` — the very schema the control server already applies at its uplink boundary (#387), so both `StreamwallState` consumers enforce one contract.

Invalid snapshots are logged via `console.warn` (matching the uplink handler's convention) and dropped. Dropping rather than clearing means the shared `useCollabConnection` hook stays on the last valid snapshot, so the UI keeps rendering usable state instead of blanking, and recovers automatically once valid snapshots resume.

## Architecture decisions

- **Validated in the transport adapter, not in `useCollabConnection`.** The shared hook is transport-agnostic; the WebSocket client's snapshots already pass the control server's `streamwallStateSchema` check, so validating centrally would double-validate that path (including its `state-delta` patches, which would need separate treatment). Keeping the gate in the IPC adapter puts it exactly at the unguarded boundary this issue is about.
- **Gate, not transform.** The original object is forwarded to `events.onState` instead of `result.data`, so fields the schema does not model are not silently stripped on the way through.
- **No new schema.** `streamwallStateSchema` shipped with #387 and is reused unchanged.

## Testing

`packages/streamwall/src/renderer/ipcCollabTransport.test.ts` gained a `state validation` block: valid snapshots forwarded by reference with no warning; malformed payloads dropped and logged, covering a non-object, `null`, missing required keys, a malformed `views` array, a non-array `streams`, and an unknown `identity.role`; last-valid-snapshot retention when an invalid one arrives; and recovery once valid snapshots resume.

Full quality gate green locally: `npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` (all packages).

## Risks

Low. A false rejection would freeze the control UI on the last valid snapshot rather than crash it, and the schema is already proven against real desktop state on the uplink path, where the identical `clientState` object is validated today.

### DIFF
--- a/packages/streamwall/src/renderer/ipcCollabTransport.test.ts
+++ b/packages/streamwall/src/renderer/ipcCollabTransport.test.ts
@@ -1,7 +1,31 @@
 import { type CollabTransportEvents } from 'streamwall-control-ui'
-import { describe, expect, it, vi } from 'vitest'
+import { type StreamwallState } from 'streamwall-shared'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { type StreamwallControlGlobal } from '../preload/controlPreload'
 import { createIpcCollabTransport } from './ipcCollabTransport'
+
+/** A minimal snapshot that satisfies `streamwallStateSchema`. */
+const VALID_STATE = {
+  identity: { role: 'local' },
+  config: {
+    cols: 3,
+    rows: 3,
+    width: 1920,
+    height: 1080,
+    frameless: false,
+    fullscreen: false,
+    activeColor: '#fff',
+    backgroundColor: '#000',
+  },
+  streams: [],
+  customStreams: [],
+  views: [],
+  fullscreenViewIdx: null,
+  streamdelay: null,
+  layoutPresets: [],
+  favorites: [],
+  dataSourceHealth: [],
+} as unknown as StreamwallState
 
 function createMockControl() {
   const onStateUnsubscribe = vi.fn()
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up from this PR: #488 (validate the patched state on the WebSocket `state-delta` path in the control client — the last unvalidated `StreamwallState` boundary).

PR #491, opened 2026-07-21
feat(control-server): replace console logging with structured Fastify logs

### LINKED ISSUE #410: Enhancement: Replace console.log with structured logging in control-server
## Summary

The control-server uses raw **`console.log/warn/error`** throughout its entrypoint (~1030 lines), while the Electron app uses **`electron-log`** with file transport and levels. The server has optional Sentry (#298) but no structured log levels, correlation IDs, or redaction.

## Problem

Examples in `packages/streamwall-control-server/src/index.ts`:

```ts
console.log('Streamwall connecting from', request.ip, tokenInfo)
console.log('Streamwall connected from', request.ip, tokenInfo)
console.warn('Rejecting Streamwall connection (already connected) from', request.ip, tokenInfo)
```

Issues:

- No log level filtering in production
- No request/client correlation (`clientId`, WS session)
- Token metadata (role, name, tokenId) logged to stdout — operational leakage in shared hosting logs
- Inconsistent with Fastify's built-in `pino` logger (not used)

## Impact

- Production debugging requires grep through unstructured stdout
- Log aggregation (Docker, systemd, cloud) gets no severity field
- Multi-tenant/self-host operators may inadvertently log identifiable session metadata

## Proposed fix

- Use Fastify logger (`app.log.info/warn`) with JSON output in production
- Redact or omit `tokenInfo` details at info level; log only `tokenId` prefix at debug
- Add `LOG_LEVEL` env var (document in server README)

## Acceptance criteria

- [ ] No raw `console.log` in server request/WS hot paths (startup banner excepted)
- [ ] Structured JSON logs behind env flag or default in production
- [ ] README documents log level + redaction policy
- [ ] Existing tests pass (mock logger in inject tests)

NilsR0711:
## Summary

The control server logged through raw `console.log/warn/error`, so production output had no severity field, no correlation ids, and printed token metadata (role, name, tokenId) straight to stdout — identifiable session data once it lands in shared hosting or an aggregated log store.

Every request and WebSocket hot path now logs through Fastify's built-in pino logger (structured JSON on stdout), with the verbosity selected by `LOG_LEVEL`.

**Design**

- Handlers take a logger explicitly rather than importing a singleton: `request.log` for uplink handlers, a `request.log.child({ clientId, role })` per browser client. Every entry therefore carries Fastify's `reqId` plus the connection's own `clientId`, so a whole session can be filtered out of a log store.
- New `logger.ts` owns the policy: `resolveLogLevel` (unknown/blank values fall back to `info`, so a typo can never silence the server), `identityFields`/`identityDebugFields`, `tokenIdPrefix`, and a request serializer that redacts token ids embedded in paths.
- `Auth` takes an optional logger, so token creation is logged at `debug` with redacted fields instead of writing `{ tokenId, role, name }` to stdout. It stays optional so `Auth` remains usable standalone.

**Redaction policy** (documented in the server README):

- the operator-supplied token **name** is never logged, at any level
- **token ids** never appear at `info` and above — including inside request paths: the ids in `/streamwall/:id/ws` and `/invite/:id` are replaced with `[redacted]`
- at `debug`, a four-character `tokenIdPrefix` is logged instead: enough to correlate entries within one process, useless as a credential
- the `role` is kept, since authorization decisions are unreadable without it and a role is not identifying on its own

The startup banner stays on `console` on purpose (and is called out as such in code and README): it is the operator's only chance to copy the uplink secret and admin invite link, so it must stay visible whatever `LOG_LEVEL` is set to. `sentry.ts`'s misconfiguration warning likewise runs before the Fastify instance exists.

Closes #410

## How was this tested?

- New `logging.test.ts` boots a live server against a capturing log stream and asserts on real emitted entries: JSON shape and severity level, presence of `reqId`/`clientId`, that no info-level entry contains the uplink token id or token name, that the debug prefix is a strict truncation, that the uplink request path is logged redacted, and that a configured level actually filters lower entries out.
- New `logger.test.ts` covers `resolveLogLevel` (default, case-insensitivity, unknown value), `getLoggerOptions` reading `LOG_LEVEL`, `tokenIdPrefix`, and `redactRequestUrl` (including that unrelated paths are untouched).
- The existing specs that spied on `console.warn` (`multiplexing`, `wsValidation`, `sentryReporting`) now assert on captured structured entries via a new `captureLogs()` test helper. Test output is silent by default, since `buildTestApp` runs at level `silent` unless a capture is passed.
- `npm run lint`, `npx prettier --check .`, `npm run typecheck` and `npm test` (1470 tests across all workspaces) pass locally.

## Checklist

### DIFF
--- a/deploy/.env.example
+++ b/deploy/.env.example
@@ -50,6 +50,12 @@ STREAMWALL_TRUST_PROXY=true
 # that must make no outbound connections at all.
 #STREAMWALL_UPDATE_CHECK=false
 
+# Optional: log verbosity. One of fatal, error, warn, info, debug, trace,
+# silent. Defaults to info. Logs are structured JSON on stdout; see the
+# redaction policy in packages/streamwall-control-server/README.md before
+# raising this to debug on a shared host.
+#LOG_LEVEL=info
+
 # Optional: crash reporting to your own Sentry project (off by default).
 #STREAMWALL_SENTRY_ENABLED=true
 #STREAMWALL_SENTRY_DSN=
--- a/packages/streamwall-control-server/README.md
+++ b/packages/streamwall-control-server/README.md
@@ -90,6 +90,36 @@ host, rate limit, malformed response) is non-fatal and simply leaves the last
 known result in place. See
 [`docs/self-hosting.md`](../../docs/self-hosting.md) for the update procedure.
 
+### Logging
+
+The server logs through Fastify's built-in [pino](https://getpino.io) logger:
+one structured JSON object per line on stdout, each carrying a severity level
+and the request correlation id (`reqId`) of the connection it belongs to.
+Client WebSocket entries additionally carry the per-connection `clientId`, so a
+whole session can be filtered out of an aggregated log store.
+
+| Variable    | Default | Description                                                                                                 |
+| ----------- | ------- | ----------------------------------------------------------------------------------------------------------- |
+| `LOG_LEVEL` | `info`  | One of `fatal`, `error`, `warn`, `info`, `debug`, `trace`, `silent`. An unknown value falls back to `info`. |
+
+**Redaction policy.** Logs are assumed to end up somewhere less trusted than
+the server itself (shared hosting, a log aggregator, a support ticket), so
+identifiable session metadata never reaches them:
+
+- The operator-supplied token **name** is never logged, at any level.
+- **Token ids** are never logged at `info` and above — neither as a field nor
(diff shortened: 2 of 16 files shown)


NilsR0711:
Follow-up for the deliberately out-of-scope `console` call sites (updateCheck announcements, the Sentry misconfiguration warning, and the two startup diagnostics in bootstrap.ts): #493.

Issue #432, opened 2026-07-20
Show determinate download progress and gate the update download behind user consent


NilsR0711:
## Problem

The in-app update notification added in #381 sits on Electron's built-in `autoUpdater` (Squirrel), which has two hard limits:

- It emits **no download-progress event**, so the banner can only show an indeterminate indicator, never a percentage or byte count.
- It **starts downloading automatically** as soon as an update is found. There is no way to offer a "Download" button and let the user decide when to spend the bandwidth.

Both were listed in #381's acceptance criteria ("download the update from the notification with visible progress") and are currently only partially met: the user sees *that* a download is happening, not how far along it is, and never gets to trigger it.

## Impact

On a metered or slow connection, Streamwall downloads a full app bundle without asking and gives no indication of how long it will take. For a tool typically run during live streaming, an unannounced background download competing for bandwidth is a real cost.

## Suggested solution

Swap Squirrel for [`electron-updater`](https://www.electron.build/auto-update), which supports `autoDownload: false` and emits `download-progress` events with transferred/total bytes.

The catch: `electron-updater` expects electron-builder's `latest.yml` / `latest-mac.yml` metadata, which Electron Forge does not produce. So this needs either a Forge hook that generates that metadata at publish time, or a move off Forge's maker pipeline.

The existing seam should make the swap contained: `AppUpdater` in `packages/streamwall/src/main/appUpdater.ts` already takes its backend as a constructor argument and translates it into the shared `UpdateStatus` type, so the change is a new backend plus a `downloading` variant carrying progress, rather than a rewrite of the IPC and UI layers.

Blocked in practice on code signing for macOS (see the existing signing issue) — unsigned macOS builds cannot install updates regardless of which updater drives them.

PR #496, opened 2026-07-21
docs: document the manual Windows upgrade path off Squirrel installs (#452)

### LINKED ISSUE #452: Existing Squirrel.Windows installs have no migration path to the NSIS releases
## Problem

The switch from `@electron-forge/maker-squirrel` to an NSIS installer driven by `electron-updater` (#432) changes the Windows release artifacts: new releases carry `streamwall-setup-<version>.exe` + `latest.yml` instead of `RELEASES` + `.nupkg` + Squirrel setup exe.

Apps installed from the old Squirrel installers (v0.9.x) check for updates through Electron's built-in Squirrel `autoUpdater` against update.electronjs.org, which looks for Squirrel artifacts (`RELEASES`/`.nupkg`) on the latest GitHub release. Once the first NSIS-only release is published, those installs will simply find no update — they are stranded on their current version without any error or notice.

## Impact

Every existing Windows user stays on their old version silently unless they manually download the new installer. The user base is currently small (first release v0.9.0 was recent), so the window to handle this cheaply is now.

## Suggested solution

Options, roughly in order of effort:

1. **Document only**: release notes of the first NSIS release tell Windows users to download and run the new installer once (NSIS installs over the Squirrel install location? It does not — it installs to its own location; the old app should be uninstalled).
2. **One transitional release**: publish one release containing *both* Squirrel artifacts (whose payload is a build that just notifies 'please reinstall') and the NSIS installer.
3. Leave a pinned issue/README note.

At minimum, option 1 should happen with the next release. Worth deciding before tagging it.

NilsR0711:
## Problem

Windows installs from v0.9.1 or older were produced by `@electron-forge/maker-squirrel` and check for updates via Squirrel artifacts (`RELEASES`, `.nupkg`). Since the switch to an NSIS installer driven by electron-updater (#432), releases no longer carry those artifacts, so those installs find no update *and* report no error — they silently stay on their old version.

## Solution

Option 1 from the issue (document it), applied in the three places a user or a release author can hit it:

- **README → Download**: an `[!IMPORTANT]` block with the one-time migration steps (uninstall the old entry, run `streamwall-setup-<version>.exe`), why it is needed (NSIS installs to its own location rather than replacing the Squirrel install), and that `%APPDATA%\Streamwall` config/logs survive.
- **README → In-app updates**: names the stranded-install case explicitly so the update section is not silently wrong for those users.
- **CONTRIBUTING → Cutting a release**: a new "Release notes: the Windows upgrade notice" step with a verbatim markdown block to paste into every release's notes until v0.9.x is out of circulation.
- **CHANGELOG**: records the NSIS switch (#432) and the manual-migration requirement.

## Testing

Docs-only — no behavior to test, so no tests were added. `npm run format:check`, `npm run lint`, `npm run typecheck` and `npm test` are green locally.

## Risks

None: no code paths change. The residual risk is procedural — if the release-notes block is dropped from a future release, stranded users are not reached; that is why the step now lives in the release checklist rather than only in this PR.

Closes #452

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ changelog are kept in step.
 
 - Release version bumps are automated across the release-tracking workspace
   manifests via `npm run release:version` (#448).
+- Windows releases ship an NSIS installer instead of Squirrel.Windows (#432).
+  Installs from v0.9.1 or older cannot see newer releases and must be
+  reinstalled once by hand — README and the release-notes checklist in
+  CONTRIBUTING now spell out the steps (#452).
 
 ### Fixed
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -243,3 +243,26 @@ When cutting a release, alongside the `release:version` bump:
 is missing an `## [Unreleased]` section or a heading for the version currently
 in `packages/streamwall/package.json`, so a version bump without a matching
 changelog entry is caught before release.
+
+### Release notes: the Windows upgrade notice
+
+Releases after v0.9.1 ship an NSIS installer plus `latest.yml` instead of the
+Squirrel.Windows artifacts (`RELEASES`, `.nupkg`) that v0.9.x installs poll for
+(#432, #452). Those installs therefore find no update and report no error —
+they stay on their version until their owner reinstalls by hand, and only a
+release note tells them to.
+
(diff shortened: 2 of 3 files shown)


NilsR0711:
Follow-up filed while working on this: #503 (open high-severity Dependabot alert for transitive `fast-uri` in the control server tree). Unrelated to the docs change here.

PR #499, opened 2026-07-21
ci: enforce a production dependency license allowlist

### LINKED ISSUE #431: Add license-checker to release workflow for production dependency compliance
## Summary

Issue #414 normalized the `license` field across all workspace `package.json` files and documented the license policy in `CONTRIBUTING.md` (PR #427). One item from that issue's proposed fix was intentionally left out of scope:

> Optional: `license-checker` in release workflow for production deps

This tracks that follow-up.

## Current state

- `CONTRIBUTING.md` documents that new production dependencies must be license-compatible with MIT distribution (MIT/Apache-2.0/BSD/ISC fine, copyleft needs discussion), but compliance is manual/reviewer-enforced only.
- Release binaries (`.github/workflows/release.yml`, `electron-forge publish`) do not ship a consolidated `THIRD_PARTY_NOTICES` file — each dependency's license remains available only via its own `node_modules` entry.
- Nothing in CI automatically flags an incompatible (e.g. GPL/AGPL) license landing in `dependencies` via a new package or a transitive bump.

## Proposed fix

- Add a `license-checker` (or equivalent, e.g. `license-checker-rseidelsohn`) step to `.github/workflows/release.yml` and/or `ci.yml` that fails on disallowed SPDX licenses for production dependencies.
- Optionally generate a consolidated `THIRD_PARTY_NOTICES` file and attach it to GitHub releases alongside the packaged binaries.

## Impact

Low urgency — current dependencies are already compatible and the policy is documented for human reviewers. This is a defense-in-depth / automation improvement, not a known compliance violation.

## Acceptance criteria

- [ ] CI (or the release workflow) fails when a production dependency's license isn't on an allowlist
- [ ] Decision recorded on whether to ship a consolidated third-party notices file with releases

NilsR0711:
## Problem

`CONTRIBUTING.md` documented that new production dependencies must be
license-compatible with MIT redistribution, but nothing enforced it. An
incompatible license (GPL/AGPL) could land through a new package or a
transitive bump and only a reviewer's attention would catch it. Release
binaries also had no recorded decision on third-party notices.

## Solution

- `scripts/check-licenses.mjs` reads the installed production tree via
  `npm ls --omit=dev --all --long --json`, normalizes the SPDX string,
  legacy `{ type }` object and legacy `licenses` array shapes, and fails on
  anything outside an allowlist of permissive licenses.
- Exposed as `npm run licenses:check`; runs as a dedicated job in `ci.yml`
  (wired into the `CI OK` gate) and in `release.yml` ahead of `publish`.
- `CONTRIBUTING.md` documents the allowlist and the notices decision.

### Architecture decisions

- **No new dependency.** `npm ls --long` already reports each installed
  package's license and path, so a `license-checker` devDependency would
  only add supply-chain surface for parsing we need anyway.
- **Fail closed.** A missing license field, `SEE LICENSE IN …`, and a `WITH`
  exception are all violations. An exception rewrites the terms of the
  license it attaches to, so it needs a human decision, not an implicit pass
  from the base identifier.
- **Expression semantics.** `(A OR B)` passes when one branch is allowed;
  `A AND B` needs both. Malformed expressions are rejected, not thrown on.
- **Skipped nodes.** `npm ls` lists unmet optional/peer dependencies as empty
  nodes and can report leftovers as extraneous; neither is installed into a
  packaged release, so both are ignored.
- **No `THIRD_PARTY_NOTICES` file.** electron-forge packages each production
  dependency together with its own license file, which is what the MIT/BSD/ISC
  notice clauses require. A generated summary would be a second source of
  truth to keep current. Recorded in `CONTRIBUTING.md`, to be revisited if a
  dependency ever demands a separately placed notice.
- **Trigger on `code`, not just `dependencies`.** The `dependencies` paths
  filter only covers manifests; the allowlist itself lives in a script, and
  `code` already covers both.

## Testing

- `test/licenses.test.mjs` (node:test, runs in `npm test`): allowlist
  matching, SPDX expression evaluation including `WITH` and malformed input,
  tree flattening/dedupe, skipping of uninstalled and extraneous nodes,
  legacy license shapes, violation reporting, plus drift guards that the
  workflows wire the check in and that `CONTRIBUTING.md` documents exactly
  the allowlist constant.
- `npm run licenses:check` against the real tree: all 233 production
  dependencies pass. Verified the failure path end to end by temporarily
  removing `MIT` from the allowlist — 184 violations reported, exit code 1.
- `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm test`:
  all green locally.

## Risks

Low. The check is additive and does not touch runtime code. Worst case is a
false positive on an unusual license expression, which fails a build rather
than shipping something wrong; the allowlist is a one-line edit plus the
documented rationale.

Closes #431

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,6 +210,24 @@ jobs:
           docker logs control-server
           exit 1
 
+  # Production dependencies are redistributed inside the release binaries, so
+  # their licenses have to stay compatible with shipping an MIT-licensed
+  # bundle. A transitive bump can introduce a new license without touching a
+  # first-party dependency, which is why this runs on every code change rather
+  # than only on manifest edits. Policy and allowlist: CONTRIBUTING.md.
+  licenses:
+    name: Dependency licenses
+    needs: changes
+    if: needs.changes.outputs.code == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/setup
+      - run: npm run licenses:check
+
   # Reviews the dependency diff of a PR for known vulnerabilities.
   dependency-review:
     name: Dependency review
@@ -273,6 +291,7 @@ jobs:
         e2e,
         docker,
         dependency-review,
+        licenses,
         codeql,
         actionlint,
       ]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,8 +17,9 @@ concurrency:
   cancel-in-progress: false
 
 # The quality gate mirrors the PR `ci-ok` inputs (format, cross-OS tests,
-# control-client build, Electron package smoke, E2E) so a tag can never ship
-# code that would have failed a PR. Unlike CI there are no docs-only path
+# control-client build, Electron package smoke, E2E, dependency licenses) so
+# a tag can never ship code that would have failed a PR. Unlike CI there are
+# no docs-only path
 # filters: a release always builds and ships binaries, so every check runs
 # unconditionally. `publish` fans out only after all of them pass.
 jobs:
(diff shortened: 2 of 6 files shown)


NilsR0711:
Follow-up filed while working on this: #510 (local `npm -w streamwall run package` fails on macOS with an ad-hoc codesign error — unrelated to this change, reproduced on a clean `main` checkout).

PR #458, opened 2026-07-21
fix(control-ui): announce command and data-source banners to assistive tech

NilsR0711:
## Problem

`ConnectionStatusBanner` exposes connection state to assistive technology via `role="status"`, but `CommandErrorBanner` and `DataSourceHealthBanner` did not. Command failures (unauthorized, rate-limited, invalid payload) and data-source outages rendered visually only, so screen-reader users got no announcement when an action they triggered failed or a configured data file went missing/stale.

## Solution

- **`CommandErrorBanner`** → `role="alert"` + `aria-live="assertive"`. A command failure is an urgent, user-triggered outcome, so it interrupts the AT queue.
- **`DataSourceHealthBanner`** → `role="status"` + `aria-live="polite"`. An outage is recoverable and not directly user-triggered, so it is announced without interrupting.

Both follow the established pattern already used by `ConnectionStatusBanner` in the same UI.

## Architecture decisions

- Matched the assertive/polite split to urgency (WCAG-recommended): user-triggered failure = assertive, ambient health change = polite.
- Kept the live-region role on the existing banner wrapper (consistent with `ConnectionStatusBanner`, which returns `null` while healthy and mounts the region on state change). No structural/markup changes, so no visual regression.

## Testing

- Added tests to `CommandErrorBanner.test.tsx` and `DataSourceHealthBanner.test.tsx` asserting the `role`/`aria-live` attributes, mirroring the `ConnectionStatusBanner` accessibility tests.
- Full `streamwall-control-ui` suite: 280 passed. Root quality gate green: format, lint, typecheck, full test suite.

## Risks / breaking changes

None. Purely additive ARIA attributes; no visual or behavioral change for sighted users.

## Acceptance criteria

- [x] `CommandErrorBanner` announces errors to AT
- [x] `DataSourceHealthBanner` announces health changes
- [x] Tests assert ARIA attributes
- [x] No visual regression

Closes #398

Issue #398, opened 2026-07-15
A11y: Command and data-source error banners lack live-region semantics


NilsR0711:
## Summary

`ConnectionStatusBanner` correctly exposes connection state to assistive technology via `role="status"`, but **`CommandErrorBanner` and `DataSourceHealthBanner` do not** — so command failures and data-source outages are invisible to screen readers unless the user manually finds the banner.

## Problem

When a command fails (unauthorized, rate-limited, invalid payload), `CommandErrorBanner` renders visually but without:

- `role="alert"` or `aria-live="assertive"` for urgent failures
- `role="status"` / `aria-live="polite"` for recoverable warnings

Same gap for `DataSourceHealthBanner` when a configured data file is missing or stale.

## Impact

- WCAG 4.1.3 Status Messages: status changes should be programmatically determinable.
- Operators using screen readers get no announcement when an action they triggered failed.
- Inconsistent with the better pattern already used elsewhere in the same UI.

## Proposed fix

```tsx
<StyledCommandErrorBanner role="alert" aria-live="assertive">
```

For health warnings, prefer `role="status"` + `aria-live="polite"`.

Add tests mirroring `connectionStatus.test.tsx` patterns.

## Acceptance criteria

- [ ] `CommandErrorBanner` announces errors to AT
- [ ] `DataSourceHealthBanner` announces health changes
- [ ] Tests assert ARIA attributes
- [ ] No visual regression

PR #500, opened 2026-07-21
fix(control-ui): keep banner live regions mounted while healthy

### LINKED ISSUE #463: A11y: banner live regions mount on error, so the first status change may not be announced
## Context

Follow-up from #398 (PR #458), which added `role`/`aria-live` semantics to `CommandErrorBanner` and `DataSourceHealthBanner`, matching the existing `ConnectionStatusBanner`.

## Problem

All three banners (`ConnectionStatusBanner`, `CommandErrorBanner`, `DataSourceHealthBanner`) return `null` in their healthy/no-error state and only mount the live-region element when a problem occurs.

With `aria-live`, the region is expected to already exist in the accessibility tree **before** its content changes. When the whole region node is inserted at the same moment its text appears, several screen reader / browser combinations do **not** reliably announce that first change — the announcement is only guaranteed for subsequent updates to an already-present region.

This means the very first command failure or data-source outage after a clean state can be silently missed by AT, which partially undermines the guarantee #398 set out to provide.

## Impact

- WCAG 4.1.3 Status Messages: the first status change is the most important one to announce and is exactly the one at risk.
- Affects the whole banner family in `streamwall-control-ui`, including the pre-existing `ConnectionStatusBanner`.

## Suggested solution

Keep a persistent, always-mounted live-region container per banner (empty while healthy) and swap only its inner content on state change, instead of unmounting the region entirely. Alternatively, hoist a single shared visually-styled `aria-live` region that the banners write into.

Verify with an actual screen reader (VoiceOver / NVDA) that a first-occurrence error is announced, since this behavior is not observable in the happy-dom unit tests.

NilsR0711:
## Summary

`ConnectionStatusBanner`, `DataSourceHealthBanner` and `CommandErrorBanner` all returned `null` in their healthy state, so their `aria-live` region was inserted into the DOM at the same moment its first message appeared. Screen readers only reliably announce changes *inside a region that already exists* in the accessibility tree, so the first command failure, disconnect, or data-source outage after a clean state could be dropped silently — exactly the announcement WCAG 4.1.3 cares about most. This is the follow-up to #398 / #458, which added the `role`/`aria-live` semantics themselves.

Each banner now always renders its live-region container and swaps only the inner content:

- `ConnectionStatusBanner` — persistent `role="status" aria-live="polite"` wrapper; the message (with `data-testid="connection-status-banner"` and the `unauthorized`/`rate-limited` colour class) is the conditional child. Moving the test id onto the message keeps the existing E2E assertion "banner is gone after reconnect" (`toHaveCount(0)`) meaningful.
- `DataSourceHealthBanner` — persistent `role="status" aria-live="polite"` wrapper; failing sources render as children, healthy state renders none.
- `CommandErrorBanner` — persistent `role="alert" aria-live="assertive"` wrapper; icon, message and dismiss button render only while an error is set.

### Architecture decisions

- **Per-banner persistent region** rather than a single hoisted shared region (the issue's alternative): the banners already own distinct politeness levels (`assertive` for a user-triggered command failure, `polite` for recoverable outages) and distinct visual styling, and a shared region would have to re-derive both.
- **No `:empty { display: none }`** — that would take the region out of the accessibility tree again and reintroduce the exact bug. An empty flex container is zero-height and the enclosing `Stack` sets no `gap`, so there is no visual or layout change.
- No new abstraction was introduced for three small components; the shared rationale lives in each component's doc comment.

## How was this tested?

- Red/green TDD: new unit tests per banner assert (a) the live region is mounted while healthy, with the right `aria-live` value, and (b) the *same* region element is reused across the healthy → failing transition (i.e. it is not remounted). All six failed before the change.
- Existing assertions that the healthy state renders nothing were tightened to "renders no message" (empty region, no test id) instead of "no region", in `ConnectionStatusBanner.test.tsx`, `CommandErrorBanner.test.tsx`, `DataSourceHealthBanner.test.tsx`, and `disconnectRendering.test.tsx`.
- `npm run lint`, `npx prettier --check .`, `npm run typecheck`, `npm test` all pass locally (777 + 321 + 320 + … tests green).
- Not verifiable in happy-dom: the actual screen reader announcement. As the issue notes, a manual VoiceOver/NVDA pass on a first-occurrence error is still worthwhile; the structural guarantee (region present before content changes) is what the tests lock in.

## Risks / breaking changes

None functional. The only externally visible change is that an empty `[role="status"]` / `.command-error-banner` element is now always present in the DOM — the `data-testid` used by E2E still appears only while disconnected.

Closes #463

## Checklist

### DIFF
--- a/packages/streamwall-control-ui/src/CommandErrorBanner.test.tsx
+++ b/packages/streamwall-control-ui/src/CommandErrorBanner.test.tsx
@@ -29,9 +29,35 @@ function renderBanner(
 }
 
 describe('CommandErrorBanner', () => {
-  test('renders nothing when there is no error', () => {
+  test('renders no message when there is no error', () => {
     const el = renderBanner(null)
-    expect(el.querySelector('.command-error-banner')).toBeNull()
+    expect(el.textContent).toBe('')
+    expect(el.querySelector('.command-error-banner button')).toBeNull()
+  })
+
+  // `aria-live` only guarantees an announcement for content that changes
+  // inside an already-present region, so the region stays mounted (empty)
+  // while there is no error - otherwise the first command failure can go
+  // unannounced (WCAG 4.1.3, issue #463).
+  test('keeps the live region mounted while there is no error', () => {
+    const el = renderBanner(null)
+    const region = el.querySelector('.command-error-banner')
+    expect(region).not.toBeNull()
+    expect(region?.getAttribute('role')).toBe('alert')
+    expect(region?.getAttribute('aria-live')).toBe('assertive')
+  })
+
+  test('reuses the same live region element when an error appears', () => {
+    const el = renderBanner(null)
+    const regionWhileHealthy = el.querySelector('.command-error-banner')
+    act(() => {
+      render(
+        <CommandErrorBanner error="unauthorized" onDismiss={() => {}} />,
+        container!,
+      )
+    })
+    expect(el.querySelector('.command-error-banner')).toBe(regionWhileHealthy)
+    expect(regionWhileHealthy?.textContent).toContain('unauthorized')
   })
 
   test('shows the error message when set', () => {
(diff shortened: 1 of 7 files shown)


NilsR0711:
Follow-up filed while working on this: #502 (ServerUpdateBanner is the one banner in the stack with no live-region semantics at all — out of scope here, same fix pattern applies).

Issue #397, opened 2026-07-15
Tech debt: Replace fragile view-index RPCs with stable view IDs


NilsR0711:
## Summary

Grid control operations (reload, blur, volume, fullscreen, drag-move) identify targets by **numeric view index** (`viewIdx` / `idx`). Indices shift when the grid is resized or views are remapped — a long-standing TODO in `GridControls.tsx`.

## Evidence

```tsx
// packages/streamwall-control-ui/src/GridControls.tsx:92
// TODO: Refactor callbacks to use streamID instead of idx.
```

The same index-based addressing flows through:

- `ControlCommand` payloads (`streamwall-shared`)
- Yjs collab state (`views` array order)
- Main-process `onCommand` handlers
- IPC and WebSocket transports

## Impact

- Race: operator clicks reload during grid resize → wrong tile affected.
- Harder to implement persistent favorites/layout presets keyed by stable identity.
- API confusion: `streamId` exists on streams but grid ops use array index.

## Proposed direction

1. Add stable `viewId` (UUID or stream-derived key) to each view slot in `StreamwallState`.
2. Migrate commands to accept `viewId` (keep `viewIdx` as deprecated alias for one release).
3. Update `GridControls` callbacks and XState view machine lookup.

## Acceptance criteria

- [ ] Grid commands address views by stable ID
- [ ] Grid resize/remap does not invalidate in-flight operations
- [ ] Schema migration documented for saved state / Yjs docs
- [ ] TODO removed from `GridControls.tsx`

PR #501, opened 2026-07-21
fix(control-ui): send stable view ids from audio and blur hotkeys

### LINKED ISSUE #470: Audio-listen and blur hotkeys send a grid cell index where a stable view id is expected
## Summary

The alt+`<key>` audio-listen and alt+shift+`<key>` blur hotkeys still address views by **grid cell index**, but since #397/#467 the underlying commands expect a **stable view id**. The hotkeys therefore send the wrong identifier.

## Details

In `ControlUI` (now `packages/streamwall-control-ui/src/hooks/useControlHotkeys.ts` after #393) the toggles resolve state by cell index and then pass that same index on as the command argument:

```ts
const toggleListening = useCallback(
  (idx: number) => {
    const isListening = stateIdxMap.get(idx)?.isListening ?? false
    handleSetListening(idx, !isListening)   // <- idx is a cell index
  },
  [stateIdxMap, handleSetListening],
)
```

but the command layer now takes a view id:

```ts
const handleSetListening = useCallback(
  (viewId: number, listening: boolean) => {
    send({ type: "set-listening-view", viewId: listening ? viewId : null })
  },
  [send],
)
```

`toggleBlurred` / `handleSetBlurred` have the same shape.

The two are different addressing axes:

- `stateIdxMap` is keyed by **grid cell index** (0..cols*rows-1).
- `viewId` is `view.webContents.id`, assigned by Electron (see `StreamWindow.ts`: `const viewId = view.webContents.id`). It is monotonically increasing and unrelated to cell position.

By contrast the grid overlay was migrated correctly and reads the id off the actor context:

```tsx
const { id: viewId, pos } = state.context
<GridControls viewId={viewId} ... />
```

## Impact

Pressing the audio-listen or blur hotkey targets a view whose id happens to equal the cell index, i.e. usually **no view at all**, or — worse — an unrelated view. Toggling audio/blur by keyboard is a core operator workflow, so this is a silent functional regression. The mouse/UI path is unaffected.

## Suggested fix

Resolve the view id from the `ViewInfo` the hotkey already looks up, rather than reusing the cell index:

```ts
const toggleListening = useCallback(
  (idx: number) => {
    const view = stateIdxMap.get(idx)
    if (!view) return
    handleSetListening(view.state.context.id, !view.isListening)
  },
  [stateIdxMap, handleSetListening],
)
```

(and the same for `toggleBlurred`). Consider exposing `id` on `ViewInfo` in `streamwallState.tsx` so callers do not have to reach through `state.context`.

Worth adding a regression test that asserts the dispatched `viewId` differs from the cell index when they diverge — the current hotkey tests only assert that a command is sent.

## Notes

Found while refactoring `ControlUI` in #393 / PR #468. That PR preserves this behavior verbatim (it is a pure structural refactor), so the bug is carried over unchanged and is not introduced by it.

NilsR0711:
## Problem

The `alt+<key>` audio-listen and `alt+shift+<key>` blur hotkeys resolved the targeted view through `stateIdxMap`, which is keyed by **grid cell index**, and then passed that same index on to `handleSetListening` / `handleSetBlurred`. Since #397/#467 those commands take a **stable view id** (`view.webContents.id`, assigned by Electron and unrelated to cell position).

Pressing an audio or blur hotkey therefore addressed a view whose id happened to equal the cell index — usually no view at all, occasionally an unrelated one. The mouse/UI path was unaffected, so this was a silent keyboard-only regression in a core operator workflow.

## Solution

`useControlHotkeys` now resolves the view id from the `ViewInfo` it already looks up:

```ts
const view = stateIdxMap.get(idx)
if (!view) return
handleSetListening(view.state.context.id, !view.isListening)
```

Same for `toggleBlurred`. A hotkey on an empty cell is now a no-op instead of dispatching a command for a non-existent view (previously it sent `viewId: idx` with a `false` default state). The handler parameter names in the hook's prop types were renamed `idx` → `viewId` to match the command layer.

## Architecture notes

The two addressing axes stay separate on purpose: the hotkey layers keep mapping keys to grid cells (that is what the operator sees), and the translation to a view id happens at the boundary to the command layer — the same place the grid overlay does it via `state.context.id`. No change to `ViewInfo`'s shape was needed, so nothing else in the state pipeline moves.

## Testing

New `packages/streamwall-control-ui/src/hooks/useControlHotkeys.test.tsx` renders the hook with a mocked `react-hotkeys-hook`, captures the registered layer handlers and invokes them directly. Fixtures deliberately keep cell index and view id apart (cell 1 → id 43, cell 20 → id 77, …) so a regression cannot pass by coincidence. Covers: both audio layers, both blur layers, toggling off from an already-active state, and the empty-cell no-op. All six failed before the fix.

Full quality gate green locally: `format`, `lint`, `typecheck`, and the complete test suite (1594 tests across all packages).

## Risks

None known. Behavior change is limited to the two hotkey toggles; the empty-cell keypress going from "send a bogus command" to "do nothing" is the intended part of the fix.

Closes #470

### DIFF
--- a/packages/streamwall-control-ui/src/hooks/useControlHotkeys.test.tsx
+++ b/packages/streamwall-control-ui/src/hooks/useControlHotkeys.test.tsx
@@ -1,0 +1,178 @@
+import { render } from 'preact'
+import { act } from 'preact/test-utils'
+import type { ViewState } from 'streamwall-shared'
+import { afterEach, describe, expect, test, vi } from 'vitest'
+import { hotkeyTriggers } from '../hotkeyLabel.ts'
+import type { ViewInfo } from '../streamwallState.tsx'
+import { useControlHotkeys } from './useControlHotkeys.ts'
+
+const useHotkeysMock = vi.hoisted(() => vi.fn())
+vi.mock('react-hotkeys-hook', () => ({
+  useHotkeys: useHotkeysMock,
+}))
+
+let container: HTMLDivElement | undefined
+
+afterEach(() => {
+  if (container) {
+    act(() => render(null, container!))
+    container.remove()
+    container = undefined
+  }
+  useHotkeysMock.mockClear()
+})
+
+function makeViewInfo(
+  id: number,
+  spaces: number[],
+  overrides: Partial<ViewInfo> = {},
+): ViewInfo {
+  const state: ViewState = {
+    state: {
+      displaying: {
+        running: {
+          playback: 'playing',
+          video: 'normal',
+          audio: 'muted',
+          pause: 'unpaused',
+          swap: 'idle',
+        },
+      },
+    },
+    context: {
+      id,
+      content: null,
+      info: null,
+      pos: { x: 0, y: 0, width: 1, height: 1, spaces },
+      error: null,
+      volume: 1,
+    },
+  }
+  return {
+    state,
+    isListening: false,
+    isBackgroundListening: false,
+    isBlurred: false,
+    volume: 1,
+    spaces,
+    ...overrides,
+  }
+}
+
+function renderHotkeys(stateIdxMap: Map<number, ViewInfo>) {
+  const handleSetListening = vi.fn<(viewId: number, on: boolean) => void>()
+  const handleSetBlurred = vi.fn<(viewId: number, on: boolean) => void>()
+  function Probe() {
+    useControlHotkeys({
+      stateIdxMap,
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up from this PR: #507 — type `stateIdxMap` at construction and separate the `viewId` / cell-index axes at the type level, so this class of mix-up cannot recur.

PR #504, opened 2026-07-21
feat(control-ui): give the sidebar stream handle a keyboard path

### LINKED ISSUE #460: A11y: Sidebar stream drag handle has no keyboard path
## Summary

The stream id/drag handle in `Sidebar.tsx` (`StyledId`) is a `<div onMouseDown=…>`. It is not focusable and offers no keyboard-operable way to place a stream on the wall — pointer users only.

## Context

Split out from #399. That issue added an interim `aria-label` to the handle so screen readers can at least identify it, but explicitly deferred the keyboard interaction:

> Drag handle: see separate keyboard-path issue; at minimum `aria-label` if kept as mouse-only interim

## Impact

- WCAG 2.1.1 Keyboard: the "add stream to wall" action is unreachable without a pointer.
- Keyboard and switch-device users cannot populate the grid from the sidebar.

## Proposed direction

- Make the handle a real focusable control (e.g. a `<button>`), or add `role="button"` + `tabIndex={0}` with `keydown` (Enter/Space) handling.
- Preserve the existing `mousedown` behaviour that keeps a destination grid input focused (see the comment in `StreamLine`).
- Assert keyboard activation in `Sidebar.test.tsx`.

NilsR0711:
## Problem

The stream id/drag handle in the sidebar (`StyledId` in `Sidebar.tsx`) was a plain `<div onMouseDown=…>`. It was not focusable and offered no keyboard-operable way to place a stream on the wall, so the "add stream to wall" action was reachable with a pointer only — a WCAG 2.1.1 (Keyboard) failure for keyboard and switch-device users.

## Solution

- `StyledId` is now a real `<button type="button">` instead of a `div`.
- `onKeyDown` activates on `Enter`/`Space` and calls `preventDefault`, so the browser's synthetic click cannot activate the handle a second time.
- `onMouseDown` keeps the existing behaviour but now calls `preventDefault`, so pressing the handle does not move focus away from a destination grid input. `useGridCommands.handleClickId` resolves the target cell from `focusedInputIdx`, so that focus must survive the press.
- Disabled rows render a `disabled` button, which also removes the handle from the tab order (previously the handler was simply omitted).
- The badge keeps its previous layout by carrying `overflow: hidden` / `text-overflow: ellipsis` / `white-space: nowrap` explicitly — as a `div` it inherited those from the row's `& > div` rule.

## Architecture decisions

Native `<button>` over `role="button"` + `tabIndex={0}`: it gets focusability, the correct role, and disabled semantics for free. The mouse path stays on `mousedown` (not `click`) because the target-cell resolution depends on the grid input still being focused at press time.

## Testing

- New `StreamList keyboard path` suite in `Sidebar.test.tsx`: the handle is a focusable, non-disabled `<button type="button">`; `Enter` and `Space` invoke `onClickId` and prevent the default; other keys are ignored; disabled rows ignore keydown and render `disabled`; `mousedown` prevents the default focus shift; a full press-and-release activates exactly once.
- Existing selectors that assumed a `div` handle (`StreamList.test.tsx`) or the star being the only `<button>` in a row (`Sidebar.test.tsx` favorites) were narrowed.
- `npm run format` / `lint` / `typecheck` / `test` all green (322 tests).

## Risks

Visual only: as a `div`, the badge's own `color: #0a0d12` was overridden by the row's more specific `& > div { color: var(--text) }` rule. As a `button` the badge now renders with its intended dark-on-light text. Playwright E2E selectors are unaffected — the `add stream` submit-button lookups are scoped to the custom-stream form.

Closes #460

### DIFF
--- a/packages/streamwall-control-ui/src/Sidebar.test.tsx
+++ b/packages/streamwall-control-ui/src/Sidebar.test.tsx
@@ -147,6 +147,124 @@ describe('StreamList', () => {
   })
 })
 
+describe('StreamList keyboard path', () => {
+  function renderHandle({
+    disabled = false,
+    onClickId = () => {},
+  }: {
+    disabled?: boolean
+    onClickId?: (id: string) => void
+  }): HTMLButtonElement {
+    container = document.createElement('div')
+    document.body.appendChild(container)
+    act(() => {
+      render(
+        <StreamList
+          rows={[baseRow({ _id: 'xyz' })]}
+          disabled={disabled}
+          onClickId={onClickId}
+          favorites={new Set()}
+        />,
+        container!,
+      )
+    })
+    return container.querySelector<HTMLButtonElement>(
+      '[aria-label="Add stream xyz to the wall"]',
+    )!
+  }
+
+  test('renders the id/drag handle as a focusable button', () => {
+    const handle = renderHandle({})
+
+    expect(handle.tagName).toBe('BUTTON')
+    expect(handle.type).toBe('button')
+    expect(handle.disabled).toBe(false)
+    expect(handle.getAttribute('tabindex')).not.toBe('-1')
+  })
+
+  test('marks the handle disabled so it drops out of the tab order', () => {
+    const handle = renderHandle({ disabled: true })
+
+    expect(handle.disabled).toBe(true)
+  })
+
+  for (const key of ['Enter', ' ']) {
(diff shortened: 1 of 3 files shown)


NilsR0711:
Follow-up from this change: #508 (no consistent `:focus-visible` style for buttons in the control UI).

Issue #425, opened 2026-07-20
CI: maker-specific packaging (Squirrel, deb, rpm, zip) is still only exercised at release time


NilsR0711:
## Context

[#423](https://github.com/NilsR0711/streamwall/pull/423) (closes #400) added an `electron-forge package` smoke build to PR CI and to the release quality gate. That covers the failure modes shared by every platform: Vite/Forge misconfiguration, asar packing, native module unpacking (`bufferutil`, `utf-8-validate`), missing packaged assets.

It deliberately stops short of `electron-forge make`.

## Gap

The makers configured in `packages/streamwall/forge.config.ts` — `MakerSquirrel`, `MakerZIP` (darwin), `MakerRpm`, `MakerDeb` — are still exercised for the first time during `electron-forge publish` on a `v*` tag. A maker-specific failure therefore:

- surfaces only after a tag has been pushed, i.e. once the version is already public,
- fails on the affected OS leg of the publish matrix while the others may succeed, leaving a partially-populated GitHub release.

Realistic examples: a Squirrel `setupIcon`/`certificateFile` path issue on Windows, missing `fpm`/`rpmbuild` tooling on the Ubuntu runner for the deb/rpm makers, or a maker config field rejected after a Forge major bump.

## Impact

Medium. It does not affect users of merged code, but it makes releases unreliable in a way that is awkward to recover from — a failed leg means either a re-tag or a manually uploaded asset.

## Suggested solution

Options, roughly in increasing cost:

1. Run `npm -w streamwall run make` in the release quality gate instead of `package`, still Ubuntu-only. Covers the deb/rpm makers before publish; leaves Squirrel and the darwin zip uncovered.
2. Add a scheduled (e.g. weekly) or manually dispatchable workflow running `make` across the full `ubuntu`/`windows`/`macos` matrix, without publishing. Full coverage, cost decoupled from PR CI.
3. Run `make` on the full matrix in the release workflow before `publish`. Most thorough, but roughly doubles release wall-clock time.

Option 2 looks like the best value: it catches maker regressions (including those introduced by Dependabot's Forge bumps) well before a release, without slowing down PRs or releases. Worth checking first whether the Ubuntu runner image ships `fpm` and `rpmbuild`, since the deb/rpm makers need them.

## Acceptance criteria

- [ ] Maker-specific packaging is exercised on all three platforms outside of a release
- [ ] A maker regression is detectable without pushing a `v*` tag
- [ ] Failures produce actionable Forge/maker logs

Issue #406, opened 2026-07-15
Dependencies: dank-twitch-irc is deprecated with no security updates


NilsR0711:
## Summary

The Twitch chat bot integration depends on **`dank-twitch-irc@4.3.0`**, which npm marks as **deprecated** by its maintainer with an explicit statement that it will receive **no dependency or security updates**.

## Evidence

From `package-lock.json`:

> *"I (randers) am no longer supporting or updating dank-twitch-irc… This package will also not receive any dependency or security updates. For this reason I recommend you to choose a different library…"*

Used in `packages/streamwall/src/main/TwitchBot.ts`.

## Impact

- Twitch IRC/API changes can break the bot silently.
- Transitive vulnerabilities in `dank-twitch-irc` deps won't be patched upstream.
- Feature is user-facing (chat commands → grid control) — failures reflect on Streamwall.

## Options

1. **Migrate** to `@twurple/chat` (actively maintained Twitch library)
2. **Fork** `dank-twitch-irc` and maintain internally (higher cost)
3. **Mark experimental** in README/config docs and document the deprecation risk
4. **Remove** if usage is negligible — only if product decision allows

## Acceptance criteria

- [ ] Decision documented in issue/ADR
- [ ] Either migration PR or README warning + tracking label
- [ ] Tests updated for new library or explicit skip with rationale

PR #509, opened 2026-07-21
ci: fail the weekly build on deprecated direct dependencies

### LINKED ISSUE #494: CI: fail the build when a direct dependency is deprecated on npm
## Problem

`dank-twitch-irc` had been marked **deprecated** on npm — with an explicit
maintainer note that it would receive no dependency or security updates — for a
long time before anyone noticed (#406). Nothing in CI surfaces this: `npm ci`
prints deprecation notices, but they scroll past in the install log and no job
fails or annotates them. Dependabot only opens PRs for version bumps and
advisories, not for a package being abandoned.

## Proposal

Add a small CI step that inspects the workspace's *direct* dependencies and
fails (or at minimum annotates the run) when any of them carries a
`deprecated` field on the registry:

```
npm ls --json --depth 0 --workspaces --include-workspace-root
# then, per package: npm view <name>@<version> deprecated --json
```

Notes:

- Restrict it to direct dependencies. Transitive deprecations are frequent,
  usually not actionable from here, and would make the check noisy.
- Consider running it on a schedule (like the weekly packaging workflow from
  #425) rather than on every PR, so it doesn't block unrelated changes and
  doesn't add registry round-trips to the hot path.
- An allowlist file for knowingly-tolerated deprecations keeps the check from
  becoming a permanently red job while a migration is planned.

## Impact

Turns "a dependency was abandoned" from something spotted by chance during an
unrelated audit into a signal that arrives on its own, while the migration is
still cheap.

NilsR0711:
## Summary

`npm ci` prints npm's deprecation notices, but they scroll past in the install
log without failing or annotating anything, and Dependabot only opens PRs for
version bumps and advisories — never for a package that was simply abandoned.
That is how `dank-twitch-irc` stayed in the tree long after its maintainer had
announced it would receive no further dependency or security updates (#406),
and it was only noticed by chance during an unrelated audit.

This adds a check that surfaces that on its own, while migrating is still cheap.

**`scripts/check-deprecated-deps.mjs`**

- Resolves the *direct* dependencies of the root manifest and of every
  workspace (`dependencies`, `devDependencies`, `optionalDependencies`) to
  their locked versions, then asks the registry
  (`npm view <name>@<version> deprecated --json`) about each one.
- Transitive dependencies are deliberately out of scope: those deprecations are
  frequent, rarely actionable from here, and would make the check noisy enough
  to be ignored.
- Exits non-zero on any deprecation that is not allowlisted, and on any
  dependency whose registry lookup failed — a failed lookup must not read as
  "not deprecated".

**`.github/workflows/deprecated-deps.yml`** — Mondays 06:23 UTC plus manual
dispatch, mirroring the weekly packaging workflow (#425).

**`.github/deprecated-dependencies-allowlist.json`** — currently empty.
Deprecations whose migration is already tracked can be parked here; every entry
requires a `reason` and an `issue`, and entries that no longer apply are
reported as stale warnings so the file does not silently accumulate.

## Architecture decisions

- **Manifests + `package-lock.json` instead of `npm ls`.** The issue suggested
  `npm ls --json --depth 0 --workspaces`. That reports the *installed* tree,
  which flattens workspace links in with real dependencies and — in a git
  worktree nested inside the main checkout — is polluted with `extraneous`
  packages resolved from the parent's `node_modules` (verified locally). The
  committed manifests plus the lockfile are the install-independent source of
  truth, so the workflow can skip `npm ci` entirely and runs in seconds.
- **Nested installs are queried separately.** A workspace that pins its own
  copy of a package installs it under its own `node_modules`; the deprecation
  notice applies to that exact version, so both it and the hoisted one are
  looked up.
- **Outside the `ci-ok` gate.** An upstream deprecation lands independently of
  any change in a PR, so gating PRs on it would block unrelated work on someone
  else's release, and each run costs a registry round-trip per dependency.

## How was this tested?

- New `test/check-deprecated-deps.test.mjs` (23 cases, `node --test`) covers
  dependency collection (hoisted vs. nested installs, shared versions, optional
  deps, workspace links, lockfile drift), allowlist validation (missing
  `allow`, missing `reason`/`issue`, duplicates), the evaluation verdicts
  (violation / tolerated / stale / unchecked), report formatting, and the
  lookup pool — including that a failing lookup is omitted rather than counted
  as clean. It also asserts the committed allowlist parses and that the
  workflow is scheduled + dispatchable but not `pull_request`-triggered.
- Ran the script against the real registry: 84 direct dependencies, ~6s, clean.
- Verified detection end-to-end against two known-deprecated packages
  (`request@2.88.2`, `dank-twitch-irc@4.2.0`) — both reported as `::error::`
  with the registry's message.
- `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm test`, and
  `actionlint` all pass locally.

## Risks

Low: nothing in the PR gate changes. Worst case the weekly job is red on a
deprecation that cannot be acted on immediately, which the allowlist covers.

Closes #494

## Checklist

### DIFF
--- a/.github/deprecated-dependencies-allowlist.json
+++ b/.github/deprecated-dependencies-allowlist.json
@@ -1,0 +1,4 @@
+{
+  "$comment": "Direct dependencies whose npm deprecation is knowingly tolerated while a migration is planned, so the weekly deprecation check (.github/workflows/deprecated-deps.yml) stays actionable instead of permanently red. Every entry needs a reason and a tracking issue; entries that no longer apply are reported as stale and must be dropped.",
+  "allow": []
+}
--- a/.github/workflows/deprecated-deps.yml
+++ b/.github/workflows/deprecated-deps.yml
@@ -1,0 +1,48 @@
+name: Deprecated dependencies
+
+# Asks the npm registry whether any *direct* dependency of the workspace root
+# or of a workspace has been marked deprecated, and fails the run if one has
+# (#494). `npm ci` already prints those notices, but they scroll past in the
+# install log without failing or annotating anything, and Dependabot only
+# opens PRs for version bumps and advisories — never for a package that was
+# simply abandoned. That is how `dank-twitch-irc` stayed in the tree long
+# after its maintainer had announced it would receive no further updates
+# (#406).
+#
+# Deliberately not part of PR CI: an upstream deprecation lands independently
+# of any change here, so gating PRs on it would block unrelated work on
+# someone else's release, and every run costs a registry round-trip per
+# dependency. Weekly is early enough — the point is to learn about an
+# abandoned package while migrating off it is still cheap.
+on:
+  schedule:
+    # Mondays, 06:23 UTC — after the packaging workflow and off the hour, to
+    # avoid GitHub's scheduling peak.
+    - cron: '23 6 * * 1'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: deprecated-deps-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  check:
+    name: Check direct dependencies
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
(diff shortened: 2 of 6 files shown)


NilsR0711:
Follow-up from this change: #514 — the two scheduled workflows (packaging, deprecation check) currently report only into the Actions tab, so a red weekly run can go unnoticed.

PR #528, opened 2026-07-21
feat(control-ui): add a shared focus-visible affordance for buttons

### LINKED ISSUE #508: A11y: control UI has no consistent focus-visible style for buttons
## Summary

`globalStyle.tsx` defines a custom focus affordance for `.stream-list input:focus` / `.stream-list select:focus` (accent border + soft ring), but nothing equivalent exists for buttons. Buttons across the control UI therefore fall back to whatever focus ring the user agent draws — which on the custom-coloured controls (the sidebar stream handle, favorite stars, grid preset buttons) can be low-contrast or barely visible against the surface colour.

## Context

Noticed while resolving #460 (PR #504), which turned the sidebar stream id/drag handle into a real `<button>`. The handle is now tab-reachable, so its focus indicator matters — but its visibility depends entirely on the browser default rather than on the design system.

## Impact

- WCAG 2.4.7 (Focus Visible) / 2.4.11 (Focus Appearance) risk on custom-coloured buttons.
- Inconsistent keyboard affordance: inputs get the accent ring, buttons do not.

## Suggested solution

Add a shared `:focus-visible` rule in `globalStyle.tsx` covering buttons (and any other interactive elements without one), reusing the existing `--accent` / `--accent-soft` tokens so the affordance matches the input styling. Cover it with a component test asserting the rule applies to at least one custom-styled button.

NilsR0711:
## Problem

`globalStyle.tsx` defined an accent focus ring for `.stream-list input:focus` / `select:focus`, but nothing equivalent for buttons. Custom-coloured buttons (sidebar stream handle, favorite stars, grid preset buttons) therefore relied on the user-agent focus ring, which can be low-contrast or invisible against their own background — a WCAG 2.4.7 / 2.4.11 risk and an inconsistent keyboard affordance.

## Solution

- One shared `:focus-visible` rule in `GlobalStyle`, built from the existing `--accent` / `--accent-soft` tokens: a 2px accent outline with `outline-offset: 2px` (so it sits outside the control and stays visible whatever the control's own colours are) plus the soft halo the inputs already draw. It matches the affordance `ResizeHandles` had introduced locally.
- Removed the local `&:focus { outline: none; box-shadow: … }` block in `StyledButton`, which has higher specificity and would otherwise suppress the shared outline on the grid tile buttons. `:focus-visible` also means the ring no longer appears on mouse clicks.

Element-specific rules with higher specificity (`.stream-list input:focus`, `GridInput`) keep their existing styling — those already had a deliberate affordance.

## Testing

- `globalStyle.test.tsx`: asserts the shared rule exists and is built from the accent tokens, and that it actually matches a custom-styled button (the sidebar stream handle) by extracting the rule from the injected CSS and matching its selector against the rendered element.
- The existing tests needed a per-test styled-components target (`StyleSheetManager`): the sheet remembers what it injected, so without it only the first render in the file emits the global CSS.
- Verified red before / green after; full quality gate (format, lint, typecheck, tests) green locally. The three pre-existing `@twurple/*` `import-x/no-unresolved` lint errors are a worktree dependency artifact, unrelated to this change.

## Risks

Low, CSS only. The affordance is now applied globally, so any element without a more specific focus rule gets the accent outline instead of the browser default.

Closes #508

### DIFF
--- a/packages/streamwall-control-ui/src/StyledButton.tsx
+++ b/packages/streamwall-control-ui/src/StyledButton.tsx
@@ -36,10 +36,8 @@ export const StyledButton = styled.button<{
       color: #fff;
     `};
 
-  &:focus {
-    outline: none;
-    box-shadow: 0 0 0 2px var(--accent-soft);
-  }
+  /* No local :focus override - it would suppress the shared :focus-visible
+     outline from GlobalStyle, which is the stronger affordance (see #508). */
 
   svg {
     width: 20px;
--- a/packages/streamwall-control-ui/src/globalStyle.test.tsx
+++ b/packages/streamwall-control-ui/src/globalStyle.test.tsx
@@ -1,9 +1,12 @@
-import { render } from 'preact'
+import { type ComponentChildren, render } from 'preact'
 import { act } from 'preact/test-utils'
+import { StyleSheetManager } from 'styled-components'
 import { afterEach, describe, expect, test } from 'vitest'
+import { StreamList } from './Sidebar.tsx'
 import { GlobalStyle } from './globalStyle.tsx'
 
 let container: HTMLDivElement | undefined
+let styleTarget: HTMLDivElement | undefined
 
 afterEach(() => {
   if (container) {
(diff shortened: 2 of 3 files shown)


NilsR0711:
Follow-up from this change: #531 — `GridInput` still overrides the shared affordance with a hardcoded black `:focus` ring.

Issue #457, opened 2026-07-21
Automate changelog / release-notes generation from Conventional Commits


NilsR0711:
## Background

#405 added a manual `CHANGELOG.md` (Keep a Changelog format) and documented the release-cut steps in `CONTRIBUTING.md`, plus a `test/changelog.test.mjs` backstop. That covered the issue's **minimum (manual)** path. The **automated** path was explicitly deferred to this follow-up.

## Problem

The `## [Unreleased]` section and the compare links must currently be maintained by hand on every merged PR and every release. Since PR titles are already enforced as Conventional Commits (`.github/workflows/pr-title.yml`) and merges are squash merges, the changelog could be derived mechanically instead.

## Proposed approach

Evaluate an automation tool — [`release-please`](https://github.com/googleapis/release-please) is the natural fit — to:

- Generate/append `CHANGELOG.md` entries from Conventional Commit subjects on `main`.
- Open/maintain a release PR that bumps the release-tracked manifests (`packages/streamwall`, `packages/streamwall-control-server`) — must stay consistent with `scripts/release-version.mjs` and `version.test.ts`.
- Populate the GitHub Release body from the changelog section on tag (`.github/workflows/release.yml`).

### Constraints / gotchas

- Only `packages/streamwall` and `packages/streamwall-control-server` track the release line; the other workspaces are pinned (see CONTRIBUTING "Cutting a release"). A monorepo release config must not bump the pinned ones.
- The `vX.Y.Z` tag drives `release.yml`; automation must keep tag ↔ manifest version in sync so `version.test.ts` stays green.
- Once automation owns the changelog, revisit whether `test/changelog.test.mjs` should be relaxed or kept as a guard.

## Impact

Removes manual changelog upkeep, eliminates a class of "forgot to update the changelog" release friction, and gives contributors automatic release-note feedback from their commit convention.

PR #529, opened 2026-07-21
ci(release): warn when the version on main was never tagged

### LINKED ISSUE #515: Warn when a merged release PR was never tagged
## Background

Since #457 (PR #505), release-please runs with `skip-github-release`: it opens
the release PR with the version bump and the changelog, but the `vX.Y.Z` tag
stays a manual step. That is deliberate — a tag pushed with a workflow's
`GITHUB_TOKEN` does not trigger other workflows, so a tag created by
release-please would never start `release.yml` and no installers would be built.

## Problem

Nothing notices when a maintainer merges the release PR and then forgets to push
the tag. `main` then claims a version (`packages/streamwall/package.json`,
`CHANGELOG.md`) that has no GitHub Release and no binaries. The control server's
update check compares its own version against release tags, so a self-hoster on
the previous release is told nothing is available, while `main` says otherwise.

## Suggested solution

A small scheduled or `push: [main]` job that compares the version in
`packages/streamwall/package.json` against the newest `v*` tag and fails (or
opens/updates an issue) when the manifest is ahead. Alternatively, a check on
the release PR itself that reminds the merger of the tag command.

## Impact

Removes the only remaining manual step's silent-failure mode.

NilsR0711:
## Problem

Since #457 (PR #505) release-please runs with `skip-github-release`: it opens the
release PR with the version bump and the changelog, but pushing the `vX.Y.Z` tag
stays a manual step, because a tag created with a workflow's `GITHUB_TOKEN` does
not trigger other workflows and would therefore never start `release.yml`.

Nothing noticed when a maintainer merged the release PR and then forgot the tag.
`main` then advertises a version that has no GitHub Release and no installers,
and the control server's update check — which compares its own version against
the release tags — keeps telling self-hosters that they are up to date.

## Solution

`scripts/check-release-tag.mjs` compares the version in the root manifest
against the `v*` tags and reports a missing tag with the exact command that
fixes it. `.github/workflows/release-tag.yml` runs it daily (07:23 UTC) and on
manual dispatch, and fails the run when the tag is missing.

Decisions:

- **Daily schedule, not `push: [main]`.** The tag follows the release PR merge
  by minutes, so a push-triggered run would flag the normal flow as a failure.
- **24 h grace period.** The check dates the current version by the last commit
  that touched `.release-please-manifest.json` — that file moves in exactly the
  commits that move the release line, unlike `package.json`, which every
  dependency bump touches. Inside the window a missing tag is a `::notice::`,
  after it an `::error::` that fails the job.
- **Fail the run rather than open an issue**, matching `deprecated-deps.yml`,
  the repo's other scheduled hygiene check.
- **`fetch-depth: 0`.** `git tag --list` and the commit date both need the full
  history; a shallow clone would report every release as untagged. An unusable
  commit date throws instead of silently reporting a missing tag.

## Testing

`test/check-release-tag.test.mjs` (12 cases, TDD): tag listing parsing, tagged /
pending / missing classification around the grace-period boundary, rejection of
a non-semver version and of an unreadable commit date, the report annotations,
plus the workflow wiring (schedule + dispatch, never `pull_request`, SHA-pinned
checkout with `fetch-depth: 0`, `contents: read`).

Full quality gate green locally: Prettier, ESLint, `tsc`, `npm test`,
`npm run licenses:check`, `actionlint`. The script also ran against this
checkout (`main is on 0.9.1 and v0.9.1 exists.`).

## Risks

None for the product: the workflow is scheduled only, reads the repository and
touches no release artifacts. Worst case is a false alarm if a release version
is ever bumped without release-please.

Closes #515

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -1,0 +1,51 @@
+name: Release tag
+
+# Checks that the version `main` claims has a matching `vX.Y.Z` tag, and fails
+# the run when it has not had one for a day (#515).
+#
+# Pushing that tag is the one manual step left in the release: release-please
+# runs with `skip-github-release`, because a tag created with a workflow's
+# GITHUB_TOKEN does not trigger other workflows and would never start
+# `release.yml`. Until now nothing noticed when the step was forgotten — `main`
+# then advertises a version with no GitHub Release and no installers, while the
+# control server's update check keeps telling self-hosters they are current.
+#
+# Daily rather than on `push: [main]`: the tag follows the release PR merge by
+# minutes, so a push-triggered run would report the normal flow as a failure.
+# The script's grace period covers the same window for the scheduled run.
+on:
+  schedule:
+    # 07:23 UTC — off the hour, to avoid GitHub's scheduling peak.
+    - cron: '23 7 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-tag-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  check:
+    name: Check the release version is tagged
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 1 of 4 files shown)


NilsR0711:
Follow-up: #533 — the tag check asserts that the tag exists, not that the release behind it actually shipped installers.

Issue #412, opened 2026-07-15
Tech debt: mediaPreload MutationObservers are never disconnected on view teardown


NilsR0711:
## Summary

`mediaPreload.ts` registers `MutationObserver` instances in `lockdownMediaTags` and `waitForQuery` to detect DOM changes in stream pages, but **never calls `observer.disconnect()`** when a view is torn down or reloaded.

## Problem

Stream views reload frequently (operator reload, source URL change, XState transitions). Each reload can register new observers while old ones remain active until GC — along with their throttled callbacks (`lodash` throttle).

## Impact

- Gradual memory/CPU growth during long sessions with many reloads
- Duplicate observers firing on the same document mutations
- Hard-to-reproduce performance issues in multi-tile grids

## Proposed fix

- Track observer handles per view lifecycle
- Disconnect on `VIEW_UNLOAD` / content change message from main process
- Add test asserting disconnect on teardown (mock `MutationObserver`)

## Acceptance criteria

- [ ] Observers disconnected when view content changes or WebContents destroyed
- [ ] No duplicate `lockdownMediaTags` observers after reload
- [ ] Test or documented manual verification steps

PR #530, opened 2026-07-21
fix(preload): keep rescanning iframes for a late-appearing player

### LINKED ISSUE #485: Iframe player scan is one-shot: a late-inserted iframe is never re-examined
## Summary

`waitForVideo()` in `packages/streamwall/src/preload/mediaPreload.ts` watches the top-level document with a `MutationObserver` for a `<video>`/`<audio>` tag, but the iframe fallback runs exactly once — after the initial wait loses its race against `INITIAL_TIMEOUT` (10s). `document.querySelectorAll('iframe')` is evaluated at that single moment.

## Problem

If a page inserts its player iframe later than that (slow SPA bootstrap, consent gate, lazy player), the frame exists but is never scanned again. The tile fails with `could not find video` even though a reachable same-origin `<video>` shows up seconds later.

The same one-shot behaviour also applies to the new cross-origin detection added in #413/#479: an iframe that is still `about:blank` at scan time reports neither the specific nor the correct cause.

## Impact

- Site-specific streams with slow iframe players fail permanently instead of recovering.
- Failure is indistinguishable from "page truly has no video".

## Suggested solution

Fold the iframe scan into the same `MutationObserver`-driven retry loop as the top-level query (re-scan on mutations plus on iframe `load` events) so the fallback keeps trying until the timeout, rather than firing once at the end of it.

Follow-up from #413 / #479.

NilsR0711:
## Problem

`waitForVideo()` watched the top-level document with a `MutationObserver`, but the iframe fallback ran exactly once — after the top-level wait lost its race against `INITIAL_TIMEOUT`. A page that inserts its player iframe later (slow SPA bootstrap, consent gate, lazy player) had that frame scanned at a single moment only, so the tile failed with a generic `could not find video`.

Two further consequences of the one-shot placement:

- An **unbounded** search (`acquireMedia(Infinity)`, used by the `emptied` re-acquisition) never times out, so the iframe scan never ran *at all* — an iframe-embedded player could never be re-acquired.
- Even when the frame was present from the start, its video was only picked up after the full 10 s timeout had elapsed.

## Solution

The iframe scan is folded into the same retry loop as the top-level query:

- `scanForMedia()` does one pass over the document *and* every embedded iframe, returning the media element or the cross-origin verdict.
- `waitForMedia()` drives that scan from the existing throttled `MutationObserver` on `document.body`, plus a `load` listener on every iframe it sees — a frame's own document lies outside the observed tree, so mutations in the embedder never announce that a frame finished loading its player.
- On abort (timeout) the loop takes one final scan before giving up, so a player that appeared inside the throttle window is still found and the reported `crossOriginIframe` verdict is the freshest one rather than an early scan's.
- Teardown removes the iframe `load` listeners alongside disconnecting the observer, keeping the leak discipline established in #412.

`waitForQuery()` is unchanged and still serves `igHacks`.

## Testing

New tests in `mediaPreload.test.ts` (`issue #485`), all red before the fix:

- an iframe-embedded video is acquired without waiting out the initial timeout
- a video from an iframe inserted after the page settled is acquired
- an already-present iframe is rescanned once it fires `load` with its player
- an unbounded re-acquisition scans iframes at all
- the cross-origin cause is still reported when the frame stays unreachable for the whole search

Full quality gate green locally: Prettier, ESLint, tsc, full test suite (all workspaces).

## Risks

No behavioural change for pages whose video is in the top-level document. Iframe-embedded players are now acquired sooner rather than later; the per-scan cost grows by one `querySelector` per iframe every 500 ms, which is negligible for stream pages.

Follow-up from #413 / #479.

Closes #485

### DIFF
--- a/packages/streamwall/src/preload/mediaPreload.test.ts
+++ b/packages/streamwall/src/preload/mediaPreload.test.ts
@@ -358,6 +358,164 @@ describe('mediaPreload iframe video extraction (issue #413)', () => {
   })
 })
 
+describe('mediaPreload late iframe rescanning (issue #485)', () => {
+  // Must match INITIAL_TIMEOUT in mediaPreload.ts; the search gives up once it
+  // elapses, so every late insertion under test happens before it.
+  const INITIAL_TIMEOUT_MS = 10 * 1000
+  // Must match SCAN_THROTTLE in mediaPreload.ts: long enough for the throttled
+  // rescan to run, short enough that these assertions land well before the
+  // timeout would have triggered a one-shot scan on its own.
+  const SCAN_THROTTLE_MS = 500
+
+  beforeEach(() => {
+    vi.useFakeTimers()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+    vi.resetModules()
+    invoke.mockClear()
+    send.mockClear()
+    on.mockClear()
+    exposeInMainWorld.mockClear()
+    document.body.innerHTML = ''
+  })
+
+  function viewErrorCalls() {
+    return send.mock.calls.filter(([channel]) => channel === 'view-error')
+  }
+
+  async function loadVideoContent() {
+    invoke.mockResolvedValueOnce({
+      content: { kind: 'video', link: 'https://example.com/stream' },
+      options: {},
+      volume: 1,
+    })
+    await import('./mediaPreload')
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up issue created for the remaining detection gap: #534 (media search does not observe inside same-origin iframe documents).

PR #532, opened 2026-07-21
fix(streamwall): stage each packaging run in its own temp directory

### LINKED ISSUE #510: Local `npm -w streamwall run package` fails on macOS with an ad-hoc codesign error
## Summary

Running `npm -w streamwall run package` on macOS (Darwin 25.5, Apple Silicon, unsigned/no certificates configured) fails during the fuses step:

```
✔ Copying files
/var/folders/.../electron-packager/tmp-XXXXXX/Electron.app: replacing existing signature
/var/folders/.../electron-packager/tmp-XXXXXX/Electron.app: No such file or directory
In subcomponent: .../Electron.app/Contents/Frameworks/Electron Framework.framework

An unhandled rejection has occurred inside Forge:
Error: Ad-hoc codesign failed with status: 1
    at flipFuses (node_modules/@electron/fuses/dist/index.js:149:19)
    at async node_modules/@electron-forge/plugin-fuses/dist/FusesPlugin.js:35:21
```

The failure happens after "Copying files" completes, while `@electron-forge/plugin-fuses` re-signs the app ad hoc after flipping the fuses.

## Impact

Contributors on macOS cannot run the packaging smoke test locally; the same script passes in CI, which only exercises it on `ubuntu-latest` (`Package smoke (Electron app)` in `.github/workflows/ci.yml` and `release.yml`). The macOS leg is only covered by the weekly `packaging.yml` matrix, so a macOS-specific packaging regression would surface late.

Noticed while working on #431; unrelated to that change (reproduced on a clean `main` checkout).

## Suggested next steps

- Reproduce on another macOS machine to tell an environment problem (stale Electron download cache, quarantine attributes, `codesign` toolchain) apart from a real configuration issue with `FusesPlugin` / `packagerConfig`.
- If it is reproducible, either pin/adjust the fuses plugin signing behavior for unsigned local builds or document the workaround in `CONTRIBUTING.md`.
- Consider whether the `Package smoke` job should also run on `macos-latest` in PR CI, or whether the weekly `packaging.yml` matrix is deliberate coverage enough.

NilsR0711:
## Summary

`npm -w streamwall run package` aborted with a misleading macOS signing error
whenever a second packaging run was active on the same machine:

```
/var/folders/.../electron-packager/tmp-XXXXXX/Electron.app: No such file or directory
Error: Ad-hoc codesign failed with status: 1
    at flipFuses (node_modules/@electron/fuses/dist/index.js:149:19)
```

Root cause is not signing at all: `@electron/packager` stages every run under
`<tmpdir>/electron-packager` and deletes that entire base directory when a run
starts (`Packager.ensureTempDir()` → `fs.remove(this.tempBase)`). A second run
therefore wipes the staging tree of the first one mid-run, and the fuses plugin
then ad-hoc re-signs an `Electron.app` that no longer exists. On a machine with
several worktrees this happens easily; a single run in isolation always
succeeded, which is why the report looked environment-specific.

Reproduced deterministically on macOS 26.5 / arm64 by removing
`$TMPDIR/electron-packager` while a run was in its "Copying files" phase — the
failure is byte-for-byte the one reported in #510.

## Technical solution

- New `packages/streamwall/forge.tmpdir.ts`: `createPackagingTmpdir()` creates a
  fresh `streamwall-packager-*` base per run, `removePackagingTmpdir()` deletes
  it again behind a prefix guard so a recursive delete can never escape.
- `forge.config.ts` sets `packagerConfig.tmpdir` to that directory and drops it
  in a new `postPackage` hook — by then packager has moved the staged app to
  `out/`, so only an empty shell is left.

Alternative considered: `tmpdir: false` (stage straight into `out/`). Rejected
because an aborted run would leave a half-built `.app` behind in the output
directory that the makers operate on.

## How was this tested?

- `npx vitest run forge.tmpdir.test.ts forge.config.test.ts -w packages/streamwall`
  — new unit tests for the helper (own directory per run, recursive removal,
  idempotent removal, foreign-path guard) and for the forge wiring.
- `npm run lint`, `npm run format`, `npm run typecheck`, `npm test` — all green.
- Manual verification on macOS 26.5 / arm64: with the fix in place, deleting
  `$TMPDIR/electron-packager` mid-run no longer affects the run; packaging
  finishes, `out/Streamwall-darwin-arm64/Streamwall.app` is ad-hoc signed with
  the expected fuses flipped, and the run's temp directory is gone afterwards.

## Risks / breaking changes

None for CI or releases: the packager option is documented and platform
independent, and the smoke/make jobs keep running unchanged. Worst case a
crashed run leaves an empty `streamwall-packager-*` directory in the system temp
directory, which the OS cleans up.

Closes #510

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -278,6 +278,13 @@ in two other places:
   tab (optionally with the `debug` input for verbose Forge logging) after a
   Forge/maker dependency bump.
 
+Two packaging runs can share a machine: `packagerConfig.tmpdir` points each
+run at a directory of its own (`forge.tmpdir.ts`). @electron/packager
+otherwise stages every run under the same `<tmpdir>/electron-packager` and
+deletes that directory when a run starts, which used to break a concurrent run
+with a misleading `Ad-hoc codesign failed with status: 1` from the fuses
+plugin (#510).
+
 ### Dependency deprecations
 
 `.github/workflows/deprecated-deps.yml` asks the npm registry every Monday
--- a/packages/streamwall/forge.config.test.ts
+++ b/packages/streamwall/forge.config.test.ts
@@ -4,6 +4,12 @@ import { describe, expect, it, vi } from 'vitest'
 const runTypecheck = vi.fn()
 vi.mock('./forge.typecheck', () => ({ runTypecheck: () => runTypecheck() }))
 
+const removePackagingTmpdir = vi.fn()
+vi.mock('./forge.tmpdir', () => ({
+  createPackagingTmpdir: () => '/tmp/streamwall-packager-test',
+  removePackagingTmpdir: (dir: string) => removePackagingTmpdir(dir),
+}))
+
 const { default: config }: { default: ForgeConfig } =
   await import('./forge.config')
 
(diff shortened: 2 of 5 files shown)


NilsR0711:
Follow-up from the open question in #510 (whether the PR-level package smoke should also run on macOS): #538.

PR #544, opened 2026-07-21
fix(control-ui): adopt the shared focus-visible ring for grid cells

### LINKED ISSUE #531: A11y: GridInput focus ring is hardcoded black and uses :focus instead of :focus-visible
## Summary

`GridInput` (`packages/streamwall-control-ui/src/GridInput.tsx`) styles its focus state with a hardcoded, non-token affordance:

```css
&:focus {
  outline: 1px solid black;
  box-shadow: 0 0 5px black inset;
  z-index: 100;
}
```

Two problems:

- It uses `:focus`, not `:focus-visible`, so the ring also appears on pointer interaction — unlike the shared affordance introduced in #508 / #528.
- The colours are hardcoded `black` rather than the `--accent` / `--accent-soft` design tokens, and the rule has higher specificity than the shared `:focus-visible` rule, so it wins. Against a dark tile colour a 1px black outline can be effectively invisible.

## Context

Noticed while resolving #508 (PR #528), which added one shared `:focus-visible` rule to `globalStyle.tsx`. `GridInput` was deliberately left untouched there — its black ring is a long-standing, intentional choice for legibility over the tile colours — but it is now the remaining inconsistency.

## Impact

- WCAG 2.4.11 (Focus Appearance) risk on dark tile colours.
- Keyboard affordance in the grid differs from the rest of the control UI.

## Suggested solution

Evaluate whether `GridInput` can adopt the shared affordance (switch to `:focus-visible`, replace `black` with the accent tokens, keep `z-index: 100` so the ring is not clipped by neighbouring tiles), or — if the tile colours genuinely need a different treatment — derive a contrast-safe outline colour from the tile colour instead of hardcoding black. Cover the outcome with a component test.

NilsR0711:
## Problem

`GridInput` styled its focus state with a hardcoded, non-token affordance:

```css
&:focus {
  outline: 1px solid black;
  box-shadow: 0 0 5px black inset;
  z-index: 100;
}
```

Two issues (#531):

- It matched `:focus`, not `:focus-visible`, so the keyboard ring also appeared on plain pointer interaction — unlike the shared affordance added in #508 / PR #528.
- The colours were hardcoded `black` instead of the `--accent` / `--accent-soft` tokens, and because the rule sat on the component class it out-specified the shared global `:focus-visible` rule. The grid was the last place in the control UI with a divergent keyboard affordance.

## Solution

`GridInput` now adopts the shared affordance and contributes no focus colours of its own. Two details make that work:

- **The idle cell divider is scoped to `:not(:focus-visible)`.** The component class rule and the global `:focus-visible` rule are both single-class / single-pseudo-class selectors, so specificity ties — an unscoped `outline: 1px solid rgba(0,0,0,.5)` would win or lose purely by stylesheet injection order. Scoping it out of the focused state removes the ambiguity instead of relying on ordering luck.
- **The `:focus-visible` block keeps `z-index: 100` and gains `position: relative`.** The shared ring is drawn *outside* the control (`outline-offset: 2px` plus a 4px halo), so a focused cell has to rise above its neighbouring tiles or the ring is clipped. The pre-existing `z-index: 100` never actually applied — `z-index` only affects positioned elements, and the input is `static` (only its wrapper is absolute). Adding `position: relative` fixes that latent no-op.

## Architecture decisions

- **Adopt rather than re-derive.** The issue offered an alternative: derive a contrast-safe outline colour from the tile colour. Not needed here — the tile background is always `lightness(75)` (or `90` when highlighted), i.e. reliably light, so the `--accent` ring (`#d6402a` light / `#f24d2e` dark) has ample contrast against every tile. Deriving a per-tile colour would reintroduce the divergence the issue is about.
- **No duplication of the shared ring declarations** in the component — the whole point is a single source of truth in `globalStyle.tsx`.

## Testing

New `GridInput` component tests covering the emitted CSS (same approach as the existing `globalStyle.test.tsx` rules):

- no bare `:focus` rule is emitted, so pointer interaction draws no keyboard ring
- the component's `:focus-visible` block declares no `outline` / `box-shadow` / `black`, leaving the shared ring to apply
- the idle divider is confined to `:not(:focus-visible)` so it cannot beat the shared rule
- the focused cell gets both `position: relative` and `z-index: 100`
- with `GlobalStyle` mounted, the accent ring rule matches the grid-cell input

Full quality gate green locally: `npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` (510 tests across all workspaces).

## Risks / breaking changes

None functionally. Visual change only: the focus ring in the grid becomes the accent ring instead of a black outline, and no longer appears on click. The `data-testid` / `data-idx` E2E hooks are untouched, and no E2E test asserts on focus styling.

Closes #531

### DIFF
--- a/packages/streamwall-control-ui/src/GridInput.test.tsx
+++ b/packages/streamwall-control-ui/src/GridInput.test.tsx
@@ -1,10 +1,13 @@
-import { render } from 'preact'
+import { type ComponentChildren, render } from 'preact'
 import { act } from 'preact/test-utils'
+import { StyleSheetManager } from 'styled-components'
 import { afterEach, describe, expect, test, vi } from 'vitest'
 
 import { GridInput } from './GridInput.tsx'
+import { GlobalStyle } from './globalStyle.tsx'
 
 let container: HTMLDivElement | undefined
+let styleTarget: HTMLDivElement | undefined
 
 afterEach(() => {
   if (container) {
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up from this change: #553 — the sidebar text inputs still use `:focus` and clear the shared outline with `outline: none`, making them the remaining divergent keyboard affordance in the control UI.

Issue #453, opened 2026-07-20
Release quality gate never exercises the makers, so a maker regression only surfaces during tag publish


NilsR0711:
## Problem

The release workflow's quality-gate job runs `npm -w streamwall run package` as a smoke test before the three-OS publish matrix fans out, but packaging stops short of the maker stage: the NSIS maker (electron-builder in prepackaged mode, added in #432), the ZIP/deb/rpm makers, and the `postMake` update-metadata hook only ever run inside the publish jobs themselves.

A regression in any maker or in `forge.updateMetadata.ts`'s integration therefore first surfaces while a tagged release is half-published — the most expensive possible moment (partially created GitHub release, re-tagging needed).

## Suggested solution

Add a `make` smoke test:

- cheapest: extend the release quality-gate job with `npm -w streamwall run make` (Linux runner exercises deb/rpm + hook),
- better: also cross-build the Windows NSIS target on the Linux/macOS runner (`make -- --platform=win32`), which works because electron-builder bundles makensis for every host OS — this was verified manually on macOS for #432.

Keeping it in the release workflow (not PR CI) preserves the change-scaled PR pipeline; a scheduled or release-branch-only job would also do.

NilsR0711:
Fixed by #498. The remaining gap — the darwin zip maker and `latest-mac.yml`, which cannot be cross-built from the Linux gate runner — is tracked in #517.

PR #550, opened 2026-07-21
ci(release): check that the release tag produced usable assets

### LINKED ISSUE #533: Warn when a v* tag produced no release assets
## Background

`.github/workflows/release-tag.yml` (#515, PR #529) now catches the case where
`main` claims a version that was never tagged. It only asserts that a matching
`vX.Y.Z` tag exists — it does not look at what the tag produced.

## Problem

A tag can exist while the release is still broken or incomplete:

- `release.yml` failed part-way through the three-platform publish matrix, so
  the GitHub Release exists but is missing one platform's installers (the
  partially populated release described in #453).
- `release.yml` never ran at all for the tag (pushed from a workflow token, or
  the run was cancelled), so there is a tag with no GitHub Release behind it.

Both leave self-hosters and the app's updater pointed at a release that cannot
be installed, and neither is reported anywhere.

## Suggested solution

Extend `scripts/check-release-tag.mjs` (or add a sibling) to ask the GitHub API
for the release belonging to the newest `v*` tag and assert that it exists, is
not a draft, and carries the expected asset kinds (`.deb`, `.rpm`,
`*-setup-*.exe`, `latest.yml`, the macOS zip) — the same artifact list the
`make` job in `release.yml` already verifies. The daily job in
`release-tag.yml` is the natural place to run it.

## Impact

Closes the second half of the release pipeline's silent-failure surface: not
only "the tag is missing" but also "the tag shipped nothing usable".

NilsR0711:
## Problem

`release-tag.yml` (#515) checks only that the version on `main` has a matching
`vX.Y.Z` tag — it never looks at what the tag produced. A tag can exist while
the release behind it is unusable:

- a leg of `release.yml`'s three-platform publish matrix failed, so the release
  is missing that platform's installers (the partially populated release of
  #453),
- `release.yml` never ran for the tag (pushed from a workflow token, or the run
  was cancelled), so there is no release behind it,
- the release was never taken out of draft, so neither the updater nor a
  self-hoster can see it.

All three point the app's updater and self-hosters at a release that cannot be
installed, and none of them were reported anywhere.

## Solution

`scripts/check-release-assets.mjs` asks the GitHub API for the release of the
version on `main` and asserts the artifact kinds the `make` job in `release.yml`
already builds: `*.deb`, `*.rpm`, `*-setup-*.exe`, `latest.yml`, `*.zip`,
`latest-mac.yml`. It runs as a second step of the daily `release-tag.yml` job,
with `if: ${{ !cancelled() }}` so a missing tag cannot hide a broken release.

### Decisions

- **Anchored on the version `main` claims, not on the highest tag.** The issue
  suggested "the newest `v*` tag", but the repository still carries
  `v2.0.0-pre1..3` from the project this started as; those sort above the
  current release line and never had a GitHub Release here, so the newest-tag
  reading reports a permanent false failure. The version on `main` is the
  release that matters and is the same anchor `check-release-tag.mjs` uses.
- **A version without a tag is skipped**, because that is exactly what
  `check-release-tag.mjs` reports — raising it twice in one run would be noise.
- **Releases before 0.9.2 are skipped.** v0.9.1 and earlier were published with
  Squirrel (`Streamwall-0.9.1.Setup.exe`, `RELEASES`, `.nupkg`) and have no
  `latest*.yml` at all (#454, and the upgrade notice in `CONTRIBUTING.md`).
  They cannot be rebuilt, so checking them would leave the daily run
  permanently red over something nobody can fix.
- **Assets still uploading do not count.** GitHub lists an asset as soon as its
  upload starts; only `state: "uploaded"` ones can be downloaded.
- **Patterns are anchored**, so `*.zip` is not satisfied by
  `installer.zip.blockmap`.

## Testing

- `test/check-release-assets.test.mjs` (21 tests): tag selection incl. the
  inherited `v2.0.0-pre*` tags and the legacy floor, the four release verdicts
  (complete / no release / draft / incomplete), partially uploaded assets,
  pattern anchoring, remote-URL parsing, every report shape, and a workflow
  assertion that the step runs with a token and survives a failing tag check.
- `npm run format:check`, `npm run lint`, `npm run typecheck`, `npm test` and
  `actionlint` all pass locally.
- Ran against the live repository at three stages: it correctly reported
  `v2.0.0-pre3` as having no release, reported `v0.9.1` as missing the NSIS and
  `latest*.yml` artifacts before the legacy floor was added, and is quiet on
  `main` today.

## Risks

None at runtime — the script only reads. The first release from 0.9.2 onwards
is the first one this actually inspects; if its artifact list ever diverges
from `release.yml`'s `make` matrix, `EXPECTED_ASSET_PATTERNS` has to follow.

Closes #533

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -50,6 +50,16 @@ jobs:
       - name: Check the release version is tagged
         run: node scripts/check-release-tag.mjs
 
+      # A tag alone proves nothing: the publish matrix may have failed
+      # half-way through, or never run for the tag at all (#533). Runs even
+      # when the step above failed — the two report different halves of the
+      # same pipeline, and one missing tag must not hide a broken release.
+      - name: Check the newest tag produced a complete release
+        if: ${{ !cancelled() }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: node scripts/check-release-assets.mjs
+
   # A failing scheduled run is invisible without this: nobody is emailed about
   # one, so the report goes into the issue tracker instead (#514) — which is
   # also where a forgotten tag belongs.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -374,6 +374,19 @@ up to date. Run it from the Actions tab, or locally:
 node scripts/check-release-tag.mjs
 ```
 
+The same job then checks what that tag actually produced: a tag can exist while
+its release is still broken — a publish leg of `release.yml` failed and left a
+partially populated release behind (#453), the workflow never ran for the tag,
+or the release was never taken out of draft. The check asks the GitHub API for
+the release of the version on `main` and asserts the artifact kinds the `make`
+job builds (`*.deb`, `*.rpm`, `*-setup-*.exe`, `latest.yml`, `*.zip`,
+`latest-mac.yml`). Releases before v0.9.2 carry Squirrel's artifact names and
+are skipped.
+
+```sh
+node scripts/check-release-assets.mjs
+```
+
 `test/changelog.test.mjs` and `test/release-please.test.mjs` run as part of
 `npm test` and fail if `CHANGELOG.md` is missing a heading for the version
 currently in `packages/streamwall/package.json`, if a hand-maintained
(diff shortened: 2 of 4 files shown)


NilsR0711:
Follow-up from this change: #554 — the inherited `v2.0.0-pre*` tags are why this check anchors on the version `main` claims instead of the newest tag, as the issue originally suggested.

Issue #551, opened 2026-07-21
A11y: keyboard focus in the wall grid is invisible because the cell layer only fades in on hover


NilsR0711:
## Summary

`StyledGridInputs` in `packages/streamwall-control-ui/src/components/ControlGrid.tsx`
is the layer that holds the per-cell placement inputs. It renders at
`opacity: 0` and is only revealed by a hover rule on the grid container:

```css
const StyledGridInputs = styled.div`
  opacity: 0;
  transition: opacity 100ms ease-out;
  overflow: hidden;
  z-index: 100;
`

const StyledGridContainer = styled.div`
  &:hover ${StyledGridInputs} {
    opacity: 0.35;
  }
`
```

There is no `:focus-within` counterpart. A keyboard user who tabs into a grid
cell therefore focuses a fully transparent control: neither the cell value nor
its focus ring is rendered at all, and the focus position in the grid cannot be
perceived without a pointer.

## Context

Found while resolving #531, which replaced the tile's hardcoded black `:focus`
ring with a tile-derived `:focus-visible` ring. That fix makes the ring
contrast-safe, but the ring is still painted into a layer that is transparent
unless the pointer happens to be over the grid.

## Impact

- WCAG 2.4.7 (Focus Visible) failure for the primary interaction surface of the
  control UI — the ring fix from #531 cannot be seen without a mouse.
- Keyboard-only placement of streams (tab to a cell, type a stream id) is
  effectively blind.

## Suggested solution

Reveal the layer while it holds focus, e.g.

```css
  &:focus-within ${StyledGridInputs} {
    opacity: 1;
  }
```

placed after the hover rule so it wins on equal specificity. Full opacity rather
than the hover value of `0.35` is deliberate: at 35% the ring's contrast is
diluted along with everything else, which would undo #531. Whether focus should
reveal the layer fully or at some intermediate value is the design decision to
settle here. Cover the outcome with a component test.

PR #552, opened 2026-07-21
fix(control-ui): derive the grid focus ring from the tile colour

### LINKED ISSUE #531: A11y: GridInput focus ring is hardcoded black and uses :focus instead of :focus-visible

NilsR0711:
## Problem

`GridInput` styled its focus state with a hardcoded, non-token affordance:

```css
&:focus {
  outline: 1px solid black;
  box-shadow: 0 0 5px black inset;
  z-index: 100;
}
```

Two defects: it triggers on `:focus`, so the ring also appears on pointer
interaction — unlike the shared affordance added in #508 / #528 — and its
colour is fixed, so contrast depends on whichever tile colour it lands on.
The rule also outranks the shared `:focus-visible` rule on specificity, so the
grid could never inherit it.

## Solution

The rule now matches the shared affordance from `globalStyle.tsx` in **shape
and trigger** — a 2px ring plus a halo, keyboard focus only — but keeps its own
colours.

Adopting `--accent` outright was evaluated and rejected: a tile's hue is
derived from the stream id (`idColor`), so the accent red can land on a
near-identical light red and fall below the 3:1 that WCAG 2.4.11 asks for. The
new `focusRingColors()` therefore takes the issue's second option and derives
the colours from the tile — whichever of black/white contrasts more becomes the
ring, the other one the halo just inside it — so the pair reads on every tile
the grid can paint.

The ring is drawn inwards (`outline-offset: -2px`) rather than outwards: the
tiles are laid out edge to edge inside an `overflow: hidden` layer, where an
outset ring would be overlapped by its neighbours and clipped at the grid
border. That also retires the `z-index: 100` the old rule carried, which never
had an effect — the input is statically positioned, so `z-index` was ignored.

`tileColor()` is extracted so the background and the ring derive from the same
value instead of repeating the lightness maths.

## Testing

`GridInput.test.tsx` gains component tests that inspect the emitted CSS
(the `globalStyle.test.tsx` `StyleSheetManager` pattern) plus unit tests for
the derivation:

- the rule is `:focus-visible` and no bare `:focus` rule remains
- the ring is 2px, contains no `black`, and clears 3:1 against its tile
- `outline-offset: -2px` keeps the ring inside the tile
- ring and halo contrast against each other
- `focusRingColors()` clears 3:1 against every tile colour the grid can paint
  (65 stream ids × both highlight states) and flips to a light ring on a dark
  tile

Full quality gate green locally: `npm run lint`, `npm run format:check`,
`npm run typecheck`, `npm test`.

## Risks

Visual only, scoped to the grid cells. The ring is now thicker and sits inside
the tile edge instead of on it.

## Follow-up

- #551 — keyboard focus in the grid is still invisible because the cell layer
  renders at `opacity: 0` unless hovered. Out of scope here (it is a separate
  rule in `ControlGrid.tsx` with its own design decision), but it gates whether
  this ring can be seen at all without a pointer.

Closes #531

### DIFF
--- a/packages/streamwall-control-ui/src/GridInput.test.tsx
+++ b/packages/streamwall-control-ui/src/GridInput.test.tsx
@@ -1,10 +1,13 @@
 import { render } from 'preact'
 import { act } from 'preact/test-utils'
+import { Color, idColor } from 'streamwall-shared'
+import { StyleSheetManager } from 'styled-components'
 import { afterEach, describe, expect, test, vi } from 'vitest'
 
-import { GridInput } from './GridInput.tsx'
+import { focusRingColors, GridInput, tileColor } from './GridInput.tsx'
 
 let container: HTMLDivElement | undefined
+let styleTarget: HTMLDivElement | undefined
 
 afterEach(() => {
   if (container) {
@@ -12,6 +15,11 @@ afterEach(() => {
     container.remove()
     container = undefined
   }
+  styleTarget?.remove()
+  styleTarget = undefined
+  document
+    .querySelectorAll('style[data-styled]')
+    .forEach((style) => style.remove())
 })
 
 function renderInput(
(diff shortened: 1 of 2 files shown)


NilsR0711:
Superseded: #531 was fixed concurrently by #544, which landed on `main` while this branch was in CI. #544 takes the issue's first option (adopt the shared `--accent` ring, scope the cell divider to `:not(:focus-visible)`, raise the focused cell with `position: relative; z-index: 100`); this branch took the second one (derive the ring colour from the tile). Closing in favour of the merged fix rather than re-litigating it here.

The residual contrast question is filed separately, and the follow-up found along the way (#551, the cell layer renders at `opacity: 0` unless hovered) still applies to `main`.

PR #555, opened 2026-07-21
refactor(control-ui): type stateIdxMap and separate the view-id / cell-index axes

### LINKED ISSUE #507: Type stateIdxMap at construction and separate the viewId / cell-index axes
## Problem

`useStreamwallState` builds the cell-index map without a type parameter and fills it with placeholder objects:

```ts
const stateIdxMap = new Map()
...
if (!stateIdxMap.has(space)) {
  stateIdxMap.set(space, {})
}
Object.assign(stateIdxMap.get(space), viewInfo)
```

(`packages/streamwall-control-ui/src/streamwallState.tsx`)

The map is only declared as `Map<number, ViewInfo>` on the `StreamwallConnection` interface, so inside the builder every value is `any`. The `{}`-then-`Object.assign` construction also means a cell could, in principle, hold a partially populated `ViewInfo`.

## Impact

This is the type hole that let #470 through: passing a grid **cell index** where a **stable view id** was expected type-checked fine, because nothing forced the two numeric axes apart and the map's element type was unchecked at the construction site. Both `number` axes remain structurally identical today, so the same class of mix-up can recur anywhere either value is threaded through.

## Suggested solution

- Type the map at construction (`new Map<number, ViewInfo>()`) and build each entry as a complete `ViewInfo` instead of mutating a `{}` placeholder.
- Optionally, introduce branded types (e.g. `type ViewId = number & { __brand: 'ViewId' }` and `type CellIdx = number & { __brand: 'CellIdx' }`) so the compiler rejects passing one where the other is expected. See the `viewId` vs. cell-index split established in #397/#467.

## Notes

Found while fixing #470 (PR #501). Out of scope there — that PR fixes the call sites, not the type model that allowed them.

NilsR0711:
## Summary

`useStreamwallState` built its cell index with an untyped `new Map()` and filled it by `Object.assign`-ing each view onto a `{}` placeholder. Inside the builder every value was therefore `any`, and a cell could in principle end up holding a partially populated `ViewInfo`. The map is now constructed as `Map<CellIdx, ViewInfo>` and each occupied cell stores the view object itself.

That type hole is what let #470 through: a grid **cell index** and a stable **view id** are both `number`, so nothing stopped one being passed where the other was expected. The two axes are now branded (`CellIdx`, `ViewId` in the new `viewAddressing.ts`) and threaded through the addressing surface:

- `stateIdxMap` keys, `ViewInfo['spaces']` and `fullscreenViewIdx` (the anchor *cell* of the expanded view, despite its name)
- every `useGridCommands` handler — `viewId` for the `set-view-*`/`reload-view`/`dev-tools` commands, `CellIdx` for `handleSetView`
- the grid components and gestures: `ControlGrid`, `GridControls`, `GridInput`, `useTileDrag`, `useControlHotkeys`

Passing a cell index to a view-id command is now a compile error, and vice versa.

### Architecture decisions

- **Branding stays inside `streamwall-control-ui`.** The shared schemas keep plain `number`s on the wire; the brands are applied where the two axes actually meet in the UI, which is where #470 happened. No runtime cost — `asViewId`/`asCellIdx` are compile-time tags.
- **Pure grid geometry keeps plain `number`s.** `gestures.ts`, `gridInteractions.ts`, `viewPlacement.ts` and `useTileResize.ts` are untouched: both brands are assignable *to* `number`, so only the boundaries where a raw number enters the addressed surface (hotkey trigger index, `cols * y + x`, `computeHoveringIdx`, the wire's `fullscreenViewIdx`) convert explicitly. This keeps the index arithmetic readable instead of littering it with conversions.
- **No behavior change.** Storing the view object instead of merging onto a placeholder is equivalent for a shared cell (the later view already won every field); the new test pins that ordering.

## How was this tested?

- New `viewAddressing.test.ts`: runtime identity of the tag helpers plus `@ts-expect-error` guards that fail the **typecheck** if either axis ever becomes interchangeable again (an `@ts-expect-error` that stops erroring is itself a compile error).
- New `streamwallState.test.tsx` cases: every occupied cell maps to the view itself (multi-space views included), and a later view claiming an already-taken cell wins.
- `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm test`, `npm -w streamwall-control-client run build` and `npm run licenses:check` all pass locally.

Closes #507

## Checklist

### DIFF
--- a/packages/streamwall-control-ui/src/GridControls.test.tsx
+++ b/packages/streamwall-control-ui/src/GridControls.test.tsx
@@ -2,6 +2,7 @@ import { render } from 'preact'
 import { act } from 'preact/test-utils'
 import { afterEach, describe, expect, test, vi } from 'vitest'
 import { GridControls } from './GridControls.tsx'
+import { asCellIdx, asViewId } from './viewAddressing.ts'
 
 // react-icons renders through preact/compat's Context.Consumer, which
 // currently crashes under this package's happy-dom test environment
@@ -35,8 +36,8 @@ function renderControls(
   act(() => {
     render(
       <GridControls
-        idx={0}
-        viewId={0}
+        idx={asCellIdx(0)}
+        viewId={asViewId(0)}
         streamId="abc"
         style={{}}
         isDisplaying={true}
@@ -85,7 +86,12 @@ describe('GridControls volume slider', () => {
     const onSetVolume = vi.fn()
     // viewId deliberately differs from idx to prove the command carries the
     // stable view id, not the grid cell index (issue #397).
-    const box = renderControls({ idx: 0, viewId: 3, volume: 1, onSetVolume })
+    const box = renderControls({
+      idx: asCellIdx(0),
+      viewId: asViewId(3),
+      volume: 1,
+      onSetVolume,
+    })
     const slider = box.querySelector('input[type="range"]') as HTMLInputElement
 
     act(() => {
@@ -100,7 +106,11 @@ describe('GridControls volume slider', () => {
 describe('GridControls double-click to toggle fullscreen', () => {
   test('toggles fullscreen for this tile when its open area is double-clicked', () => {
     const onToggleFullscreen = vi.fn()
-    const box = renderControls({ idx: 0, viewId: 4, onToggleFullscreen })
+    const box = renderControls({
+      idx: asCellIdx(0),
+      viewId: asViewId(4),
+      onToggleFullscreen,
+    })
     const controls = box.firstElementChild as HTMLElement
 
     act(() => {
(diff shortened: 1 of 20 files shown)


NilsR0711:
Superseded by #560 (which brands the `ViewId` / `CellIdx` axes in `streamwall-shared` rather than inside `streamwall-control-ui`) and #572 (which extends the `CellIdx` brand to the cell-geometry helpers). Issue #507 is already closed.

Everything this PR set out to do is on `main`: `stateIdxMap` is constructed as `Map<CellIdx, ViewInfo>`, each occupied cell stores the view object itself, the brands are threaded through the grid, hotkey and gesture surfaces, and the `@ts-expect-error` compile guards for both axes live in `streamwallState.test.tsx`. Rebasing would only produce conflicts against already-identical code and an empty diff.

The one case not carried over — pinning that a later view claiming an already-occupied cell wins — will follow as a separate test-only change.

PR #556, opened 2026-07-21
test(control-server): cut the per-test cost of the live-server integration suites

### LINKED ISSUE #536: test(control-server): cut the per-test cost of the live-server integration suites
## Context

While fixing #492 (a whole test file being cancelled by `--test-timeout`), the
wall-clock cost of the control-server integration suites was measured. The
timeout was raised to 60s, which removes the flake, but the underlying cost is
still high and will creep back toward any ceiling as suites grow.

## Measurements

`src/wsValidation.test.ts` (11 tests, each booting a live server) takes 6–13s
locally. A stripped-down probe of the shared setup shows where the time goes,
per test:

| Step                                             | Cost    |
| ------------------------------------------------ | ------- |
| `buildTestApp` + `listenTestApp`                  | ~5 ms   |
| `connectStreamwallUplink` (token + handshake)     | ~210 ms |
| `redeemInviteAndConnectClient` (token + redeem)   | ~160 ms |

Two things dominate, both avoidable in a test context:

1. **scrypt.** `auth.ts` pins `N: 16384, r: 8, p: 1`. Every test pays roughly
   four derivations (mint uplink token, verify it on connect, mint invite,
   verify it on redeem), ~50 ms each — ~200 ms per test that has nothing to do
   with what is under test.
2. **Fixed sleeps.** `connectStreamwallAndClient` waits a blind `delay(150)`
   after seeding the uplink state and `delay(50)` after connecting the client
   — ~200 ms per test, and a source of its own flakiness on a slow machine,
   since the wait is not tied to an observable signal.

Across the control-server suites that is on the order of tens of seconds of
pure overhead per run, locally and in CI.

## Suggested solution

- Allow the scrypt work factor to be lowered for tests only, without weakening
  the production default — e.g. an injected parameter on the `Auth` factory
  (`initApp` already takes injected dependencies), not an env var that a
  deployment could accidentally set. Cover the production default with a test
  so it cannot be lowered silently.
- Replace the blind `delay()`s in the test helpers with waits on an observable
  signal (the uplink state being registered, the client's first state frame),
  keeping a bounded timeout as the failure path. `recordJsonMessages().waitFor`
  already provides the pattern.

## Impact

Faster feedback locally and in CI, and less timing-sensitivity in the suites
that have historically produced flakes (#221, #492).

NilsR0711:
## Summary

The control-server integration suites boot a live server per test and paid two
avoidable costs, measured in #536: the production scrypt work factor for tokens
they never assert on (~200ms per test, four derivations), and blind `delay()`
calls that were both slow (~200ms per test) and timing-sensitive, since the
wait was not tied to any observable signal.

**Injectable work factor.** `SCRYPT_PARAMS` was a module constant; it is now
`DEFAULT_SCRYPT_PARAMS` plus a `scryptParams` option on the `Auth` factory,
passed through `initApp` alongside the other injected dependencies. It is
deliberately not reachable from the environment, so no deployment can end up
with a lowered work factor by accident, and `buildTestApp` is the only caller
that injects a cheap one. Two specs pin the production behavior: the stored
hash of a default `Auth` must be reproducible with `N=16384, r=8, p=1`, and
`initApp` must keep that default when nothing is injected. A third spec
documents that the work factor is part of the stored hash — tokens minted
under a different one no longer validate.

**Waits on observable signals.** `captureLogs` grows the `waitFor` /
`waitForMessage` pattern `recordJsonMessages` already provides: entries
captured before the wait satisfy it (no listener-attachment race) and a
bounded timeout stays the failure path. The suites now wait for the log entry
that marks the step as finished — the uplink accepting or rejecting a seeded
state, a client connecting, a send failure being caught — instead of sleeping.
Where the awaited entry *was* the assertion, the redundant `hasMessage` check
was dropped in favour of the wait itself.

The one remaining fixed delay (`wsRateLimit.test.ts`) asserts that nothing
happens within a window, which no signal can replace.

## How was this tested?

`npm run lint`, `npm run format:check`, `npm run typecheck` and `npm test`
(all workspaces) pass locally; the control-server suite is 173/173 green.

Wall-clock for `npm test --workspace streamwall-control-server`, two runs each
on the same machine:

| | Before | After |
| --- | --- | --- |
| Full control-server suite | 26.3s / 30.2s | 18.2s / 21.1s |
| `src/wsValidation.test.ts` alone | 6.0s | 2.5s |

New coverage: production-default guards for `Auth` and `initApp`, a
cross-work-factor rejection case, and `src/testHelpers.test.ts` for the new
log-capture waits (already-captured entry, later entry, substring match,
bounded timeout, independent waiters).

## Checklist

Closes #536

### DIFF
--- a/packages/streamwall-control-server/src/auth.test.ts
+++ b/packages/streamwall-control-server/src/auth.test.ts
@@ -6,7 +6,9 @@ import { promisify } from 'node:util'
 import type { StreamwallRole, StreamwallState } from 'streamwall-shared'
 import {
   type AuthToken,
+  type ScryptParams,
   Auth,
+  DEFAULT_SCRYPT_PARAMS,
   rand62,
   StateWrapper,
   uniqueRand62,
@@ -23,6 +25,27 @@ const base62 = baseX(
 )
 async function legacyStoredHash(secret: string, salt: string): Promise<string> {
   return base62.encode((await scrypt(secret, salt, 24)) as Buffer)
+}
+
+/**
+ * Independent re-implementation of the *current* stored-hash format under an
+ * explicit work factor, so a spec can assert which parameters a hash was
+ * actually derived with.
+ */
+function storedHash(
+  secret: string,
+  salt: string,
+  params: ScryptParams,
+): Promise<string> {
+  return new Promise((resolve, reject) => {
+    scryptCb(secret, salt, 24, params, (err, derivedKey) => {
+      if (err) {
+        reject(err)
+      } else {
+        resolve(base62.encode(derivedKey))
+      }
+    })
+  })
 }
 
 test('rand62 emits only base62 characters within the expected length bound', () => {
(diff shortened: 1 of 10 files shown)


NilsR0711:
Follow-up from this work, out of scope here: #558 (test files leak rate-limit env overrides into each other).

PR #560, opened 2026-07-21
refactor(shared): brand the view-id and cell-index axes

### LINKED ISSUE #507: Type stateIdxMap at construction and separate the viewId / cell-index axes

NilsR0711:
## Summary

`useStreamwallState` built the cell-index map without a type parameter and filled it with `{}` placeholders it then `Object.assign`ed into. Inside the builder every value was therefore `any`, and a cell could in principle hold a partially populated `ViewInfo` — the map is only typed on the `StreamwallConnection` interface, never at the construction site.

That was also the type hole behind #470: the wall's two numeric addressing axes are structurally identical, so passing a **grid cell index** where a **stable view id** was expected type-checked fine.

This PR fixes both:

- **Typed construction.** `stateIdxMap` is now `Map<CellIdx, ViewInfo>`, and every occupied cell stores the very `ViewInfo` object that was pushed to `views` instead of a mutated `{}`. A view spanning several cells is reachable from each of them via the same object, so the map and `views` cannot drift and no entry is ever half-built.
- **Separated axes.** New `packages/streamwall-shared/src/viewAddressing.ts` declares the nominal types `ViewId` and `CellIdx` plus the explicit `asViewId`/`asCellIdx` tagging helpers. Both are plain numbers at runtime; only the compiler sees the difference.

The brands are threaded through the layers where the two axes actually meet:

| Layer | Change |
| --- | --- |
| `schemas.ts` | `viewIdSchema` / `viewIdxSchema` and `ViewPos.spaces` now yield the branded types (identity transform — validation is unchanged) |
| `types.ts` | `ViewState.context.id: ViewId`, `StreamwallState.fullscreenViewIdx: CellIdx \| null` |
| control-UI | `ViewInfo`, `stateIdxMap`, `useGridCommands`, `useControlHotkeys`, `useTileDrag`/`gestures`, `viewPlacement`, `ControlGrid`/`GridControls`/`GridInput` |
| main process | `viewStateMachine` context/input `id`, `StreamWindow`'s view-id methods, `getViewAnchorIdx` returning a `CellIdx` |

`GridControls` is the component where both axes coexist (`idx` = cell, `viewId` = actor) — it is now impossible to swap them there, or to feed a cell index into a `ControlCommand`.

### Architecture decisions

- **Nominal types over a runtime wrapper.** Zero runtime cost, no serialization impact; the axes stay plain numbers on the wire and in Yjs.
- **`viewAddressing.ts` is import-free** so both `schemas.ts` and `geometry.ts` can depend on it without an import cycle.
- **Identity `.transform()` instead of `z.brand()`** so the branded types are declared in one place and shared by the hand-written types and the schemas alike. Parsing behavior, bounds, and error messages are unchanged.
- **Pure cell-geometry helpers keep plain `number`s** (`gridInteractions`, `remapGridAssignments`, `computeBoxRect`): they never see a view id, so branding them would add noise without adding safety. `CellIdx` is assignable to `number`, so callers are unaffected.

## How was this tested?

- New `packages/streamwall-control-ui/src/streamwallState.test.tsx` covers the builder's runtime behavior (multi-cell view resolving to one complete, shared `ViewInfo`; unoccupied cells absent; empty map before the first state) **and** the type contract via `@ts-expect-error` guards that `npm run typecheck` enforces: the map rejects placeholder entries and view-id keys, and neither axis accepts the other or a bare number. Both guards failed before the fix.
- `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm test` (1468 tests across all workspaces), and `npm -w streamwall-control-client run build` pass locally.

## Risks / breaking changes

None at runtime — the change is entirely type-level plus the `stateIdxMap` construction, which produces the same entries as before (`Object.assign` already copied every `ViewInfo` key). Test fixtures across the repo now tag their literals with `asViewId`/`asCellIdx`.

Closes #507

## Checklist


NilsR0711:
Follow-up filed: #567 — extend the `CellIdx` brand to the pure cell-geometry helpers (`gridInteractions`, `useTileResize`, `ResizeHandles`, the `geometry.ts` remap helpers), which this PR deliberately left on plain `number`.

PR #566, opened 2026-07-21
test(control-server): stop test files leaking rate-limit env overrides into each other

### LINKED ISSUE #558: test(control-server): stop test files leaking rate-limit env overrides into each other
## Context

Noticed while working on #536 / #556.

Several control-server suites widen the rate limits for a whole test file by
writing process-wide environment variables and never restoring them:

```ts
process.env.STREAMWALL_RATE_LIMIT_MAX = '10000'
process.env.STREAMWALL_AUTH_RATE_LIMIT_MAX = '10000'
```

This happens in `wsValidation.test.ts`, `multiplexing.test.ts`,
`logging.test.ts` and `sentryReporting.test.ts`, among others (~20 assignments
across the package). Only a handful of variables are ever cleaned up again —
`wsValidation.test.ts` deletes `STREAMWALL_WS_UPDATE_MAX_BYTES` in an `after()`
hook precisely because leaking it would change what a later file sees, and
`rateLimit.test.ts`, `storage.test.ts`, `logger.test.ts` and `sentry.test.ts`
do the same for their own variables. The rate-limit overrides do not.

## Impact

The suite currently runs with `--test-concurrency=1`, so the leak is mostly
invisible today, but:

- a file that relies on the *production* rate limits silently inherits a
  10000-request allowance if it happens to run after one of these files, so a
  regression in the limits could pass unnoticed;
- test outcomes become order-dependent, which is exactly the class of
  flakiness #221 and #492 came from;
- it blocks ever raising `--test-concurrency`, because concurrent files would
  then race on the same process-wide state.

## Suggested solution

Give the overrides the same treatment the injected dependencies already get:
either restore them in an `after()` hook per file, or (preferably) route the
limits through the injected options `initApp` already accepts —
`getRateLimitConfig()` is read once at boot, so `buildTestApp` could take a
`rateLimit` override the same way it takes `scryptParams`, and the env vars
would stop being test API at all.

NilsR0711:
## Summary

Several control-server suites widened the per-IP rate limits for a whole file by
writing `STREAMWALL_RATE_LIMIT_MAX` / `STREAMWALL_AUTH_RATE_LIMIT_MAX` and never
restoring them. With `--test-concurrency=1` the leak is mostly invisible, but it
makes outcomes order-dependent (a file relying on the production limits silently
inherits a 10000-request allowance) and blocks ever raising the concurrency.

This takes the route the issue prefers: the limits become an injected option
rather than test API.

- `initApp` accepts `rateLimit?: Partial<RateLimitConfig>`, merged over what
  `getRateLimitConfig()` reads from the environment, alongside the existing
  `scryptParams` / `updateChecker` / `sentryClient` injections. Unset fields keep
  the environment value, so production behaviour is unchanged.
- `buildTestApp` passes it through and exports `WIDE_RATE_LIMITS`; `logging`,
  `multiplexing`, `sentryReporting`, `uplinkAuth`, `wsRateLimit` and
  `wsValidation` use that instead of touching `process.env`.
- `rateLimit.test.ts` still covers the environment contract, but through a new
  `setEnvForTest()` helper that snapshots and restores the previous values in an
  `after()` hook — including the `STREAMWALL_TRUST_PROXY` and
  `STREAMWALL_WS_MSG_*` overrides that leaked the same way.
- `testEnvHygiene.test.ts` fails the suite if a bare assignment (or `delete`) of
  the rate-limit variables reappears in any spec.

Closes #558

## How was this tested?

- `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm test` — all green.
- Red-first: the injection test fails (`an injected rate limit takes precedence
  over the environment`) when `initApp` ignores the override, and the hygiene
  spec fails against the pre-change test files.
- New coverage: injected override beats the environment; a partial override
  still reads the rest from the environment; `setEnvForTest` applies and then
  restores values.

## Checklist

### DIFF
--- a/packages/streamwall-control-server/src/index.ts
+++ b/packages/streamwall-control-server/src/index.ts
@@ -15,6 +15,7 @@ import {
   getRateLimitConfig,
   getWsMessageLimitConfig,
   parseTrustProxy,
+  type RateLimitConfig,
 } from './config.ts'
 import {
   type AppContext,
@@ -54,6 +55,7 @@ export async function initApp({
   db: injectedDb,
   logLevel,
   logStream,
+  rateLimit: injectedRateLimit,
   sentryEnabled: injectedSentryEnabled,
   sentryClient,
   scryptParams,
@@ -70,6 +72,13 @@ export async function initApp({
   logLevel?: LogLevel
   /** Test-only sink for log output; defaults to pino's stdout destination. */
   logStream?: { write(line: string): void }
+  /**
+   * Overrides individual per-IP rate limits, so specs that are not about
+   * throttling widen them without writing the process-wide environment (which
+   * would leak into whichever file runs next). Unset fields keep the value
+   * `getRateLimitConfig()` reads from the environment.
+   */
+  rateLimit?: Partial<RateLimitConfig>
   /** Test-only override so specs can exercise Sentry-enabled paths without a real DSN. */
   sentryEnabled?: boolean
   /** Test-only override for the client `captureException(...)` reports to. */
@@ -121,7 +130,7 @@ export async function initApp({
     app.log,
   )
 
-  const rateLimitConfig = getRateLimitConfig()
+  const rateLimitConfig = { ...getRateLimitConfig(), ...injectedRateLimit }
 
   const ctx: AppContext = {
     log: app.log,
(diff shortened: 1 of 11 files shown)


NilsR0711:
Follow-ups filed: #568 (route the remaining env overrides through `setEnvForTest`) and the concurrency revisit tracked alongside it.

NilsR0711:
Also filed #569 — revisit `--test-concurrency=1` for the control-server suite now that the rate-limit leaks are gone.

PR #585, opened 2026-07-22
refactor(main): split the extractProgress guard into named predicates

NilsR0711:
Closes the Repowise `complex_conditional` finding at `packages/streamwall/src/main/appUpdater.ts` in `extractProgress` — the repo's highest single health impact (2.17). The guard combined six boolean operators in one `if`, mixing three unrelated concerns: field types, finiteness, and whether the total is usable as a denominator.

### Before

```ts
const { percent, transferred, total } = info as Record<string, unknown>
if (
  typeof percent !== 'number' ||
  typeof transferred !== 'number' ||
  typeof total !== 'number' ||
  !Number.isFinite(percent) ||
  !Number.isFinite(transferred) ||
  !Number.isFinite(total) ||
  total <= 0
) {
  return null
}
return { percent, transferred, total }
```

### After

```ts
const { percent, transferred, total } = info as Record<string, unknown>

// A NaN/Infinity (or entirely absent) field means the backend cannot measure
// this download - e.g. a server that omits Content-Length.
const allFieldsMeasured =
  isFiniteNumber(percent) &&
  isFiniteNumber(transferred) &&
  isFiniteNumber(total)
if (!allFieldsMeasured) {
  return null
}

// A zero or negative total would make the banner's percentage meaningless.
const hasKnownTotal = total > 0
if (!hasKnownTotal) {
  return null
}

return { percent, transferred, total }
```

with

```ts
function isFiniteNumber(value: unknown): value is number {
  return typeof value === 'number' && Number.isFinite(value)
}
```

`isFiniteNumber` is a type guard, so the `typeof` and `Number.isFinite` pair collapses into one predicate per field and the returned object still narrows to `number` without a cast.

### Behaviour-preserving

Identical semantics: the type check still short-circuits before `Number.isFinite`, and `total > 0` is still only evaluated once all three fields are known-finite numbers. No public surface changed — `extractProgress` remains module-private and `AppUpdater`'s API is untouched.

### Tests

Added characterisation tests driving the guard through the `download-progress` event, one case per branch of the original condition:

- non-object payloads (string, `null`, `undefined`), and an object with no fields at all
- a non-numeric `percent` / `transferred` / `total`
- a `NaN` `percent` / `transferred` / `total`
- an infinite `percent` / `transferred` / `total`
- a zero total and a negative total
- the smallest accepted payload (`total: 1`), plus progress events arriving outside the `downloading` state

`appUpdater.test.ts` goes from 26 to 41 tests.

### Verification

- `npm run test` — pass (all workspaces)
- `npm run lint` — pass
- `npm run typecheck` — pass

### DIFF
--- a/packages/streamwall/src/main/appUpdater.test.ts
+++ b/packages/streamwall/src/main/appUpdater.test.ts
@@ -125,6 +125,85 @@ describe('AppUpdater status', () => {
     })
   })
 
+  // One case per branch of the progress guard, so the named predicates it was
+  // split into keep rejecting exactly what the original condition rejected.
+  it.each([
+    ['a non-object payload', 'not-an-object'],
+    ['a null payload', null],
+    ['an undefined payload', undefined],
+    ['no fields at all', {}],
+    ['a non-numeric percent', { percent: '40', transferred: 400, total: 1000 }],
+    [
+      'a non-numeric transferred',
+      { percent: 40, transferred: '400', total: 1000 },
+    ],
+    ['a non-numeric total', { percent: 40, transferred: 400, total: '1000' }],
+    ['a NaN percent', { percent: Number.NaN, transferred: 400, total: 1000 }],
+    [
+      'a NaN transferred',
+      { percent: 40, transferred: Number.NaN, total: 1000 },
+    ],
+    ['a NaN total', { percent: 40, transferred: 400, total: Number.NaN }],
+    [
+      'an infinite percent',
+      { percent: Number.POSITIVE_INFINITY, transferred: 400, total: 1000 },
+    ],
+    [
+      'an infinite transferred',
+      { percent: 40, transferred: Number.POSITIVE_INFINITY, total: 1000 },
+    ],
+    [
+      'an infinite total',
+      { percent: 40, transferred: 400, total: Number.POSITIVE_INFINITY },
+    ],
+    ['a zero total', { percent: 0, transferred: 0, total: 0 }],
+    ['a negative total', { percent: 0, transferred: 0, total: -1 }],
+  ])('keeps the progress indeterminate for %s', (_label, payload) => {
+    const { backend, updater } = createUpdater()
+    backend.emit('update-available', { version: '0.9.2' })
+    updater.download()
+
+    backend.emit('download-progress', payload)
+
+    expect(updater.getStatus()).toEqual({
+      state: 'downloading',
+      version: '0.9.2',
+      progress: null,
+    })
+  })
+
+  it('accepts a fully measured payload down to a single-byte total', () => {
+    const { backend, updater } = createUpdater()
+    backend.emit('update-available', { version: '0.9.2' })
+    updater.download()
+
+    backend.emit('download-progress', { percent: 0, transferred: 0, total: 1 })
+
+    expect(updater.getStatus()).toEqual({
+      state: 'downloading',
+      version: '0.9.2',
+      progress: { percent: 0, transferred: 0, total: 1 },
+    })
+  })
+
+  it('ignores progress events outside the downloading state', () => {
+    const { backend, updater } = createUpdater()
+    backend.emit('update-available', { version: '0.9.2' })
+
+    backend.emit('download-progress', {
+      percent: 40,
+      transferred: 400,
+      total: 1000,
+    })
+
+    expect(updater.getStatus()).toEqual({
+      state: 'available',
+      version: '0.9.2',
+      releaseUrl,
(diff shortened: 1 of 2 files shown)


NilsR0711:
**Verdict: no blocking findings — the behaviour-preserving claim holds, and I could not find an input where old and new disagree.**

### What I verified

**Equivalence.** Reconstructed the pre-change condition from `git show origin/main:packages/streamwall/src/main/appUpdater.ts` and ran both versions side by side over a generated input matrix: 28 values per field (`0`, `-0`, `±1`, `±0.5`, `±Number.MIN_VALUE`, `5e-324`, `±Number.MAX_VALUE`, `NaN`, `±Infinity`, `'1000'`, `'0'`, `''`, `true`, `false`, `null`, `undefined`, `{}`, `[]`, `[1]`, `10n`, `new Number(5)`, a function, `Object.create(null)`) across all three fields, plus non-object top-level payloads (`'x'`, `0`, `null`, `undefined`, `Symbol`, `BigInt`, `Map`, a getter-backed object). 21,966 inputs, 0 divergences, compared with `Object.is` so `-0` vs `0` would have shown up.

That matches the manual argument: `isFiniteNumber(x) === (typeof x === 'number' && Number.isFinite(x))` per field, and `Number.isFinite` on a non-number is `false` anyway, so interleaving the type and finiteness checks per field instead of doing all three `typeof`s first cannot change the result. The `&&` chain still short-circuits, `total > 0` is still only reached once all three are known-finite numbers, and `!(total > 0)` is exactly `total <= 0` for a finite number. All operands are already-destructured locals, so no getter runs twice and no evaluation-order side effect exists.

**Boundaries specifically checked:** `total: 0`, `total: -0`, `total: -1`, `total: 5e-324`, `total: 1` (the smallest integer accepted), `NaN`/`±Infinity` in each of the three positions, numeric strings, and a missing field (`undefined`).

**Type narrowing.** No `any`, no `as`-cast added, no non-null assertion. `isFiniteNumber` is a real `value is number` predicate, and the return object narrows through the aliased `const allFieldsMeasured`. `npm run typecheck` passes, so the narrowing is genuine rather than papered over. The only cast in the function, `info as Record<string, unknown>`, is pre-existing and untouched.

**Test quality.** I mutation-tested the added cases rather than trusting that they pass:

- `total > 0` → `total >= 0`: fails *keeps the progress indeterminate for a zero total*. Boundary is pinned.
- `isFiniteNumber` → `typeof value === 'number'` only: fails 5 cases (NaN percent/transferred, infinite percent/transferred/total). Finiteness is pinned.
- `isFiniteNumber` → `Number.isFinite(Number(value))`: fails all 3 non-numeric cases. The `typeof` half is pinned.

No pre-existing assertion is weakened or deleted — the test diff is 79 added lines and 0 removed.

**Scope / conventions.** Diff touches only `appUpdater.ts` and `appUpdater.test.ts`. Conventional Commit title, single commit, English throughout, no attribution footer.

### Findings

**Non-blocking, test design.** The 15 `it.each` rejection cases assert `progress: null`, which is also the state `download()` leaves behind before any event arrives. So each of them individually cannot distinguish "the guard rejected the payload" from "the handler never ran at all". I confirmed this: bypassing `extractProgress` entirely in the `download-progress` listener (forcing `progress: null`) leaves all 15 rejection tests green — only the two accept-side tests catch it. The suite as a whole is still sound because those accept-side tests exist, but the rejection cases would be strictly stronger if each emitted a valid progress event first and then the bad payload, asserting that progress falls *back* to `null`. That also documents the actual user-visible behaviour (a malformed event mid-download resets the banner to indeterminate), which nothing currently pins.

**Nit.** `'a NaN total'` does not independently exercise the finiteness check for `total` — `NaN > 0` is `false`, so `hasKnownTotal` rejects it regardless. It confirmed this in the mutation run above (it was the one NaN/Infinity case that stayed green when finiteness was removed). Harmless as a characterisation case, just not load-bearing.

### Worth a follow-up

`hasKnownTotal`'s narrowing depends on TypeScript's aliased-condition analysis over a `const` holding a direct `&&` chain of type guards. Extracting `allFieldsMeasured` into a helper function or switching it to `let` would silently drop the narrowing and force a cast back in. Not a problem today, just a sharp edge for whoever touches this next.

### Local verification

Ran on the PR head (`973c615`) in a clean checkout with a fresh `npm install`:

- `npm run typecheck` — pass (all 6 workspaces)
- `npm run lint` — pass
- `npm run test` — pass; `appUpdater.test.ts` 41/41, no failures anywhere in the monorepo

PR #588, opened 2026-07-22
test(control-server): move the shared websocket setup into testHelpers

NilsR0711:
## What

`multiplexing.test.ts` and `wsValidation.test.ts` each carried their own copy of the same live-server WebSocket setup: booting the app with wide rate limits and a log capture, listening on a random port, connecting and seeding a Streamwall uplink, redeeming an invite for a client socket, plus identical message-narrowing predicates.

This moves that setup into the package's existing `testHelpers.ts`:

- `startTestServer(overrides)` — boots a listening app with `WIDE_RATE_LIMITS` and a log capture, registers the `app.close()` cleanup, and returns a `connectClient(role)` factory that redeems a fresh invite and opens an authenticated `/client/ws` socket.
- `bootServerWithUplink({ stateMessage, ...overrides })` — `startTestServer` plus a connected Streamwall uplink seeded with a state message (`VALID_STATE` by default). It waits for the server to either accept the snapshot or log a structured rejection, so both specs' seeding semantics are covered without a sleep.
- `TEST_BASE_URL`, the extracted `TestAppOverrides` type, and the `isResponseTo` / `isCommandType` / `isBareError` / `isStateMessage` / `isStateDelta` predicates.

Both specs now consist of their test cases plus a small file-local wrapper where they need one.

## Repowise findings closed

- `packages/streamwall-control-server/src/multiplexing.test.ts` — 63% of the file duplicated; worst clone shared 26 lines with `wsValidation.test.ts` (co-changed 5x).
- `packages/streamwall-control-server/src/wsValidation.test.ts` — 40% duplicated (same clone pair).

## Notes

- No test case, assertion or test name was changed, added or removed.
- The package still runs without `--test-concurrency=1`; the `src/testConcurrency.test.ts` guard is untouched and passing.

## Verification

- `npm run test` — pass (control-server: 183 tests, 0 fail)
- `npm run lint` — pass
- `npm run typecheck` — pass

### DIFF
--- a/packages/streamwall-control-server/src/multiplexing.test.ts
+++ b/packages/streamwall-control-server/src/multiplexing.test.ts
@@ -6,89 +6,26 @@ import WebSocket from 'ws'
 import type { Delta } from 'jsondiffpatch'
 import {
   stateDiff,
-  type ClientCommandResponse,
   type ClientStateDeltaMessage,
-  type ClientStateMessage,
-  type ControlCommandMessage,
-  type ServerToClientMessage,
-  type StreamwallRole,
   type StreamwallState,
 } from 'streamwall-shared'
 import {
-  buildTestApp,
-  captureLogs,
-  connectStreamwallUplink,
-  listenTestApp,
+  bootServerWithUplink,
+  isCommandType,
+  isResponseTo,
+  isStateDelta,
+  isStateMessage,
   messageCollector,
   mintUplinkToken,
   redeemInviteAndConnectClient,
+  startTestServer,
+  TEST_BASE_URL,
   VALID_STATE,
-  WIDE_RATE_LIMITS,
 } from './testHelpers.ts'
-
-const BASE_URL = 'http://localhost:3000'
-
-/** Narrows to the server's reply to the client command with the given `id`. */
-function isResponseTo(id: number) {
-  return (m: ServerToClientMessage): m is ClientCommandResponse =>
-    'response' in m && m.response === true && m.id === id
-}
-
-/** Narrows to a forwarded control command of the given `type`. */
-function isCommandType<Type extends ControlCommandMessage['type']>(type: Type) {
-  return (
-    m: ControlCommandMessage,
-  ): m is Extract<ControlCommandMessage, { type: Type }> => m.type === type
-}
-
-const isStateMessage = (m: ServerToClientMessage): m is ClientStateMessage =>
-  'type' in m && m.type === 'state'
-
-const isStateDelta = (m: ServerToClientMessage): m is ClientStateDeltaMessage =>
-  'type' in m && m.type === 'state-delta'
 
 /** Applies a `state-delta`'s patch to a client's last-known state, as a real client would. */
 function applyDelta(state: StreamwallState, delta: Delta): StreamwallState {
   return stateDiff.patch(structuredClone(state), delta) as StreamwallState
-}
-
-/**
- * Boots a live server, connects a Streamwall uplink and seeds it with
- * `VALID_STATE`, and returns a `connectClient` helper that redeems a
- * freshly-minted invite for the given role and opens an authenticated
- * `/client/ws` socket.
- */
-async function bootServerWithUplink() {
-  const logs = captureLogs()
-  const { app, auth } = await buildTestApp({
-    baseURL: BASE_URL,
-    logs,
-    rateLimit: WIDE_RATE_LIMITS,
-  })
-  after(() => app.close())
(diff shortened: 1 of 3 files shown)


NilsR0711:
Reviewed independently — no blocking findings; the "no test case, assertion or test name changed" claim holds up under verification.

## What I verified

**Test-name parity.** Extracted the `test(...)` title list from `origin/main` and from the branch head for both spec files: identical, in identical order — 12 in `multiplexing.test.ts`, 10 in `wsValidation.test.ts`. No adds, drops or renames.

**Test bodies.** For `wsValidation.test.ts` the region from the first `test(` to EOF is byte-identical to `origin/main` (same md5) — the entire diff there is confined to the imports and the file-local `connectStreamwallAndClient` wrapper. `multiplexing.test.ts` has exactly three hunks: the import/helper header, `startUplinkServer` (now delegating to `startTestServer`, which performs the same `captureLogs` → `buildTestApp(WIDE_RATE_LIMITS)` → `after(app.close())` → `listenTestApp` sequence), and one `BASE_URL` → `TEST_BASE_URL` substitution where both resolve to `http://localhost:3000`.

**The OR-wait in `bootServerWithUplink` (the part I was most suspicious of).** Three checks:

1. `Streamwall connected` is logged in `ws/uplink.ts` only inside the `clientState === null` branch, i.e. only after `controlStateMessageSchema` *and* `streamwallStateSchema` have both accepted the payload. It is not a connect-time log, so the wait is a genuine "the seed was accepted" signal, not a socket-open signal.
2. `LogCapture.waitFor` scans `entries` before registering a waiter, so there is no lost-wakeup race if the log lands before the await, and it rejects after 2 s rather than hanging.
3. Empirically: I patched every `bootServerWithUplink()` call in `multiplexing.test.ts` to seed a payload the server rejects, so the wait resolves on the *wrong* branch. Result: 11 of 11 affected tests fail. Zero pass vacuously. Restored afterwards.

**Isolation / concurrency.** The new exports are pure functions plus one `const` string; no module-level mutable state was introduced, and nothing writes `process.env`. `docUpdateLimits` stays injected per-app rather than env-set, as the original comment demanded. `package.json` and `src/testConcurrency.test.ts` are untouched — the suite still runs without `--test-concurrency=1`. Cleanup is still `after(() => app.close())` registered inside the helper, which is the same file-scoped (not per-test) semantics as before — unchanged, not worsened.

**Flakiness.** No new sleeps; the diff removes none of the existing awaits and adds no unawaited promise. Ran the control-server suite 5 consecutive times: 183/183 pass each time, 3.7–4.9 s.

**Scope and conventions.** Diff touches only `packages/streamwall-control-server/`. Commit title is a valid Conventional Commit; commit body and PR body are English and carry no attribution footer.

## Local results

- `npm run test` — pass (control-server 183/183, control-ui 363/363, all other workspaces green)
- `npm run lint` — pass
- `npm run typecheck` — pass
- `npx prettier --check` on the touched files — clean
- control-server suite x5 — pass, no flake

## Findings

No blocking findings.

Non-blocking:

1. **The OR-wait loosens the failure mode for `multiplexing.test.ts`.** On `main` those tests waited strictly for `Streamwall connected`; a rejected seed produced one clear "timed out waiting for matching log entry". Now it also resolves on `Rejected invalid Streamwall state*`, so the same regression surfaces as a scatter of downstream assertion failures instead. Still loud (proven above), just noisier to diagnose. Consider making the tolerant branch opt-in — e.g. only accept the rejection log when the caller passed an explicit `stateMessage`, or take an `expectSeedRejection` flag — so the default path keeps the sharp signal. Your call; not worth blocking on.

2. **`startTestServer` replaces `rateLimit` wholesale rather than merging.** `...overrides` is spread after `rateLimit: WIDE_RATE_LIMITS`, so a caller passing a partial `rateLimit` silently loses the wide defaults. No current caller does this, so it is latent only — worth a one-line doc note if you expect callers to tune a single limit.

3. **Minor typing looseness in the `wsValidation` wrapper.** `stateMessage` is still declared in the wrapper's parameter type but is no longer destructured; it reaches `bootServerWithUplink` via `...overrides`, which works but relies on the reader noticing. Dropping it from the wrapper's own type annotation and letting the rest type flow through would make the pass-through explicit.

## Follow-ups

None that need an issue. The `after(() => app.close())`-inside-a-helper pattern means every server booted in a file stays alive until the file's hooks run rather than being torn down per test — pre-existing, unchanged by this PR, and harmless at the current suite size, but it is the thing to revisit first if the control-server suite ever starts exhausting sockets on CI.

PR #589, opened 2026-07-22
test(control-ui): extract the shared render setup out of the duplicated specs

NilsR0711:
Closes the Repowise DRY findings in the control-UI test suite.

## Findings addressed

| File | Finding |
|------|---------|
| `packages/streamwall-control-ui/src/hotkeysOptions.test.tsx` | 100% duplicated, worst clone 41 lines shared with `transientProps.test.tsx` |
| `packages/streamwall-control-ui/src/transientProps.test.tsx` | 89% duplicated (same clone pair) |
| `packages/streamwall-control-ui/src/gridFullscreen.test.tsx` ↔ `packages/streamwall-control-ui/src/GridViewIdentity.test.tsx` | worst clone 70 shared lines |

## What changed

- New `src/testHelpers.tsx` holds the setup every ControlUI spec had its own copy of:
  - the container mount/unmount lifecycle (`renderControlUI`, `rerenderControlUI`, plus the `afterEach` teardown that unmounts and detaches the container),
  - the fake data builders `makeConnection`, `makeDelayState`, `makeStream`, `makeStreamWindowConfig` and `makeView`.
- New `src/testIconStubs.tsx` holds the `react-icons` stubs. It is deliberately import-free, because the specs pull it in from inside their `vi.mock('react-icons/...')` factories — importing the shared helper there would drag the component tree (and with it `react-icons`) into the factory.
- The four specs above now build on those helpers. Specs keep thin local wrappers where their fixtures differ in intent (e.g. `gridFullscreen` deliberately uses view ids offset from the cell index, per #397).

No test case was removed, renamed or weakened; all assertions are byte-for-byte the same as before.

## Known remaining item

`packages/streamwall-control-ui/src/GridPreviewBox.test.tsx` is reported as 85% duplicated, with its worst clone (24 lines) shared with `packages/streamwall/src/renderer/OverlayViewTile.test.tsx`. That clone crosses a package boundary, so deduping it would mean introducing a cross-package test-helper package. That is out of scope here and left as a follow-up.

## Verification

- `npm run test` — all packages pass (control-ui: 48 files, 363 tests)
- `npm run lint` — clean
- `npm run typecheck` — clean

### DIFF
--- a/packages/streamwall-control-ui/src/GridViewIdentity.test.tsx
+++ b/packages/streamwall-control-ui/src/GridViewIdentity.test.tsx
@@ -1,150 +1,51 @@
-import { render } from 'preact'
-import { act } from 'preact/test-utils'
+import { asCellIdx, type CellIdx } from 'streamwall-shared'
+import { describe, expect, test, vi } from 'vitest'
+import type { StreamwallConnection, ViewInfo } from './index.tsx'
 import {
-  asCellIdx,
-  asViewId,
-  type CellIdx,
-  type StreamData,
-  type StreamDelayStatus,
-  type StreamWindowConfig,
-} from 'streamwall-shared'
-import { afterEach, describe, expect, test, vi } from 'vitest'
-import * as Y from 'yjs'
-import {
-  ControlUI,
-  type StreamwallConnection,
-  type ViewInfo,
-} from './index.tsx'
+  makeConnection,
+  makeStream,
+  makeStreamWindowConfig,
+  makeView,
+  renderControlUI,
+  rerenderControlUI,
+} from './testHelpers.tsx'
 
-vi.mock('react-icons/fa', () => ({
-  FaExchangeAlt: () => null,
-  FaExclamationTriangle: () => null,
-  FaRedoAlt: () => null,
-  FaRegLifeRing: () => null,
-  FaRegWindowMaximize: () => null,
-  FaSyncAlt: () => null,
-  FaVideoSlash: () => null,
-  FaVolumeUp: () => null,
-}))
-vi.mock('react-icons/md', () => ({
-  MdOutlineStayCurrentLandscape: () => null,
-  MdOutlineStayCurrentPortrait: () => null,
-}))
+vi.mock(
+  'react-icons/fa',
+  async () => (await import('./testIconStubs.tsx')).faIconStubs,
+)
+vi.mock(
+  'react-icons/md',
+  async () => (await import('./testIconStubs.tsx')).mdIconStubs,
+)
 vi.mock('react-hotkeys-hook', () => ({
   useHotkeys: () => {},
 }))
 
-let container: HTMLDivElement | undefined
-
-afterEach(() => {
-  if (container) {
-    act(() => render(null, container!))
-    container.remove()
-    container = undefined
-  }
-})
-
-function makeStream(id: string): StreamData {
-  return {
-    _id: id,
-    _dataSource: 'test',
-    kind: 'video',
-    link: `https://example.com/${id}`,
-  }
-}
-
-function makeView(streamIdx: number, cells: number[]): ViewInfo {
-  const spaces = cells.map(asCellIdx)
-  return {
-    state: {
-      state: {
-        displaying: {
-          running: {
-            playback: 'playing',
-            video: 'normal',
-            audio: 'muted',
-            pause: 'unpaused',
-            swap: 'idle',
-          },
-        },
-      },
-      context: {
-        id: asViewId(streamIdx),
-        content: { url: `https://example.com/s${streamIdx}`, kind: 'video' },
-        info: null,
-        pos: { x: spaces[0] * 100, y: 0, width: 100, height: 100, spaces },
-        error: null,
-        volume: 1,
-      },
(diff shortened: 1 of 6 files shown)


NilsR0711:
**Verdict: no blocking findings.** The "nothing removed, renamed or weakened" claim holds up under checking, including on the view-id axis.

## What I verified

**Test-name sets.** Extracted every `describe`/`test` title from `git show origin/main:<path>` and from the branch for all four specs and diffed them: identical, no additions, no removals, no renames — GridViewIdentity 2, gridFullscreen 4, hotkeysOptions 15, transientProps 2. Assertion bodies are unchanged line-for-line in all four.

**Fixture comparison (old inline literal vs. shared builder).**
- `hotkeysOptions` / `transientProps`: the old inline `StreamwallConnection` literal and the old inline `delayState` match `makeConnection()` / `makeDelayState()` field for field, including `config: undefined` and `sharedState: { views: {} }`. Same input.
- `gridFullscreen`: `baseConnection` still overrides `delayState: undefined` (it did on `main` too), so the new non-`undefined` `makeDelayState()` default does not silently leak in.
- `GridViewIdentity`: the old config was `{cols: viewCount, rows: 1, width: 100*viewCount, height: 100, ...}`; `makeStreamWindowConfig({cols, width})` reproduces it exactly. One defaults difference exists and is benign: the old local `makeView` hardcoded `pos.width: 100`, the shared one uses `100 * spaces.length`. Identical for the single-cell views this spec builds, and it matches what `gridFullscreen` used on `main`, which is the only spec with multi-cell views.
- Behavioural detail in `GridViewIdentity`: the re-render used to happen inside a single `act()` that also constructed the connection; it now constructs outside and `act`s only around `render`. No semantic difference — fixture construction has no scheduler side effects.

**viewId vs. cell index (the risk item).** Not collapsed. `makeView` takes `id` and `cells` as two separate inputs and is documented as such, and `gridFullscreen` keeps a local `makeFullscreenView` that offsets by `1000 + cells[0]`, with the #397 rationale comment preserved. The `expect(sent).toContainEqual({... viewId: 1001 ...})` / `viewId: 1000` assertions are unchanged, so both tests still fail if the component starts sending the cell index. Old code computed `1000 + spaces[0]`, new computes `1000 + cells[0]` — the same number, since `asCellIdx` is a type-level brand.

**Teardown.** I mounted a throwaway probe spec against `testHelpers.tsx`. The module-scope `afterEach` does register on the importing spec's root suite and does really unmount (`render(null, container)`, not just `remove()`): after a test that mounts, `document.body.childElementCount` is back to 0. Not a leak in current usage.

**vi.mock / testIconStubs.** The claimed cycle is real, not defensive folklore — I ran a probe whose `react-icons/fa` factory pulls in `testHelpers.tsx` (which imports `index.tsx`, which imports `react-icons`) and the run hung until I killed it at 5 minutes. Keeping the stubs in an import-free module is the right call. The factories are hoist-safe: they reference no outer bindings and use `await import(...)` inside an async factory, which is the documented vitest pattern.

**Local runs on the branch head:** `npm run test` green (control-ui 48 files / 363 tests; control-server 183; all packages pass), `npm run lint` clean, `npm run typecheck` clean. The diff is confined to `packages/streamwall-control-ui/`. Commit title is a Conventional Commit; commit body and PR body are English.

## Findings (non-blocking)

1. `renderControlUI` tracks only one `container` in a module-scope `let`. Probe result: two `renderControlUI` calls in a single test leak the first container and its Preact root permanently (`document.body.childElementCount` stays at 1 after teardown). No current spec does this — but the point of a shared helper is that the next spec author will not read its implementation first. A `const containers: HTMLDivElement[] = []` plus a loop in the `afterEach` removes the trap for one line of code.
2. Partial migration. Roughly a dozen further specs still carry their own copy of the same icon stub literals (`responsiveLayout`, `serverStatusGating`, `disconnectRendering`, `connectionStatus`, `gridDragMoveRoleGating`, `swapHotkeyRoleGating`, `GridControls`, `StreamList`, `CustomStreamInput`, `preactCompatInterop`, `gridTestHooks`, …), and most of them also hand-roll the container lifecycle. Not a defect in this PR, but the duplication finding is only partly closed, and two idioms living side by side is its own small cost. Worth a follow-up issue covering the remaining specs.
3. Naming: `testHelpers.tsx` and `testIconStubs.tsx` sit in `src/` without a `.test.` marker while importing `vitest` at module scope. Harmless today (the package is private and nothing in the component tree imports them), but a `*.testkit.tsx` suffix or a `src/testing/` folder would make it structurally impossible for app code to pull `afterEach` into a production bundle.

## On over-abstraction

The split lands on the right side of the line. `makeConnection` / `makeStreamWindowConfig` take `Partial<>` overrides, so each spec's non-default fields — `fullscreenViewIdx`, the `sharedState.views` mapping, `send` — stay visible at the call site, which is where the reader looks. `makeView`'s explicit `{ id, contentUrl, cells }` reads better than the old 40-line literal, not worse. The one default doing load-bearing work invisibly is `delayState` in `makeConnection`; `gridFullscreen` correctly overrides it, and a one-line comment on the field saying which specs depend on the default would be worth adding.

Issue #582, opened 2026-07-22
Split main/data.ts into polling, file watching and parsing modules


NilsR0711:
Narrowed after the I/O half was resolved. Original scope was a Repowise report on `packages/streamwall/src/main/data.ts` (defect score 2.35/10, the worst file in the repository) covering two things: N+1 I/O findings, and top-3% change entropy.

### Resolved — the N+1 findings were false positives

PR #593 examined both flagged sites and left them serial, deliberately:

- `pollDataURL` is a paced interval poller over a **single** endpoint. Iteration N is the next refresh cycle of the same URL and must not start before N-1 finished. Parallelising it would turn a paced poller into a request flood.
- `watchDataFile` is event-driven over a **single** path. Reading ahead would only re-read the same file.
- The genuine fan-out already exists one level up: `buildDataSources` creates one generator per source and `combineDataSources` advances them concurrently.

Both loops now carry comments recording why the `await` stays inside them, plus tests pinning that a single URL is never polled concurrently while independent URLs are. See #593 for the mutation-tested verification.

Separately, #596 / PR #605 replaced `Repeater.latest` in `combineDataSources` so a silent source no longer withholds every other source's data, and #603 tracks the remaining `fetch`-timeout gap.

### Remaining — the structural half

Change entropy in the top 3% means changes to this file are scattered across many unrelated commits. That is a symptom of the file carrying several independent responsibilities at once:

- source polling (`pollDataURL`)
- file watching (`watchDataFile`)
- parsing and normalising the fetched payloads
- the combinator (`combineDataSources` / `latestPerSource`)

Splitting these into separate modules would mean a change to, say, the TOML parsing no longer touches the same file as a change to the poll cadence.

### Acceptance criteria

- [ ] `data.ts` split along the responsibility lines above, with the public surface unchanged for its consumers
- [ ] Existing tests keep passing without weakening; test files split alongside the modules
- [ ] No behaviour change — this is structural only
- [ ] `npm run test`, `npm run lint`, `npm run typecheck` green

Not urgent. Worth doing the next time this file needs a real change anyway, rather than as a standalone refactor.

PR #593, opened 2026-07-22
test(streamwall): pin data source I/O concurrency and error semantics

NilsR0711:
Audits the two N+1 I/O findings reported against `packages/streamwall/src/main/data.ts` in #582 and pins the resulting behaviour with tests.

## Verdict per site

**`pollDataURL` (~line 36) — deliberately left serial.**
The `await fetch(url)` is inside a polling *interval* loop over a single endpoint, not a fan-out over independent items. Iteration N is the next refresh cycle of the same URL and must not start before iteration N-1 has finished and the interval has elapsed. Parallelising it would turn a paced poller into an unbounded request flood against the operator's endpoint.

**`watchDataFile` (~line 85) — deliberately left serial.**
The `await fsPromises.readFile(path)` is inside an event-driven loop over a single file. Each iteration is triggered by a filesystem event for that one path, so there is nothing independent to fan out over — reading ahead would only re-read the same file.

**Concurrency across sources already exists, one level up.**
`buildDataSources` creates one generator per URL / per TOML file, and `combineDataSources` advances all of them at once via `Repeater.latest`. N URLs are therefore already polled in parallel with each other, while each individual URL stays strictly serial. There is no unbounded fan-out to cap: the number of sources is bounded by the `--data.json-url` / `--data.toml-file` / `--presets` flags, and each source is a single long-lived generator rather than a per-iteration burst of requests.

Since the reported serial awaits are correct as written, no production behaviour is changed here.

## What this PR does

- Documents both loops in `data.ts` with the reasoning above, so the static finding is not "fixed" into a regression later.
- Adds tests in `data.test.ts`:
  - `pollDataURL` never issues overlapping requests for a single URL (peak in-flight request count stays at 1 across several poll cycles).
  - Independent URLs are polled concurrently rather than one after another — the test server withholds every response until two requests have arrived, so a value can only be produced if both polls are in flight simultaneously.
  - Error semantics of the fan-out: a source whose fetch always fails does not suppress or delay data from a healthy source.

## Remaining work

The issue's second suggestion — splitting source polling, file watching and parsing into separate modules to address the change-entropy signal — is a larger structural change and is intentionally **not** part of this PR. It should be tracked separately.

## Verification

- `npm run test` — pass
- `npm run lint` — pass
- `npm run typecheck` — pass

Refs #582 (the I/O findings are addressed as false positives; the module split remains open, so the issue is not auto-closed).

### DIFF
--- a/packages/streamwall/src/main/data.test.ts
+++ b/packages/streamwall/src/main/data.test.ts
@@ -437,6 +437,148 @@ describe('pollDataURL', () => {
       await gen.return(undefined)
     }
   })
+
+  // The serial `await fetch()` in pollDataURL's loop is deliberate: the loop
+  // is a polling interval over one endpoint, so a request must never be
+  // issued while the previous one for the same URL is still in flight.
+  test('never issues overlapping requests for a single URL', async () => {
+    let inFlight = 0
+    let maxInFlight = 0
+    let counter = 0
+    server = createServer((_req, res) => {
+      inFlight++
+      maxInFlight = Math.max(maxInFlight, inFlight)
+      setTimeout(() => {
+        inFlight--
+        res.setHeader('content-type', 'application/json')
+        res.end(
+          JSON.stringify([
+            { link: `https://a.example/${counter++}`, kind: 'video' },
+          ]),
+        )
+      }, 20)
+    })
+    await new Promise<void>((resolve) =>
+      server!.listen(0, '127.0.0.1', resolve),
+    )
+    const { port } = server.address() as AddressInfo
+    const url = `http://127.0.0.1:${port}/`
+
+    const gen = pollDataURL(url, 0.01)
+    try {
+      for (let i = 0; i < 3; i++) {
+        await gen.next()
+      }
+      expect(maxInFlight).toBe(1)
+    } finally {
+      await gen.return(undefined)
+    }
+  })
+})
+
+describe('data source fan-out', () => {
+  let server: Server | undefined
+
+  afterEach(() => {
+    server?.close()
+    server = undefined
+  })
+
+  // Concurrency across sources lives one level up from the per-source loops:
+  // combineDataSources advances every source at once, so N URLs are polled in
+  // parallel even though each individual URL is polled serially.
+  test('polls independent URLs concurrently rather than one after another', async () => {
+    let inFlight = 0
+    let maxInFlight = 0
+    let pendingResponses: Array<() => void> = []
+
+    const flush = () => {
+      const responses = pendingResponses
+      pendingResponses = []
+      for (const respond of responses) {
+        respond()
+      }
+    }
+
+    server = createServer((req, res) => {
+      inFlight++
+      maxInFlight = Math.max(maxInFlight, inFlight)
+      pendingResponses.push(() => {
+        inFlight--
+        res.setHeader('content-type', 'application/json')
+        res.end(
+          JSON.stringify([
+            { link: `https://example.invalid${req.url}`, kind: 'video' },
+          ]),
+        )
+      })
+      if (pendingResponses.length >= 2) {
+        flush()
+      } else {
+        // Safety valve so a serial implementation fails the assertion below
+        // instead of hanging the test forever.
+        setTimeout(flush, 500)
+      }
+    })
+    await new Promise<void>((resolve) =>
+      server!.listen(0, '127.0.0.1', resolve),
+    )
+    const { port } = server.address() as AddressInfo
+    const base = `http://127.0.0.1:${port}/`
+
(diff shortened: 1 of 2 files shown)


NilsR0711:
**Verdict: non-blocking — the false-positive claim survives. I tried to refute it and could not; all three tests bite under mutation.**

## Could the "both findings are false positives" claim be refuted?

No, on all three sub-claims:

- **`pollDataURL` is a paced interval poller over one URL.** The deciding line is the signature itself — `pollDataURL(url: string, intervalSecs: number, ...)` takes a single `url`, and the loop body ends in `await sleep(refreshInterval)`. There is no collection to iterate; iteration N is refresh cycle N of the same endpoint. `Promise.all` here would multiply requests against one endpoint, not shorten a fan-out.
- **`watchDataFile` is event-driven over one path.** Same shape: a single `path`, and the bottom of the loop is `await Promise.race([once(watcher, 'all'), stop])`. The next `readFile` is gated on a filesystem event for that one file, so there is nothing to read ahead — a second concurrent read would return the same bytes.
- **The upper-level fan-out exists and really is concurrent.** `buildDataSources` (`main/dataSources.ts`) maps one `pollDataURL`/`watchDataFile` generator per URL / per path, and `combineDataSources` drives them through `Repeater.latest`, which keeps one speculative `.next()` in flight per contender. I verified this empirically rather than by reading: see mutation 2 below. It is not a sequential drain.

## Mutation testing of the three new tests

| Mutation | Result |
|---|---|
| `pollDataURL`: `const resp = await fetch(url)` → `const [resp] = await Promise.all([fetch(url), fetch(url)])` (i.e. "parallelise the loop" as the issue suggests) | **red** — "never issues overlapping requests for a single URL": `expected 2 to be 1` |
| `combineDataSources`: replace `Repeater.latest(dataSources)` with a hand-written sequential drain (`for (const src of dataSources) await src.next()`) | **red** — "polls independent URLs concurrently": `expected 1 to be 2` (plus the pre-existing `return()` teardown test times out, which is a second, independent kill) |
| `pollDataURL`: rethrow from the `catch` so a failing source propagates its error | **red** — "keeps delivering data from healthy sources when one source fails" |
| Bonus: `pollDataURL` catch → `await sleep(10000)` (source that *hangs* instead of failing fast) | **red** by timeout — see the finding below |

No test stayed green under its own mutation.

## Comment accuracy

Both added comments match the code as written; I found nothing incorrect in them. Two calibration notes:

1. **PR body, not the code:** "the number of sources is bounded by the `--data.json-url` / `--data.toml-file` / `--presets` flags" is slightly imprecise — `cliArgs.ts` layers `config.toml` from the user data dir and any `--config` file over the CLI via yargs `.config()`, and `data.json-url` is an array option, so a config file can supply the list too. Still operator-authored and fixed at startup, so the "no cap needed" conclusion holds; only the wording is narrower than reality.
2. The `pollDataURL` comment asserts `combineDataSources` "advances all of them at once". True for *starting* I/O, but see below for what that does **not** imply.

## Findings

**Non-blocking — the "does not delay" claim is broader than the evidence (filed as #596).** `Repeater.latest` needs one value from *every* contender before it can emit, so `combineDataSources` withholds all stream data until the slowest source has produced its first batch. A probe with one healthy source plus one never-yielding source never resolves `combined.next()`. The new test passes only because `pollDataURL` fails *fast* (connection refused → empty batch immediately); a `--data.json-url` that black-holes the connection has no `fetch` timeout and would stall the wall indefinitely. Pre-existing, not introduced here, so it is not a blocker for this PR — filed as #596 with the repro and two suggested directions.

**Nit — possible CI-flake vector in the concurrency test.** The safety valve `setTimeout(flush, 500)` assumes both polls arrive within 500 ms of each other. On a heavily loaded CI runner that could flush early and fail `expect(maxInFlight).toBe(2)` for scheduling reasons rather than for a real regression. Consider `expect(maxInFlight).toBeGreaterThanOrEqual(2)` plus a longer valve, or drop the valve and rely on the test timeout. I ran `data.test.ts` 8× and the full `streamwall` suite (831 tests) 3× locally with zero failures, so this is speculative, not observed.

## Verification

- `npm run test` — pass (all packages)
- `npm run lint` — pass
- `npm run typecheck` — pass
- `packages/streamwall` suite 3× and `data.test.ts` 8× — 831/831 and 42/42 every run, no flakes
- Scope is confined to `packages/streamwall/` (2 files), English throughout, Conventional Commit title, no stray attribution.

## Follow-ups

- #596 — `combineDataSources` gating the first value on the slowest source (+ missing `fetch` timeout).
- #582 stays open for the change-entropy / module-split half, correctly not auto-closed here.

Issue #311, opened 2026-07-14
StreamWindow.setViews never reuses a running actor across a genuine content change, so seamless swap (#238) and the #198 reload fix never actually trigger


NilsR0711:
## Background

Both #198 (fixed in #234) and #238 (fixed in #306) target the `running.on.DISPLAY` transition in `viewStateMachine` -- the path that's supposed to fire when a grid cell's content changes while it's already displaying something (a playlist advance via `PlaylistScheduler`, or a manual drag-to-place reassignment). #234 made that transition skip the offscreen shuffle; #306 went further and made it preload a second `WebContentsView` in the background before swapping it in, instead of reloading in place.

Both fixes are correctly implemented and unit-tested at the `viewStateMachine` level. However, while verifying #306, I confirmed empirically that **`StreamWindow.setViews()` never actually reaches that transition for a genuine content change**.

## The problem

`setViews()` is the only place in the codebase that ever sends a `DISPLAY` event (`packages/streamwall/src/main/StreamWindow.ts`, in the `setViews` method). It reconciles the current grid against a freshly-rebuilt `viewContentMap` using three matchers, all of which require the *candidate view's current content to already equal the box's requested content*:

```js
const matchers = [
  (v, content, spaces) =>
    isEqual(v.context.content, content) &&
    v.matches({ displaying: 'running' }) &&
    intersection(v.context.pos?.spaces, spaces).length > 0,
  (v, content) =>
    isEqual(v.context.content, content) &&
    v.matches({ displaying: 'running' }),
  (v, content) => isEqual(v.context.content, content),
]
```

When a cell's content genuinely changes (e.g. a playlist advances from stream A to stream B), no existing view already shows B, so none of the matchers select the actor that's currently showing A for the space that now wants B. That actor ends up in `unusedViews` and is torn down; a brand-new actor is created via `createView()` for the new content instead. This means the actor showing A never receives a `DISPLAY` event with content B at all -- it's simply replaced.

I verified this directly against the real `setViews()` implementation (calling it twice with the same grid space but different requested content between calls, with a faked `createView`/actor) and confirmed: `createView()` is invoked again for the second call, and the original actor's `send` is never called with the new content.

`updateViewsFromStateDoc` (`packages/streamwall/src/main/index.ts`) confirms there's no continuity carried across calls either -- `viewContentMap` is rebuilt from scratch from the current Yjs `viewsState` on every update, with no memory of what a space previously showed.

## Impact

[...]

PR #595, opened 2026-07-22
refactor(streamwall): decompose StreamWindow's brain methods

### LINKED ISSUE #581: StreamWindow is a god class: setViews and the constructor are brain methods
Repowise ranks `packages/streamwall/src/main/StreamWindow.ts` as the **highest-leverage file in the repo** (weighted deficit 2734) and gives it a defect score of **2.5/10** — one of only two files in the "alert" band.

### Findings

- **God class**: 447 lines across 20 methods.
- **Brain method** `setViews` (line ~454): 98 lines, CCN 19, in a file imported by 6 others.
- **Brain method** `constructor` (line ~81): 138 lines, CCN 10.
- Churn: 21 commits in 90 days (95.7th percentile), with a sustained bug-fix history.

### Why it matters

This is the file the seamless-swap and reload work keeps returning to (see the closed #311). Every fix lands in the same oversized methods, which is exactly what the churn and bug-fix numbers show. Splitting `setViews` into explicit phases (diff the requested views against the current ones, decide reuse vs. teardown, apply) and extracting the constructor's wiring into focused private setup methods would shrink the blast radius of future changes.

### Suggested direction

- Pin current behaviour with characterisation tests first — `StreamWindow.test.ts` is itself flagged for high change entropy.
- Extract the view-diffing logic out of `setViews` into a pure, separately testable function.
- Move the constructor's IPC/window/event wiring into named private methods.

Behaviour must stay identical; this is a structural refactor, not a feature change.

NilsR0711:
Structural refactor of `packages/streamwall/src/main/StreamWindow.ts`, the file Repowise ranks as the highest-leverage in the repo (defect score 2.5/10, weighted deficit 2734). No behaviour change.

## What was extracted

**`planViewLayout()` (new `src/main/viewLayoutPlan.ts`)** — the view-diffing decision that used to live inline in `setViews`. Given the requested boxes and a list of reuse candidates, it returns a plan:

- `reused` — boxes matched to an existing actor,
- `unmatchedBoxes` — boxes that still need a brand-new view,
- `unusedViews` — actors no box claimed.

It is pure: actors are opaque payloads, and it only reads the three snapshot facts the matchers ever used (current content, occupied cells, whether the actor is in `displaying.running`). The four matchers moved across verbatim, comments included.

The two addressing axes stay separate (#397/#467/#507): the planner reasons about `CellIdx` only and never looks at a `ViewId` — reuse is decided from content and occupied cells, and `StreamWindow` keys the resulting actors by `context.id` afterwards, exactly as before.

**`setViews` is now a thin executor**: collect candidates (`reuseCandidates()`), plan, create views for the unmatched boxes, then apply via `displayPlannedViews()` and `retireUnusedViews()`. It went from 98 lines / CCN 19 to ~30 lines of sequencing.

**The constructor** keeps the same statement order but delegates to `createWallWindow()`, `createLayerView()` and `setupIpcHandlers()`, going from 138 lines / CCN 10 to a 10-line sequencer.

## Why it is behaviour-preserving

Unchanged: the matchers and their order; the outer-loop-per-matcher structure (every remaining box gets a chance at matcher *n* before matcher *n+1*); box and candidate iteration order; the tie-break by candidate order, so live views still win over parked ones; the leak-avoiding fallthrough that routes a box with no content or no matching stream into the teardown pass; the `parkedViews` clear/repopulate points and the `previouslyParkedIds` RESUME snapshot; and the point at which `this.views` is replaced (still after teardown).

Precomputing each candidate's snapshot facts once is equivalent to re-reading `getSnapshot()` inside the loop, because nothing mutates any actor between the start of the matching pass and its end — `createView()` is only called afterwards.

In particular, an actor is **still never reused across a genuine content change unless it is already `running`** (#311): the space-overlap fallback remains scoped to `displaying.running`, since `loading`/`error` have no DISPLAY handler of their own and the event would be swallowed by `displaying`'s `contentUnchanged` guard.

No `viewStateMachine` regions were added or touched, so `viewStateValueSchema` needs no mirroring.

## How behaviour was pinned

Characterisation tests were added to `StreamWindow.test.ts` **first** and confirmed green against the unmodified implementation, covering `setViews` paths that were previously unpinned:

- the same-content view already sitting in the box space wins over an equally-matching one elsewhere,
- a still-loading view with the requested content is reused rather than replaced,
- a live view beats an equally-matching parked one,
- a view no box wants is stopped and disposed.

`viewLayoutPlan.test.ts` then tests the extracted planner directly (14 cases: empty inputs, each matcher tier, the non-running fallback exclusion, no double-assignment, candidate-order tie-breaks, multi-cell boxes).

## Public surface

Unchanged — every new member is `private`, the only new export is the planner module used by its own tests.

## Verification

- `npm run test` — pass (844 + 338 + 26 + 183 + 363)
- `npm run lint` — clean
- `npm run typecheck` — clean
- `npx prettier --check .` — clean

Closes #581

### DIFF
--- a/packages/streamwall/src/main/StreamWindow.test.ts
+++ b/packages/streamwall/src/main/StreamWindow.test.ts
@@ -559,6 +559,132 @@ describe('StreamWindow.setViews reusing an actor across a genuine content change
   })
 })
 
+describe('StreamWindow.setViews matcher precedence', () => {
+  const streamA: ViewContent = { url: 'https://example.com/a', kind: 'video' }
+
+  it('prefers the same-content view already sitting in the box space over an equally-matching one elsewhere', () => {
+    const sw = makeStreamWindow(makeConfig({ cols: 2, rows: 1 }))
+    sw.win = {
+      contentView: { removeChildView: vi.fn() },
+    } as unknown as InstanceType<typeof StreamWindow>['win']
+
+    const inPlace = makeReuseTestActor({
+      id: 1,
+      content: streamA,
+      spaces: [0],
+      running: true,
+    })
+    const elsewhere = makeReuseTestActor({
+      id: 2,
+      content: streamA,
+      spaces: [1],
+      running: true,
+    })
+    // `elsewhere` is registered first, so only the space-overlap tie-break in
+    // the first matcher can make `inPlace` win.
+    sw.views = new Map([
+      [2, elsewhere.actor],
+      [1, inPlace.actor],
+    ])
+    sw.createView = vi.fn()
+
+    sw.setViews(new Map([['0', streamA]]), {
+      byURL: new Map([[streamA.url, {}]]),
+    })
+
+    expect(sw.createView).not.toHaveBeenCalled()
+    expect(inPlace.send).toHaveBeenCalledWith(
+      expect.objectContaining({ type: 'DISPLAY', content: streamA }),
+    )
+    expect(sw.views.get(1)).toBe(inPlace.actor)
+    // The redundant copy is torn down rather than left dangling.
+    expect(elsewhere.stop).toHaveBeenCalled()
+    expect(sw.views.has(2)).toBe(false)
+  })
+
+  it('reuses a still-loading view that already has the requested content instead of creating a new one', () => {
+    const sw = makeStreamWindow(makeConfig({ cols: 1, rows: 1 }))
+    sw.win = {
+      contentView: { removeChildView: vi.fn() },
+    } as unknown as InstanceType<typeof StreamWindow>['win']
+
+    const loading = makeReuseTestActor({
+      id: 1,
+      content: streamA,
+      spaces: [0],
+      running: false,
+    })
+    sw.views = new Map([[1, loading.actor]])
+    sw.createView = vi.fn()
+
+    sw.setViews(new Map([['0', streamA]]), {
+      byURL: new Map([[streamA.url, {}]]),
+    })
+
+    expect(sw.createView).not.toHaveBeenCalled()
+    expect(loading.stop).not.toHaveBeenCalled()
+    expect(loading.send).toHaveBeenCalledWith(
+      expect.objectContaining({ type: 'DISPLAY', content: streamA }),
+    )
+    expect(sw.views.get(1)).toBe(loading.actor)
+  })
+
+  it('prefers a live view over an equally-matching parked one', () => {
+    const sw = makeStreamWindow(makeConfig({ cols: 1, rows: 1 }))
+    sw.win = {
+      contentView: { removeChildView: vi.fn() },
+    } as unknown as InstanceType<typeof StreamWindow>['win']
+
+    const live = makeReuseTestActor({
+      id: 1,
+      content: streamA,
+      spaces: [0],
+      running: true,
+    })
+    const parked = makeReuseTestActor({
+      id: 2,
+      content: streamA,
+      spaces: [0],
+      running: true,
+    })
+    sw.views = new Map([[1, live.actor]])
+    sw.parkedViews = new Map([[2, parked.actor]])
+    sw.createView = vi.fn()
+
+    sw.setViews(new Map([['0', streamA]]), {
+      byURL: new Map([[streamA.url, {}]]),
+    })
+
+    expect(sw.views.get(1)).toBe(live.actor)
+    expect(live.stop).not.toHaveBeenCalled()
+    expect(parked.stop).toHaveBeenCalled()
+    // The parking bookkeeping is reset by every setViews call.
+    expect(sw.parkedViews.size).toBe(0)
+  })
+
+  it('tears down a view that no box wants and that is not parked', () => {
+    const sw = makeStreamWindow(makeConfig({ cols: 1, rows: 1 }))
+    sw.win = {
+      contentView: { removeChildView: vi.fn() },
+    } as unknown as InstanceType<typeof StreamWindow>['win']
+
+    const orphan = makeReuseTestActor({
+      id: 1,
+      content: streamA,
+      spaces: [0],
+      running: true,
+    })
+    sw.views = new Map([[1, orphan.actor]])
+    sw.createView = vi.fn()
+
+    sw.setViews(new Map(), { byURL: new Map() })
+
+    expect(orphan.stop).toHaveBeenCalled()
+    expect(orphan.disposeView).toHaveBeenCalledTimes(1)
+    expect(sw.views.size).toBe(0)
+  })
+})
+
 describe('StreamWindow.setViews expanding a view to fill the wall (issue #362)', () => {
   it('reuses the running actor and spans it across every grid cell', () => {
     const sw = makeStreamWindow(makeConfig({ cols: 2, rows: 2 }))
--- a/packages/streamwall/src/main/StreamWindow.ts
+++ b/packages/streamwall/src/main/StreamWindow.ts
@@ -8,7 +8,6 @@ import {
   WebContentsView,
 } from 'electron'
 import EventEmitter from 'events'
-import { intersection, isEqual } from 'lodash-es'
 import path from 'path'
 import {
   asViewId,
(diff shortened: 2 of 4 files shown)


NilsR0711:
**Verdict: no behavioural divergence found — approve with non-blocking test-coverage notes.** I diffed the refactor against `origin/main` statement by statement and then mutation-tested every invariant the PR claims to preserve; two of them stay green under their own mutation (both pre-existing gaps, but both now easier to break silently).

## What I verified, and how

**Matcher semantics.** All four matchers are the same predicates, only re-expressed over `ViewCandidate` fields instead of the snapshot (`viewContent` ← `v.context.content`, `viewSpaces` ← `v.context.pos?.spaces`, `isRunning` ← `v.matches({ displaying: 'running' })`). Order preserved; loop nesting preserved (matcher outer / box middle / candidate inner); box order is `boxesFromViewContentMap`'s array order fed into a `Set` in both versions; candidate order is `[...views.values(), ...parkedViews.values()]` in both, so the live-beats-parked tie-break is intact. Confirmed by mutation (M1, M6, M7 below), including the contested-cell test from #573.

**Snapshot precomputation.** Verified, not taken on trust. Between `reuseCandidates()` and the end of the matching pass the only code that runs is `planViewLayout`, whose entire body is `Set` bookkeeping plus lodash `isEqual`/`intersection`. It never touches the `view` payload (it is generic `TView`, so it *cannot*), never calls `send`/`getSnapshot`, and contains no `await`. `createView()` — the first thing that can start an actor or fire a subscription — runs strictly after the plan is returned, exactly as before. Since the whole pass is synchronous, no Electron IPC handler and no xstate transition can interleave, so the original's per-triple `getSnapshot()` re-read was already returning the same value every time. Equivalent.

Same argument for a second, unmentioned move: `width/height/cols/rows` are now read at the top of `displayPlannedViews` rather than at the top of `setViews`, i.e. across the `createView()` calls. `this.config` is only mutated by `handleResize` (Electron `resize` event) and `setGridSize` (IPC), both of which need a turn of the event loop. Synchronously equivalent.

**#311 invariant.** The space-overlap fallback is still `isRunning && intersection(...)`. Removing `isRunning` (M2) turns two tests red — the new planner unit test *and* the pre-existing "does not reuse a still-loading actor across a content change" characterisation test. Pinned from both sides.

**Leak paths.** Both fallthroughs (`!content`, `!stream`) still route the actor into `unusedViews` rather than dropping the reference, and `this.views = newViews` still happens *after* `retireUnusedViews`. Teardown order is preserved (`plan.unusedViews` is the planner's `Set` in candidate order, re-wrapped). `parkedViews.clear()` still sits after the `previouslyParkedIds` snapshot, after the candidates are read, and before `createView()`; the RESUME check in `displayPlannedViews` is at the same point relative to `DISPLAY`/`OPTIONS`. See the M3/M4 caveats below.

**Addressing axes.** `PlannableBox.spaces: CellIdx[]`, `ViewCandidate.spaces: CellIdx[] | undefined`, `previouslyParkedIds: Set<ViewId>` — the brands survive, and there is no `as any` / `as unknown` / `as number[]` anywhere in either new file. The only cast is `matchers as Array<Matcher<TView>>`, a variance widening on the payload type, not on an addressing type. The planner has no `ViewId` in scope at all, so conflation is structurally impossible.

**Constructor ordering.** Statement-for-statement identical in execution order: config fields → `resolveWindowPlacement` → `new BrowserWindow` → `removeMenu` → `loadURL` → `close` → `ready-to-show` → `resize` → `this.win` → background layer (create / bg-colour / addChildView / setBounds / loadHTML) → overlay layer (same five) → the eight `ipcMain` registrations in the same order, with the same `webPreferences` per layer. Two cosmetic deltas, both no-ops: `this.win` is assigned from the `createWallWindow()` return rather than after the handler registrations (nothing reads it in between), and `devtools-overlay` closes over `this.overlayView` instead of the local `overlayView` (same object).

**Scope.** 4 files, all under `packages/streamwall/`. No `viewStateMachine` region added or touched, so `viewStateValueSchema` needs no mirroring and the `viewStateMachineSchemaDrift.test.ts` guard stays valid. Public surface of `StreamWindow` unchanged — every addition is `private`, the only new export is the planner module. English throughout, Conventional Commit title, no stray attribution.

## Mutation table

Each bug introduced, streamwall suite run, tree restored.

| # | Mutation | Result |
|---|---|---|
| M1 | Swap matcher 1 (same content, **in place**) with matcher 2 (same content, anywhere) | **killed** — 2 failures (`matches a running view showing the same content in the same space first`, `prefers the same-content view already sitting in the box space…`) |
| M2 | Drop `isRunning` from the space-overlap fallback | **killed** — 2 failures, incl. the pre-existing #311 loading-actor test |
| M3 | Move `this.views = newViews` **before** `retireUnusedViews` | **SURVIVED** — full suite 844/844 green |
| M4 | Drop `unusedViews.add(view)` from the `!content` fallthrough | **SURVIVED** — full suite 844/844 green |
| M5 | Drop `unusedViews.add(view)` from the `!stream` fallthrough | **killed** — 2 failures (`tears down a newly created view whose box content has no matching stream…`, `also disposes an in-flight preload…`) |
| M6 | Reverse candidate order (parked before live) | **killed** — `prefers a live view over an equally-matching parked one` |
| M7 | Swap loop nesting to box-outer / matcher-inner | **killed** — 10 failures |
| M8 | Snapshot `previouslyParkedIds` **after** `parkedViews.clear()` | **killed** — `sends RESUME to a previously-parked view once it is reused after collapse` |
| M9 | Move `parkedViews.clear()` down to just before `retireUnusedViews` | survived, but this is an *equivalent mutant* — nothing between the two points reads or writes `parkedViews` — so it is not a coverage gap |
| M10 | Swap background/overlay layer creation order in the constructor | **SURVIVED** — the constructor has no coverage at all (see finding 3) |

## Findings

**Non-blocking**

1. **M4 — the `!content` leak-avoidance fallthrough is pinned by nothing.** Its sibling (`!stream`, M5) is. Pre-existing gap rather than a regression, but the PR body lists it among the invariants the characterisation tests hold, and it is the one that isn't. A test mirroring the `!stream` case — a box whose `content` is undefined must still reach `stop()` + `disposeView()` — would close it.
2. **M3 — the `this.views` / teardown ordering is pinned by nothing either.** It is observable: `view.stop()` fires the actor subscription, which calls `emitState()`, which reads `this.views`; with the assignment hoisted, that intermediate `state` event carries the new list instead of the old one. Also pre-existing and correctly preserved here — but now that teardown lives in its own method, the ordering is a two-line move away from breaking silently.
3. **The constructor has zero test coverage.** `makeStreamWindow` builds instances with `Object.create(StreamWindow.prototype)`, so it never runs; M10 and any other constructor reordering are trivially green. I verified constructor equivalence by statement-order diff only — CI cannot back that up. Not this PR's job to fix, but worth an issue given the file's churn.
4. **Latent, currently unreachable.** The original de-duplicated the reuse pool by *actor identity* (`Set<ViewActor>`); the planner de-duplicates by *candidate-object identity*. If an actor were ever present in both `views` and `parkedViews` it would now appear as two candidates and could be assigned to two boxes. Today the maps are provably disjoint (`displayPlannedViews` and `retireUnusedViews` partition the actors), so it cannot happen — but the structural immunity is gone. Keying `reuseCandidates()` by view, or a one-line comment recording the disjointness precondition, would preserve it.

**Nit**

5. The PR body says "14 cases" for `viewLayoutPlan.test.ts`; there are 12.

## Verification run locally

- `npm run test` — pass (844 / 338 / 26 / 183 / 363), matching the PR's numbers.
- `npm run lint` — clean. `npm run typecheck` — clean.
- `packages/streamwall` suite 5× in a row — 844/844 every time, no flakiness.

## Suggested follow-ups

- Pin the `!content` fallthrough and the `this.views`-after-teardown ordering (findings 1 and 2).
- Give the constructor a test seam so the `createWallWindow` / `createLayerView` / `setupIpcHandlers` sequence is actually checkable (finding 3).

PR #597, opened 2026-07-22
refactor(streamwall): extract main() startup phases into bootstrap.ts

### LINKED ISSUE #583: main/index.ts: main() is a 290-line brain method in the repo's highest-churn file
Repowise flags `packages/streamwall/src/main/index.ts` (defect score 4.4/10) for a **brain method**.

### Findings

- `main` (line ~49): **290 lines, CCN 11**, in a file imported by 19 others.
- **46 commits in 90 days — the 99.4th churn percentile**, the busiest file in the repository.
- 22 bug fixes recorded in its history.

### Why it matters

Every startup concern — config loading, window creation, IPC wiring, state setup, updater bootstrap — lives in one function, which is why nearly every feature change has to touch it. That concentration is what drives both the churn and the bug-fix count.

### Suggested direction

Extract the distinct startup phases from `main` into named, individually testable bootstrap functions, keeping `main` as a thin sequencer. Purely structural; no behaviour change.

NilsR0711:
`main()` in `packages/streamwall/src/main/index.ts` had grown into a 290-line
brain method (CCN 11) holding every startup concern at once — the reason
nearly every feature change had to touch the repository's highest-churn file.

The distinct startup phases now live in a sibling `bootstrap.ts` as named
functions that take their collaborators explicitly instead of reaching for
module-level state, and `main()`/`init()` are thin sequencers over them.

### Phases extracted

| Function | Phase |
| --- | --- |
| `setupApplicationMenu` | Resolve the user config location and install the application menu |
| `createPersistedLocalStreamData` | Local stream data store, mirrored back into storage on every update |
| `createInitialClientState` | Initial broadcast state, seeded from persisted storage |
| `restoreStateDoc` | Restore the base64 Yjs snapshot, tolerating a corrupt one |
| `createStateDocPersister` | Throttled state-doc persistence (returns the throttled fn so shutdown can flush it) |
| `seedAndObserveViewsState` | Seed the grid's view state, render once, then observe |
| `startPlaylistScheduler` | Playlist cycling for configured views |
| `createBrowseWindow` | Browse window creation + session hardening + popup denial |
| `wireWindowIpc` | Stream/control window and state-doc message flow |
| `wireWindowLifecycle` | Close/hide, dock activate, window-all-closed, storage flush before quit |
| `setupStreamdelayClient` | Optional Streamdelay connection |
| `setupTwitchBot` | Optional Twitch chat bot |
| `createDataSourceHealthReporter` | Per-data-source health reporting into the broadcast state |
| `configureSentry` | Sentry init plus the subprocess opt-in switch |
| `configureElectronRuntime` | Chromium display switches and sandbox mode |

### Ordering is unchanged

This is purely structural — no behaviour change. Every side effect keeps its
exact relative order, including the order-sensitive parts:

- `viewsState.observeDeep` fires synchronously at the end of a `transact`, so
  the grid config the observer reads is still set before the seeding
  transaction, and the observer is still registered *after* the seeding plus
  the single explicit initial render (`seedAndObserveViewsState`).
- `wireWindowLifecycle` registers its handlers in the original order, so the
  quit-intent `before-quit` listener still runs before the storage-flush
  `before-quit` listener; the flush's own `app.quit()` still passes straight
  through on the second pass.
- `configureSentry` still appends the Sentry switch before the display
  switches, which are still applied before `app.enableSandbox()`.

### Left inline (and why)

- `updateViewsFromStateDoc` and `updateState` — both read and reassign the
  mutable `clientState` binding that half of `main()` closes over. Extracting
  them would mean either passing a mutable state container everywhere or
  changing when `clientState` is reassigned, which is exactly the reordering
  risk this refactor avoids. They stay as local closures.
- The `createOnCommand(...)` call — already a call into an extracted module;
  what remains is argument wiring bound to those same closures.
- The `for await (const streams of combineDataSources(...))` loop — it is the
  long-lived tail of `main()`, not a startup phase.

### Tests

New `bootstrap.test.ts` covers all extracted phases (32 cases), including
regression guards for the two ordering traps above: the views-state seeding
renders exactly once with the full grid already present, and the shutdown
hook defers the first quit until storage is flushed, then quits once.
No existing test was changed or weakened.

`npm run test`, `npm run lint` and `npm run typecheck` all pass.

Closes #583

### DIFF
--- a/packages/streamwall/src/main/bootstrap.test.ts
+++ b/packages/streamwall/src/main/bootstrap.test.ts
@@ -1,0 +1,651 @@
+import EventEmitter from 'node:events'
+import { join } from 'node:path'
+import type { StreamwallState, ViewId } from 'streamwall-shared'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import * as Y from 'yjs'
+import {
+  type LifecycleApp,
+  type LifecycleWindow,
+  type PreventableEvent,
+  configureElectronRuntime,
+  configureSentry,
+  createBrowseWindow,
+  createDataSourceHealthReporter,
+  createInitialClientState,
+  createPersistedLocalStreamData,
+  createStateDocPersister,
+  restoreStateDoc,
+  seedAndObserveViewsState,
+  setupApplicationMenu,
+  setupStreamdelayClient,
+  setupTwitchBot,
+  startPlaylistScheduler,
+  wireWindowIpc,
+  wireWindowLifecycle,
+} from './bootstrap'
+import type { StreamwallConfig } from './cliArgs'
+import log from './logger'
+
+afterEach(() => {
+  vi.restoreAllMocks()
+  vi.useRealTimers()
+})
+
+function fakeStreamwallState(
+  overrides: Partial<StreamwallState> = {},
+): StreamwallState {
+  return {
+    ...createInitialClientState({
+      config: { width: 0, height: 0, x: 0, y: 0, cols: 2, rows: 2 },
+      layoutPresets: [],
+      favorites: [],
+    }),
+    ...overrides,
+  }
+}
+
+describe('setupApplicationMenu', () => {
+  it('resolves the user config path inside the userData directory', () => {
+    const installMenu = vi.fn()
+    const result = setupApplicationMenu({
+      userDataPath: '/user/data',
+      logFilePath: '/logs/main.log',
+      fileExists: () => true,
+      installMenu,
+    })
+
+    expect(result).toEqual({
+      userConfigPath: join('/user/data', 'config.toml'),
+      hasUserConfig: true,
+    })
+    expect(installMenu).toHaveBeenCalledWith(
+      join('/user/data', 'config.toml'),
+      '/logs/main.log',
+      true,
+    )
+  })
+
+  it('reports a missing user config to the menu', () => {
+    const installMenu = vi.fn()
+    const result = setupApplicationMenu({
+      userDataPath: '/user/data',
+      logFilePath: '/logs/main.log',
+      fileExists: () => false,
+      installMenu,
+    })
+
+    expect(result.hasUserConfig).toBe(false)
+    expect(installMenu).toHaveBeenCalledWith(
+      expect.any(String),
+      expect.any(String),
+      false,
+    )
+  })
+})
+
+describe('createPersistedLocalStreamData', () => {
+  it('seeds from stored entries and persists later updates', () => {
+    const persistEntries = vi.fn()
+    const localStreamData = createPersistedLocalStreamData({
+      initialEntries: [{ link: 'http://a.test/', kind: 'video' }],
+      persistEntries,
+    })
+
+    expect(localStreamData.dataByURL.get('http://a.test/')).toBeDefined()
+    expect(persistEntries).not.toHaveBeenCalled()
+
+    localStreamData.update('http://b.test/', { kind: 'video' })
+
+    expect(persistEntries).toHaveBeenCalledTimes(1)
+    const entries = persistEntries.mock.calls[0][0]
+    expect(entries.map((entry: { link?: string }) => entry.link)).toContain(
+      'http://b.test/',
+    )
+  })
+})
+
+describe('createInitialClientState', () => {
+  it('carries the persisted presets and favorites into a local identity', () => {
+    const state = createInitialClientState({
+      config: { width: 1, height: 2, x: 3, y: 4, cols: 3, rows: 2 },
+      layoutPresets: [],
+      favorites: [],
+    })
+
+    expect(state.identity).toEqual({ role: 'local' })
+    expect(state.config.cols).toBe(3)
+    expect(state.streams).toEqual([])
+    expect(state.fullscreenViewIdx).toBeNull()
+    expect(state.streamdelay).toBeNull()
+    expect(state.dataSourceHealth).toEqual([])
+  })
+})
+
+describe('restoreStateDoc', () => {
+  it('applies a stored snapshot', () => {
+    const source = new Y.Doc()
+    source.getMap('views').set('0', 'stored')
+    const encoded = Buffer.from(Y.encodeStateAsUpdate(source)).toString(
+      'base64',
+    )
+
+    const target = new Y.Doc()
+    restoreStateDoc(target, encoded)
+
+    expect(target.getMap('views').get('0')).toBe('stored')
+  })
+
+  it('keeps startup alive when the snapshot is corrupt', () => {
+    const warn = vi.spyOn(log, 'warn').mockImplementation(() => undefined)
+    const target = new Y.Doc()
+
+    expect(() => {
+      restoreStateDoc(target, 'bm90LWEteWpzLXVwZGF0ZQ==')
+    }).not.toThrow()
+    expect(warn).toHaveBeenCalled()
+  })
+})
+
+describe('createStateDocPersister', () => {
+  it('throttles snapshot writes and can flush a pending one', () => {
+    vi.useFakeTimers()
+    const stateDoc = new Y.Doc()
+    const saveSnapshot = vi.fn()
+    const persist = createStateDocPersister({
+      stateDoc,
+      saveSnapshot,
+      wait: 1000,
+    })
+
+    stateDoc.getMap('views').set('0', 'a')
+    stateDoc.getMap('views').set('1', 'b')
+    expect(saveSnapshot).toHaveBeenCalledTimes(1)
+
+    persist.flush()
+    expect(saveSnapshot).toHaveBeenCalledTimes(2)
+
+    const restored = new Y.Doc()
+    const lastCall = saveSnapshot.mock.calls.at(-1)
+    Y.applyUpdate(restored, Buffer.from(String(lastCall?.[0]), 'base64'))
+    expect(restored.getMap('views').get('1')).toBe('b')
+  })
+})
+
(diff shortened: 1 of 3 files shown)


NilsR0711:
**Verdict: no reordering found — the "ordering is unchanged" claim holds. Non-blocking notes only.**

## Side-effect ordering diff

I enumerated every observable side effect in `main()` and `init()` on `origin/main` (f8d1349) top to bottom, then re-enumerated the new `main()`/`init()` in execution order by inlining each phase function. Both lists, 1:1:

| # | Side effect | old | new |
| --- | --- | --- | --- |
| 1 | `loadStorage(...)` (await) | ✓ | ✓ |
| 2 | resolve `config.toml` path, `existsSync`, `installApplicationMenu` | ✓ | `setupApplicationMenu` |
| 3 | `new StreamIDGenerator` | ✓ | ✓ |
| 4 | `new LocalStreamData(db…)` → `.on('update')` persist | ✓ | `createPersistedLocalStreamData` |
| 5 | `new LocalStreamData()` (overlay) | ✓ | ✓ |
| 6 | `buildStreamWindowConfig` / `buildRetryConfig` | ✓ | ✓ |
| 7 | `new StreamWindow` | ✓ | ✓ |
| 8 | `new ControlWindow` | ✓ | ✓ |
| 9 | `setupAppUpdater` | ✓ | ✓ |
| 10 | initial `clientState` literal | ✓ | `createInitialClientState` |
| 11 | `new Y.Doc()` + `getMap('views')` | ✓ | ✓ |
| 12 | `Y.applyUpdate` of stored snapshot (guarded, try/catch) | ✓ | `restoreStateDoc` |
| 13 | `throttle(...)` + `stateDoc.on('update', persistStateDoc)` | ✓ | `createStateDocPersister` |
| 14 | `initializeViewsState(cols, rows)` (transact) | ✓ | `seedAndObserveViewsState` |
| 15 | `updateViewsFromStateDoc()` — the single initial render | ✓ | ✓ |
| 16 | `viewsState.observeDeep(updateViewsFromStateDoc)` | ✓ | ✓ |
| 17 | `new PlaylistScheduler` + `.start()` | ✓ | `startPlaylistScheduler` |
| 18 | `createOnCommand(...)` | ✓ | ✓ |
| 19 | `new EventEmitter` (stateEmitter) | ✓ | ✓ |
| 20 | `streamWindow.on('state'/'resize'/'load')` | ✓ | `wireWindowIpc` |
| 21 | `stateDoc.on('update')` → control-window echo | ✓ | ✓ |
| 22 | `controlWindow.on('load')`, `on('ydoc')`, `setCommandHandler` | ✓ | ✓ |
| 23 | `app.on('before-quit')` — quit intent | ✓ | `wireWindowLifecycle` |
| 24 | `app.on('activate')` | ✓ | ✓ |
| 25 | `streamWindow.on('close')`, `controlWindow.on('close')` | ✓ | ✓ |
| 26 | `app.on('window-all-closed')` | ✓ | ✓ |
| 27 | `app.on('before-quit')` — storage flush | ✓ | ✓ |
| 28 | `connectControlUplink` | ✓ | ✓ |
| 29 | Streamdelay client + `.connect()` | ✓ | `setupStreamdelayClient` |
| 30 | Twitch bot + `.connect()` | ✓ | `setupTwitchBot` |
| 31 | `new DataSourceHealthTracker` | ✓ | `createDataSourceHealthReporter` |
| 32 | `buildDataSources` | ✓ | ✓ |
| 33 | `for await (combineDataSources)` loop | ✓ | ✓ |

`init()`: `initLogger` → `parseArgs` → `setLogLevel` → help bail → `Sentry.init` (if enabled) → `appendSwitch(SENTRY_ENABLED_SWITCH, …)` → `high-dpi-support` → `force-device-scale-factor` → `enableSandbox()` → `whenReady().then(main)`. Identical on both sides, including the three `log.debug` markers, which moved into the phases rather than disappearing.

**Diff result: empty.** No side effect added, removed, or moved relative to any other.

## Traps

**`observeDeep`** — clean. `seedAndObserveViewsState` seeds → renders once → registers the observer, in that order, and `streamWindowConfig` (the `cols`/`rows` the observer reads via `updateViewsFromStateDoc`) is still built at the top of `main()`, ~70 lines before the seeding transaction. Nothing the observer reads moved after the transact. `clientState` is likewise initialised before the seeding.

**`before-quit` order** — registration order is preserved (quit-intent, then flush). But the PR description overstates why this matters: both listeners run in the *same synchronous* `before-quit` emit, and `preventDefault()` does not stop later listeners, so `isQuitting = true` is set on that emit regardless of which handler is registered first. The flush's `app.quit()` only lands in a `.finally()`, i.e. a later tick. I verified this by swapping the two registrations — all 32 tests still pass (see the mutation table). That is not a test gap: the swap is behaviour-neutral. The genuinely load-bearing part is the `storageFlushed` latch, and that *is* covered.

**Yjs snapshot relocation** — verified sound. Old: `Y.encodeStateAsUpdate` ran inside the `safeUpdate` mutator. New: it runs immediately before `saveSnapshot(...)`. `safeUpdate` calls `await db.update(fn)`; lowdb's `Low.update` is `async update(fn) { fn(this.data); await this.write() }`, and an async function body runs synchronously up to its first `await` — so `fn` is invoked synchronously from the same call frame. Encode → assign therefore still happens in one uninterrupted synchronous run inside the throttle callback, with no `await`, no microtask boundary and no possible doc mutation in between. No blocking finding.

**Inline leftovers** — reasoning holds. Every extracted phase that needs `clientState` receives a `getClientState: () => clientState` / `getStreams: () => clientState.streams` accessor or `updateState` itself; none captures a `clientState` *value*. `createDataSourceHealthReporter(updateState)` captures the hoisted function declaration, which reads and writes the live binding. `updateViewsFromStateDoc` (which reassigns `clientState` on `clearedFullscreen`) is passed by reference and still reassigns the same binding. No stale-closure path exists.

## Mutation testing the ordering guards

| # | Mutation | Result |
| --- | --- | --- |
| 1 | Move the quit-intent `before-quit` registration *after* the flush registration | **survives** — 32/32 pass. Behaviour-neutral (see above), so this is not a real gap, but the "registration order determines execution order" framing invites the wrong conclusion. |
| 2 | `observeDeep` registered *before* `initializeViewsState` | **killed** — `seeds every grid cell before rendering the wall once`: expected 1 call, got 2 |
| 3 | Call `updateViews()` twice during seeding | **killed** — same test, same assertion |
| 4 | Drop the `storageFlushed` latch from the flush handler | **killed** — `defers the first quit until storage is flushed, then quits once` |
| 5 | `enableSandbox()` before the display switches | **killed** — `applies the display switches before enabling the sandbox` |

Tree restored to the PR head after each; `git status` clean.

## Exports / dependents

`index.ts` has **zero exports** on both sides — confirmed. Grepping the monorepo for importers of `main/index` returns exactly one real reference, `forge.config.ts:82` (`entry: 'src/main/index.ts'`); the other four hits are prose mentions inside comments in `sentryConfig.ts`, `renderer/UpdateBanner.tsx` and `main/linuxUpdateChecker.ts`. The "imported by 19 others" figure quoted in #583 is the *outbound* import count of `index.ts`, not its dependents. Worth correcting in the issue, because it changes the risk story: this file is a leaf entry point, so the blast radius of the refactor is the Electron startup path alone.

## Findings

**Blocking:** none.

**Minor**
1. `createStateDocPersister({ wait })`, `setupStreamdelayClient({ createClient })` and `setupTwitchBot({ createBot })` add optional seams that only the tests use. They are cheap, but they are production surface that exists for tests; a note in the JSDoc saying so would keep the next reader from thinking they are configurable.
2. `wireWindowLifecycle`'s JSDoc ("the quit-intent flag is set before the storage flush hook runs") is factually right but implies a dependency that does not exist. Consider saying instead that both listeners fire in the same emit and the real invariant is the `storageFlushed` latch — otherwise a future reader may "preserve" the wrong thing.
3. `startPlaylistScheduler.resolveStreamId` now snapshots `getStreams()` once instead of reading `clientState.streams` twice. Equivalent (both reads were synchronous), and arguably tighter — just noting it as a real, harmless textual change.
4. `setupTwitchBot` returns the bot, but `main()` discards it. Fine as-is (the original also let it go out of scope); the return value is currently only for tests.

**Non-blocking observation**
`streamdelayClient` is now assigned only after `setupStreamdelayClient` returns, i.e. after `.on('state')` and `.connect()`. The original assigned the binding before both. `StreamdelayClient.connect()` only constructs a `ReconnectingWebSocket` and emits `'state'` on message, so nothing can call `getStreamdelayClient()` during that window — unobservable today, but it is the one place where a future synchronous emit inside `connect()` would behave differently than before.

## Verification

`npm run test`: pass, twice (streamwall 860/860 · control-server 183/183 · control-ui 363/363 · shared 338/338 · control-client 26/26 · root 131/131), no flakes. `npm run lint`: clean. `npm run typecheck`: clean across all packages. `bootstrap.test.ts` alone: 32/32.

Scope is confined to `packages/streamwall/src/main/{index,bootstrap,bootstrap.test}.ts`. Conventional Commit title, English throughout, no attribution footers.

## Follow-ups worth filing

- Correct the "imported by 19 others" claim in #583 and note that `index.ts` is a zero-export entry point.
- Nothing exercises `main()` itself end to end. The five mutations above all live inside `bootstrap.ts`; a mutation to the *call order in `main()`* (e.g. `wireWindowLifecycle` before `wireWindowIpc`) would still go unnoticed. A single startup-sequence test that stubs the phase module and asserts the call order would close that, and is a cheap guard for the next change to this file.

Issue #578, opened 2026-07-21
fix(release): dispatched runs do not satisfy the release PR's required checks


NilsR0711:
Follow-up to #574. With the dispatch step working, `release-please.yml` does start `ci.yml` and `pr-title.yml` against the release branch and both complete successfully as check runs on the branch head. But a `workflow_dispatch` run is not associated with the pull request, so the checks never appear in the PR's status check rollup — for #518 the rollup contained only `CodeQL`, and merging failed with:

```
Repository rule violations found
2 of 2 required status checks are expected.
```

The v0.10.0 release was unblocked by closing and reopening #518, which raises a real `pull_request` (reopened) event; both `ci.yml` and `pr-title.yml` already listen for it, the checks then attached to the PR, `mergeStateStatus` became `CLEAN` and the merge went through.

So the dispatch shim (#547) does not do what it was added for. Options:

- drop the dispatch and document close/reopen as the release step, or
- have the workflow close and reopen the release PR itself (needs a token other than `GITHUB_TOKEN` — an event raised with that token would again not trigger the workflows, cf. #549), or
- keep the dispatch only as a smoke signal and stop implying in CONTRIBUTING.md that it satisfies branch protection.

`CONTRIBUTING.md` ("Cutting a release", step 2) currently describes the dispatch as what makes the required checks pass and needs to be corrected either way.

Issue #549, opened 2026-07-21
Run release-please with a PAT or App token instead of GITHUB_TOKEN


NilsR0711:
## Problem

`release-please.yml` runs with the workflow's `GITHUB_TOKEN`. GitHub raises no
workflow-triggering events for that token, which costs the release flow two
manual or synthetic steps:

- The release PR gets no `pull_request` events, so neither `CI OK` nor
  `Conventional Commits title` starts (#521). PR #547 works around this by
  dispatching both workflows via `workflow_dispatch`, the one event GitHub
  still delivers for `GITHUB_TOKEN`.
- A tag pushed by the token would not trigger `release.yml`, so
  release-please runs with `skip-github-release` and the `vX.Y.Z` tag stays a
  manual `git push` (#515). `release-tag.yml` exists only to notice when that
  step is forgotten.

## Suggested solution

Run `googleapis/release-please-action` with a fine-grained PAT or a GitHub App
installation token (`actions/create-github-app-token`). Events raised by either
are indistinguishable from a user's, so:

- the release PR starts its required checks by itself, and the dispatch shim
  added in #547 (`workflow_dispatch` triggers on `ci.yml` and `pr-title.yml`,
  the `actions: write` scope and the dispatch step in `release-please.yml`) can
  be removed again;
- `skip-github-release` and the manual tag push can go, since a
  release-please-created tag would start `release.yml` normally;
- `release-tag.yml`'s daily "is the release tagged?" watchdog becomes
  redundant, or at least far less likely to fire.

A GitHub App is preferable to a PAT: the token is short-lived, scoped to this
repository, and not tied to a personal account.

## Impact

Removes the last manual step from the release (the tag push), plus the shim
that only exists because the token cannot trigger workflows. Cost: one App
installation (or one PAT plus its rotation) to set up and own.

Follow-up to #521 / PR #547; supersedes the workaround documented in #520.

NilsR0711:
Deferred for now — labelled `priority:low`.

The change needs a credential only a maintainer can create (a GitHub App installation, or a fine-grained PAT) plus the matching repository secret; the repo currently has no Actions secrets or variables at all. Nothing in the code can be prepared meaningfully ahead of that: merging the workflow rewrite before the token exists would break `release-please.yml` outright.

Weighed against that, the payoff is small and the current flow works:

- Releasing is still gated where it matters — the release PR has to be merged by hand, and branch protection plus the required checks are unaffected either way.
[...]

PR #609, opened 2026-07-22
docs(contributing): document close/reopen as the release PR's check trigger

NilsR0711:
Addresses the documentation half of #578. No workflow changes — the dispatch shim and `GITHUB_TOKEN` setup stay as they are.

## The problem

"Cutting a release", step 2 told the reader that `release-please.yml` dispatches `ci.yml` and `pr-title.yml` against the release branch, and to "wait for the two checks to pass before merging". That implies the dispatched runs satisfy branch protection. They do not.

A `workflow_dispatch` run is not associated with the pull request, so its check runs report against the branch head and never enter the PR's status check rollup. For #518 the rollup contained only `CodeQL`, and the merge failed with:

```
Repository rule violations found
2 of 2 required status checks are expected.
```

The v0.10.0 release was unblocked by closing and reopening the release PR, which raises a real `pull_request` (reopened) event. Both workflows listen for it — `ci.yml` via a bare `pull_request:` trigger (whose default types include `reopened`) and `pr-title.yml` by listing `reopened` explicitly — so the checks attached to the PR, `mergeStateStatus` became `CLEAN`, and the merge went through.

Anyone following the current instructions waits for green checks, tries to merge, and hits the same wall. The documented workaround does not work, and the one that does was undocumented.

## The change

Step 2 now:

- states close/reopen as the actual, non-optional step, with the reason
- demotes the dispatched runs to a smoke signal and says explicitly that they cannot satisfy branch protection
- keeps the existing note that a dispatched run has nothing to diff against and therefore always runs the full gate
- points at #549 (PAT / GitHub App token) as the change that would make both workarounds unnecessary

## Scope

`CONTRIBUTING.md` only. #578 stays open for the workflow-side decision, which needs a token other than `GITHUB_TOKEN` and is therefore blocked on repository configuration.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -381,14 +381,28 @@ breaking change bumps the minor version.
    manifests, their `package-lock.json` entries and the `CHANGELOG.md` section
    for the pending release. Its description previews the release notes;
    anything merged afterwards updates the PR.
-2. Merge the release PR when you want to cut the release. A pull request opened
-   by a workflow's `GITHUB_TOKEN` raises no `pull_request` events, so neither
-   `CI OK` nor `Conventional Commits title` would start on its own. Instead,
-   `release-please.yml` dispatches both workflows against the release branch
-   after every update — `workflow_dispatch` is the one event GitHub still
-   delivers for that token. Wait for the two checks to pass before merging.
-   Because a dispatched run has nothing to diff against, it always runs the
-   full gate rather than the change-scaled subset.
+2. Close and reopen the release PR, then merge it when the checks have passed.
+
+   The close/reopen step is not optional. A pull request opened by a workflow's
+   `GITHUB_TOKEN` raises no `pull_request` events, so neither `CI OK` nor
+   `Conventional Commits title` starts on its own, and branch protection blocks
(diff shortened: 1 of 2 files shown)


NilsR0711:
The first CI run failed, and the cause turned out to be more interesting than the doc wording.

`test/release-please.test.mjs` carried a guard asserting that `CONTRIBUTING.md` must **not** mention reopening the release PR:

```js
assert.doesNotMatch(
  contributing,
  /reopen the release PR/i,
  'release-please.yml starts the checks itself now (#521)',
)
```

That assumption came from #547, where the dispatch shim was expected to make the close/reopen workaround unnecessary. It does not — which is the whole finding of #578. So the test was actively enforcing documentation that sends the next release into the same wall v0.10.0 hit.

The guard is now inverted: it asserts that the close/reopen step is documented, and that the dispatched runs are described as a smoke signal rather than as something that satisfies branch protection. Both assertions carry the issue reference, and the reasoning is recorded in a comment above the test so the next person does not "fix" it back.

Verified in both directions: against the previous `CONTRIBUTING.md` from `main` the new guard fails (`not ok 12`), against this PR's version it passes — so it pins the change rather than just accommodating it.

PR #638, opened 2026-07-23
fix(control-client): derive WebSocket scheme from the page protocol

### LINKED ISSUE #617: Control client hardcodes ws://, so the published Docker image cannot connect behind TLS
## Summary
The control client derives its WebSocket endpoint as `ws://${location.host}` whenever `BASE_URL` is `/`, and the published control-server image always builds the client with `BASE_URL = '/'` — so any HTTPS deployment using the GHCR image gets a mixed-content-blocked socket and a control UI that never connects.

## Evidence
`packages/streamwall-control-client/src/index.tsx:8-10`:

```ts
const connection = useStreamwallWebsocketConnection(
  (BASE_URL === '/' ? `ws://${location.host}` : BASE_URL) + '/client/ws',
)
```

`packages/streamwall-control-client/vite.config.ts:7` sets `base: process.env.STREAMWALL_CONTROL_URL ?? '/'`, but `packages/streamwall-control-server/Dockerfile` has no `ARG`/`ENV` for `STREAMWALL_CONTROL_URL` before `npm -w streamwall-control-client run build` (line 33), and `deploy/docker-compose.yml` passes no build args either (`env_file` applies at runtime only). The image published by `release.yml` therefore always ships a client built with base `/`.

## Failure scenario
Self-hoster follows `docs/self-hosting.md`: Caddy terminates TLS, page loads from `https://control.example.com`, client opens `ws://control.example.com/client/ws`. Browsers block insecure WebSockets from secure contexts, so the UI sits disconnected with no clear diagnostic. The e2e suite runs over plain `http://127.0.0.1`, where `ws://` is fine, so this never surfaces in CI.

## Suggested fix
Derive the scheme from the page: `location.protocol === 'https:' ? 'wss:' : 'ws:'` (and treat an `http(s)://` `BASE_URL` the same way). That also removes the need to bake `STREAMWALL_CONTROL_URL` into the image at build time. Consider an e2e/docker smoke that exercises the client over TLS or at least asserts the derived endpoint scheme.

## Impact
High - the documented self-hosting path (published image + Caddy TLS) produces a control UI that cannot connect.

NilsR0711:
## Summary

The control client hardcoded `ws://` whenever `BASE_URL` was `/`, and the published control-server Docker image always builds the client with that default base (no `STREAMWALL_CONTROL_URL` build arg exists in the Dockerfile). Any HTTPS deployment — the documented Caddy self-hosting path — therefore opened an insecure WebSocket from a secure context, which browsers block, leaving the control UI permanently disconnected with no clear diagnostic.

## Changes

- New `getWebsocketEndpoint(baseUrl, location)` helper in `packages/streamwall-control-client/src/wsEndpoint.ts`:
  - relative bases (`/`, `/path/`) resolve against the page host with the scheme derived from the page protocol (`https:` → `wss:`, otherwise `ws:`),
  - absolute `http(s)://` bases (`STREAMWALL_CONTROL_URL` builds) are mapped to `ws(s)://` the same way,
  - `ws://` / `wss://` bases pass through unchanged.
- `src/index.tsx` uses the helper instead of the hardcoded `ws://${location.host}` fallback.
- Unit tests cover the TLS regression, local dev (plain http/ws stays `ws://`), absolute and path bases, ports, and trailing-slash handling.

This also removes any need to bake `STREAMWALL_CONTROL_URL` into the image at build time.

## Testing

- `npm run test` (all workspaces), `npm run lint`, `npm run typecheck`, Prettier — all green locally.

Closes #617

### DIFF
--- a/packages/streamwall-control-client/src/index.tsx
+++ b/packages/streamwall-control-client/src/index.tsx
@@ -1,12 +1,13 @@
 import { render } from 'preact'
 import { ControlUI, GlobalStyle } from 'streamwall-control-ui'
 import { useStreamwallWebsocketConnection } from './useStreamwallWebsocketConnection.ts'
+import { getWebsocketEndpoint } from './wsEndpoint.ts'
 
 function App() {
   const { BASE_URL } = import.meta.env
 
   const connection = useStreamwallWebsocketConnection(
-    (BASE_URL === '/' ? `ws://${location.host}` : BASE_URL) + '/client/ws',
+    getWebsocketEndpoint(BASE_URL, location),
   )
 
   return (
--- a/packages/streamwall-control-client/src/wsEndpoint.test.ts
+++ b/packages/streamwall-control-client/src/wsEndpoint.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it } from 'vitest'
+import { getWebsocketEndpoint } from './wsEndpoint.ts'
+
+const httpPage = { protocol: 'http:', host: 'localhost:3000' }
+const httpsPage = { protocol: 'https:', host: 'control.example.com' }
+
+describe('getWebsocketEndpoint', () => {
+  it('uses ws:// for the default base on an http page (local dev)', () => {
+    expect(getWebsocketEndpoint('/', httpPage)).toBe(
+      'ws://localhost:3000/client/ws',
+    )
+  })
+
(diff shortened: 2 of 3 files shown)


NilsR0711:
Follow-up for the TLS e2e coverage suggested in #617: #639

PR #640, opened 2026-07-23
fix(streamwall): ignore invalid rotation instead of clobbering the current one

### LINKED ISSUE #627: RotationController applies an invalid rotation class instead of ignoring it, clobbering the current rotation
## Summary
`_update()` warns that an invalid rotation is "ignored" but then unconditionally overwrites `className`, replacing a previously valid rotation class with an inert one.

## Evidence
`packages/streamwall/src/preload/mediaPreload.ts:112-118`:

```ts
_update() {
  const rotation = this.customRotation % 360
  if (![0, 90, 180, 270].includes(rotation)) {
    console.warn('ignoring invalid rotation', rotation)   // says "ignoring"...
  }
  this.video.className = `__rot${rotation}__`             // ...but applies it anyway
}
```

JS `%` keeps the sign, so e.g. -90 yields `__rot-90__`; only 0/90/180/270 have CSS rules.

## Failure scenario
A stream rotated to a valid 90° receives a bad rotation value; instead of keeping 90°, the video loses its `__rot90__` class and renders unrotated — contradicting the log message.

## Suggested fix
Early-return (keeping the current class) when the normalized rotation is not one of the four supported values, and normalize negatives (`((r % 360) + 360) % 360`) before validating.

## Impact
Low - requires an out-of-range rotation value to trigger, but the behavior contradicts its own warning.

NilsR0711:
## Summary

`RotationController._update()` in `packages/streamwall/src/preload/mediaPreload.ts` logged that an invalid rotation was being "ignored" but then unconditionally overwrote the video's `className`, replacing a previously valid rotation class (e.g. `__rot90__`) with an inert one — so a stream that received a bad rotation value lost its rotation and rendered unrotated, contradicting its own warning.

Two related problems:
- JS `%` keeps the operand's sign, so `-90 % 360 === -90`, producing `__rot-90__`, for which no CSS rule exists.
- Even for a genuinely unsupported value, the class was applied instead of skipped.

## Fix

- Normalize into `[0, 360)` with `(((r % 360) + 360) % 360)` so negatives map correctly (`-90` -> `270`).
- Early-return when the normalized rotation is not one of `0/90/180/270`, keeping the current rotation class intact.

## Tests

Added a `RotationController` describe block to `mediaPreload.test.ts` covering:
- each supported rotation maps to its class,
- negative and `>= 360` values normalize correctly,
- an invalid value keeps the current class and warns.

Verified the two regression tests fail against the old code and pass with the fix. Full streamwall suite green (899 tests). Lint, typecheck, and Prettier all pass.

Closes #627

### DIFF
--- a/packages/streamwall/src/preload/mediaPreload.test.ts
+++ b/packages/streamwall/src/preload/mediaPreload.test.ts
@@ -985,3 +985,49 @@ describe('SnapshotController poster object URLs', () => {
     expect(revokeObjectURL).toHaveBeenCalledTimes(2)
   })
 })
+
+describe('RotationController', () => {
+  afterEach(() => {
+    vi.restoreAllMocks()
+    vi.resetModules()
+  })
+
+  async function makeController() {
+    const { RotationController } = await import('./mediaPreload')
+    const video = document.createElement('video')
+    return { controller: new RotationController(video), video }
+  }
+
+  it('applies each supported rotation as its matching class', async () => {
+    const { controller, video } = await makeController()
+    for (const rotation of [0, 90, 180, 270]) {
+      controller.setCustom(rotation)
+      expect(video.className).toBe(`__rot${rotation}__`)
+    }
+  })
+
+  it('normalizes negative rotations into [0, 360) instead of emitting an inert class', async () => {
+    const { controller, video } = await makeController()
+    controller.setCustom(-90)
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up (out of scope): #643 — `RotationController.siteRotation` is a dead field, spotted while fixing this issue.

PR #641, opened 2026-07-23
fix(control-client): derive the WebSocket scheme from the page protocol

### LINKED ISSUE #617: Control client hardcodes ws://, so the published Docker image cannot connect behind TLS

NilsR0711:
## Problem
The control client derives its WebSocket endpoint as `ws://${location.host}` whenever `BASE_URL` is `/`, and the published control-server image always builds the client with `BASE_URL = '/'`. Any HTTPS deployment using the GHCR image therefore gets a mixed-content-blocked socket and a control UI that never connects.

## Solution
The endpoint is now resolved by a small pure function, `controlWebSocketEndpoint(baseURL, location)`:
- `BASE_URL === '/'`: the scheme follows the page protocol (`https:` → `wss://`, else `ws://`), so the same image works over plain HTTP and behind TLS without baking `STREAMWALL_CONTROL_URL` into the build.
- An explicit `http(s)://` base is mapped to its WebSocket counterpart (`http` → `ws`, `https` → `wss`) and upgraded to `wss://` whenever the page itself is secure, since a secure context can never open an insecure socket.
- A trailing slash on an explicit base no longer produces a double slash in the endpoint path.

## Testing
- New unit tests in `packages/streamwall-control-client/src/wsEndpoint.test.ts` cover both schemes for `BASE_URL = '/'`, explicit http/https bases, the https-page upgrade, and trailing-slash normalization.
- `npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` all pass.

Closes #617

### DIFF
--- a/packages/streamwall-control-client/src/index.tsx
+++ b/packages/streamwall-control-client/src/index.tsx
@@ -1,12 +1,13 @@
 import { render } from 'preact'
 import { ControlUI, GlobalStyle } from 'streamwall-control-ui'
 import { useStreamwallWebsocketConnection } from './useStreamwallWebsocketConnection.ts'
+import { controlWebSocketEndpoint } from './wsEndpoint.ts'
 
 function App() {
   const { BASE_URL } = import.meta.env
 
   const connection = useStreamwallWebsocketConnection(
-    (BASE_URL === '/' ? `ws://${location.host}` : BASE_URL) + '/client/ws',
+    controlWebSocketEndpoint(BASE_URL, location),
   )
 
   return (
--- a/packages/streamwall-control-client/src/wsEndpoint.test.ts
+++ b/packages/streamwall-control-client/src/wsEndpoint.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it } from 'vitest'
+
+import { controlWebSocketEndpoint } from './wsEndpoint.ts'
+
+describe('controlWebSocketEndpoint', () => {
+  it('derives ws:// from an http page when BASE_URL is "/"', () => {
+    expect(
+      controlWebSocketEndpoint('/', {
+        protocol: 'http:',
+        host: '127.0.0.1:3000',
+      }),
+    ).toBe('ws://127.0.0.1:3000/client/ws')
+  })
+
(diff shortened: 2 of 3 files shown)


NilsR0711:
Closing as a duplicate: #638 landed the same page-protocol-derived WebSocket scheme fix for #617 minutes earlier and is already merged.

PR #642, opened 2026-07-23
fix(streamwall): surface an error when a stalled view has no retry budget

### LINKED ISSUE #622: A stalled view with an exhausted retry budget shows a loading spinner forever instead of an error
## Summary
The stalled watchdog only leaves `stalled` when `canRetry` holds; when the retry budget is spent (or retry is disabled) there is no fallback transition, so the machine sits in `displaying.running.playback.stalled` indefinitely — which the overlay renders as a loading spinner.

## Evidence
`packages/streamwall/src/main/viewStateMachine.ts:650-661`:

```ts
stalled: {
  after: {
    stalledTimeout: {
      target: '#view.displaying.loading',
      guard: 'canRetry',
      actions: 'incrementRetry',
    },
  },
},
```

Contrast the load-error path, which stops retrying but lands in `error` with `context.error` populated. The overlay maps `playback.stalled` to `isLoading` (spinner). The existing test "does not reload a stalled view when retry is disabled" (`viewStateMachine.test.ts:274-288`) documents the machine staying in `stalled` forever.

## Failure scenario
A stream freezes permanently. After `maxRetries` stall-reloads the tile shows an eternal spinner with no error text — the operator gets no signal that the tile is dead, unlike the load-failure path.

## Suggested fix
When the stalled watchdog fires and `canRetry` is false, transition to `#view.displaying.error` with a "stream stalled" reason, mirroring the error-budget-exhausted behavior, so it surfaces on the wall and in the control UI.

## Impact
Medium - silent, terminal failure mode that is invisible to operators.

NilsR0711:
## Summary

The stalled watchdog (`displaying.running.playback.stalled`, `after: stalledTimeout`) only had a single transition guarded by `canRetry`. When the retry budget was spent — retry disabled or `maxRetries: 0` — the guard blocked the only exit, so the machine sat in `stalled` indefinitely, which the wall overlay maps to `isLoading` and renders as an eternal loading spinner. The operator got no signal that the tile was dead, unlike the load-failure path which lands in the terminal `error` state with a reason.

## Changes

- `viewStateMachine.ts`: the stalled watchdog now has a fallback transition — when `canRetry` is false it targets `#view.displaying.error`, logging and recording a `Stream stalled` reason (exported as `STALLED_ERROR_MESSAGE`). This mirrors the exhausted-budget behavior of the `error` state's backoff transition. No new state values are introduced (`displaying.error` already exists), so `viewStateValueSchema` in streamwall-shared is unaffected.
- `viewStateMachine.test.ts`:
  - Replaced the test that documented the old behavior ("does not reload a stalled view when retry is disabled" — staying in `stalled` forever) with one asserting the transition to `displaying.error` with the recorded reason.
  - Added a regression test for the no-budget case (`maxRetries: 0`), asserting the error state is terminal (no further auto-reload).

Note: because reaching `running` resets the retry streak (`resetRetryState` on entry), `retryCount` is always 0 while in `stalled`; `canRetry` is false there exactly when retry is disabled or `maxRetries` is 0. The tests cover both.

## Verification

- `npm run test`, `npm run lint`, `npm run typecheck`, Prettier check — all green locally.

Closes #622

### DIFF
--- a/packages/streamwall/src/main/viewStateMachine.test.ts
+++ b/packages/streamwall/src/main/viewStateMachine.test.ts
@@ -11,8 +11,11 @@ vi.mock('electron', () => ({
 }))
 
 const { createActor, fromPromise, matchesState } = await import('xstate')
-const { default: viewStateMachine, DEFAULT_RETRY_CONFIG } =
-  await import('./viewStateMachine')
+const {
+  default: viewStateMachine,
+  DEFAULT_RETRY_CONFIG,
+  STALLED_ERROR_MESSAGE,
+} = await import('./viewStateMachine')
 type RetryConfig = import('./viewStateMachine').RetryConfig
 
 const noop = () => {}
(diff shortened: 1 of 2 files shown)


NilsR0711:
Out-of-scope follow-up filed while working on this: #645 (repeated stall-reload cycles never exhaust the retry budget or grow backoff, since reaching running resets the streak).

PR #646, opened 2026-07-23
feat(control-ui): revalidate /admin/status so long sessions see updates

### LINKED ISSUE #624: useServerStatus fetches /admin/status exactly once, so the update banner never appears in long sessions
## Summary
`useServerStatus` runs a single fetch on mount (its only dependency, `enabled`, is stable for a session), with no timeout and no revalidation, and casts the JSON response unchecked.

## Evidence
`packages/streamwall-control-ui/src/useServerStatus.ts:22-45`:

```ts
useEffect(() => {
  if (!enabled) { setStatus(null); return }
  let cancelled = false
  fetch('/admin/status', { credentials: 'same-origin' })
    .then((res) => (res.ok ? (res.json() as Promise<ServerStatus>) : null))
    ...
}, [enabled])
```

## Failure scenario
- The server refreshes its update check daily, but an operator who keeps the control UI open (walls run for days) never sees the "update available" notice advertised in `docs/self-hosting.md` until a manual reload.
- A hung fetch (no `AbortController`) leaves `status` permanently `null` for the session — the version label and `ServerUpdateBanner` silently never render.
- `res.json() as Promise<ServerStatus>` is the one cross-boundary payload in the repo without runtime validation (everything else goes through zod in `streamwall-shared/src/schemas.ts`).

## Suggested fix
Add an `AbortController` timeout, re-fetch on an interval (e.g. hourly), and validate the body with a small zod schema.

## Impact
Low - degraded/absent update notice; no data corruption.

NilsR0711:
## Summary
`useServerStatus` fetched `GET /admin/status` exactly once on mount (its only dependency, `enabled`, is stable for a session), so:

- an operator who keeps the control UI open for days (walls run for days) never saw a newly published "update available" notice until a manual reload;
- a hung fetch left `status` permanently `null` for the session, so the version label and `ServerUpdateBanner` silently never rendered;
- the response body was cast unchecked (`res.json() as Promise<ServerStatus>`), the one cross-boundary payload in the repo without runtime validation.

## Changes
- **Interval revalidation** — poll `/admin/status` on an hourly interval while `enabled`, cleaning up the interval on unmount so no polling leaks after the component is gone. Hourly is ample since the server refreshes its update check daily.
- **Abort timeout** — each request runs under an `AbortController` with a 10s timeout, so a hung fetch can no longer stick for the session; the next interval retries. Overlapping polls supersede any still-inflight request.
- **Runtime validation** — the body is parsed with a small zod schema (`serverStatusSchema`, mirroring the server's `UpdateStatus`); a malformed payload yields no status rather than a mistyped object leaking into the UI. `ServerStatus` is now inferred from the schema.
- **Last-known-good on transient failure** — a single network blip or timeout no longer blanks a working version label / banner; the previous status stays until the next successful poll.

## Tests
Added vitest fake-timer coverage in `useServerStatus.test.tsx`: interval re-fetch, interval + abort cleanup on unmount (asserts the inflight `AbortSignal` is aborted and no further polls fire), zod rejection of a malformed body, and last-known-good retention across a failed refresh. Existing `fetch` call-site assertions updated for the new `signal` argument.

Closes #624

### DIFF
--- a/packages/streamwall-control-ui/src/serverStatusGating.test.tsx
+++ b/packages/streamwall-control-ui/src/serverStatusGating.test.tsx
@@ -98,9 +98,10 @@ describe('server status gating (#436)', () => {
       json: () => Promise.resolve(ADMIN_STATUS),
     })
     const el = await renderControlUI('admin')
-    expect(fetchMock).toHaveBeenCalledWith('/admin/status', {
-      credentials: 'same-origin',
-    })
+    expect(fetchMock).toHaveBeenCalledWith(
+      '/admin/status',
+      expect.objectContaining({ credentials: 'same-origin' }),
+    )
     expect(el.querySelector('.server-version-label')?.textContent).toContain(
       '0.9.1',
     )
--- a/packages/streamwall-control-ui/src/useServerStatus.test.tsx
+++ b/packages/streamwall-control-ui/src/useServerStatus.test.tsx
@@ -1,7 +1,11 @@
 import { render } from 'preact'
 import { act } from 'preact/test-utils'
 import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest'
-import { type ServerStatus, useServerStatus } from './useServerStatus.ts'
+import {
+  REFRESH_INTERVAL_MS,
+  type ServerStatus,
+  useServerStatus,
+} from './useServerStatus.ts'
 
 let container: HTMLDivElement | undefined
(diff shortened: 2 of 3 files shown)


NilsR0711:
Follow-up (out of scope for this PR): #649 - share the `/admin/status` shape via a single zod schema in `streamwall-shared` so the server `UpdateStatus` interface and the UI `serverStatusSchema` can't drift.

PR #648, opened 2026-07-23
fix(streamwall): tear down StreamWindow ipcMain handlers on dispose

### LINKED ISSUE #629: StreamWindow registers global ipcMain handlers with no teardown, so a second instance would crash
## Summary
`setupIpcHandlers()` registers `ipcMain.handle('layer:load' | 'view-init', ...)` and several `ipcMain.on(...)` listeners in the constructor, capturing `this`, with no `removeHandler`/`removeListener` anywhere and no dispose path on the class.

## Evidence
`packages/streamwall/src/main/StreamWindow.ts:194-262`. `ipcMain.handle` throws on duplicate channel registration, so constructing a second `StreamWindow` (e.g. a future recreate-on-config-reload path) crashes at `ipcMain.handle('layer:load')`. The handlers also outlive the window and dereference `this.overlayView.webContents` unconditionally (e.g. `devtools-overlay` at `:259`), which throws once those webContents are destroyed. This is masked today because `main()` constructs exactly one instance for the process lifetime, and the unit tests stub `ipcMain.handle` as a no-op (`StreamWindow.test.ts:20`), hiding the duplicate-registration throw.

## Suggested fix
Store handler references and add a `dispose()` that calls `ipcMain.removeHandler`/`ipcMain.removeListener`; guard the send/devtools paths against destroyed webContents. A test constructing two instances (with a real-enough ipcMain stub that throws on duplicates) would pin the contract.

## Impact
Low - latent; blocks any future multi-window or recreate flow and hides behind the current singleton.

NilsR0711:
## Summary
`StreamWindow.setupIpcHandlers()` registered `ipcMain.handle('layer:load' | 'view-init', ...)` and several `ipcMain.on(...)` listeners in the constructor, capturing `this`, with no `removeHandler`/`removeListener` and no dispose path on the class.

Since `ipcMain` is a process-global singleton and `ipcMain.handle` throws on duplicate channel registration, constructing a second `StreamWindow` (e.g. a future recreate-on-config-reload path) would crash at `ipcMain.handle('layer:load')`. The listeners also outlived the window and dereferenced the overlay's `webContents` unconditionally in the `devtools-overlay` handler, which throws once those webContents are destroyed. This was masked today because `main()` constructs exactly one instance for the process lifetime and the unit tests stubbed `ipcMain.handle` as a no-op, hiding the duplicate-registration throw.

## Changes
- Record a teardown thunk for every `ipcMain` handler/listener registered in the constructor (via new `registerIpcHandle` / `registerIpcOn` helpers).
- Add an idempotent `dispose()` that calls `ipcMain.removeHandler` / `ipcMain.removeListener` to release the process-global channels.
- Guard the `devtools-overlay` handler against an already-destroyed overlay `webContents` (`isDestroyed()` check).
- Upgrade the test's `ipcMain` double to a realistic stub that tracks registered channels and throws on duplicate `handle` (mirroring Electron), plus `removeHandler`/`removeListener`/`isDestroyed`. This pins the duplicate-registration contract that the previous no-op stub hid.

## Tests
New `StreamWindow.dispose` suite: dispose removes every registered handler/listener; a second instance can only construct after the first is disposed (the concrete crash from the issue); constructing + disposing twice does not throw; dispose is idempotent; and `devtools-overlay` no-ops on a destroyed overlay webContents. Full streamwall package suite green (900 tests).

Closes #629

### DIFF
--- a/packages/streamwall/src/main/StreamWindow.test.ts
+++ b/packages/streamwall/src/main/StreamWindow.test.ts
@@ -12,13 +12,53 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
  * constructor test can assert *what* was built and *in which order* without a
  * real Electron runtime.
  */
-const electronStub = vi.hoisted(() => ({
-  windows: [] as Array<InstanceType<typeof import('electron').BrowserWindow>>,
-  webContentsViews: [] as Array<
-    InstanceType<typeof import('electron').WebContentsView>
-  >,
-  ipcMain: { handle: vi.fn(), on: vi.fn() },
-}))
+const electronStub = vi.hoisted(() => {
+  // A stand-in for Electron's process-global `ipcMain`. It tracks which
+  // channels are currently registered and throws on a duplicate `handle`,
+  // mirroring the real runtime, so tests can pin the duplicate-registration
+  // crash (issue #629) and verify `dispose()` releases the channels again.
+  const registeredHandlers = new Set<string>()
+  const registeredListeners = new Map<string, Set<(...args: never[]) => void>>()
+  const handle = vi.fn((channel: string) => {
+    if (registeredHandlers.has(channel)) {
+      throw new Error(`Attempted to register a second handler for '${channel}'`)
+    }
+    registeredHandlers.add(channel)
+  })
+  const removeHandler = vi.fn((channel: string) => {
+    registeredHandlers.delete(channel)
+  })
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up (out of scope for this tightly-scoped ipcMain-teardown PR): #651 tracks guarding the `onState` overlay/background `webContents.send` paths against a destroyed webContents — the sibling of the `devtools-overlay` guard added here, and the other half of #629's "guard the send/devtools paths" note.

PR #650, opened 2026-07-23
perf(streamwall): resolve wall view cells via a byId stream index

### LINKED ISSUE #628: deriveWallViews scans the stream list linearly per grid cell on every state tick
## Summary
`deriveWallViews` resolves each cell's stream with `streams.find(...)`, an O(cells × streams) scan that runs on every data-source tick, every `viewsState` observer fire (each drag/collab edit), and every resize.

## Evidence
`packages/streamwall/src/main/wallViews.ts:59` and `:74`:

```ts
const stream = streams.find((s) => s._id === streamId)   // per fullscreen lookup
for (const [key, viewData] of viewsState) {
  const stream = streams.find((s) => s._id === streamId) // per cell, every call
```

`StreamList` already carries a lookup map for URLs (`byURL`, `streamwall-shared/src/types.ts:57`) but none for ids.

## Why it matters
With aggregated JSON feeds (hundreds of streams) and a dense grid this is thousands of comparisons per tick for what should be an O(1) lookup. Not user-visible today, but it sits on the hottest state path in the main process.

## Suggested fix
Add a `byId` map to `StreamList` alongside `byURL` (built in the same place) and use it in `deriveWallViews`.

## Impact
Low - straightforward efficiency cleanup on a hot path.

NilsR0711:
## Problem

`deriveWallViews` resolved each grid cell's stream with `streams.find(...)`, an O(cells × streams) scan that runs on every data-source tick, every `viewsState` observer fire (each drag/collab edit), and every resize. With aggregated JSON feeds (hundreds of streams) and a dense grid this is thousands of comparisons per tick on the hottest state path in the main process. `StreamList` already carried a `byURL` lookup index, but none for ids.

## Solution

- Add an optional `byId` map to `StreamList` alongside `byURL` (`streamwall-shared/src/types.ts`).
- Build it in `combineDataSources` right where `byURL` is built, after the id generator has assigned ids; entries the generator skipped (no usable id base) are omitted rather than keyed on `undefined`.
- `deriveWallViews` now resolves both the fullscreen override and each grid cell through that index. Lists that don't carry the index fall back to building a map once per call, so lookups stay O(1) per cell either way.

## Testing

- New tests: `combineDataSources` attaches the `byId` index and omits id-less entries; `deriveWallViews` resolves grid cells and the fullscreen stream through the index (asserting the index is the lookup actually used).
- Full quality gate: `npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` — all green.

Closes #628

### DIFF
--- a/packages/streamwall-shared/src/types.ts
+++ b/packages/streamwall-shared/src/types.ts
@@ -54,7 +54,10 @@ export interface StreamData extends StreamDataContent {
 
 export type LocalStreamData = Omit<StreamData, '_id' | '_dataSource'>
 
-export type StreamList = StreamData[] & { byURL?: Map<string, StreamData> }
+export type StreamList = StreamData[] & {
+  byURL?: Map<string, StreamData>
+  byId?: Map<string, StreamData>
+}
 
 /**
  * Mirrors `viewStateMachine.ts`'s snapshot `.value`. Derived from the schema so
--- a/packages/streamwall/src/main/data/combine.test.ts
+++ b/packages/streamwall/src/main/data/combine.test.ts
@@ -191,6 +191,36 @@ describe('combineDataSources', () => {
     }
   })
 
+  test('attaches a byId index to the yielded list for quick lookups', async () => {
+    async function* sourceA() {
+      yield [{ kind: 'video', link: 'https://a.example/s', source: 'Example' }]
+    }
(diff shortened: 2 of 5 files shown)


NilsR0711:
Closing as duplicate: issue #628 was just fixed by the merged PR #654 (perf(streamwall): resolve wall-view cells via a byId index).

Issue #621, opened 2026-07-23
A parked (paused) view starts playing again when its media is re-acquired or swapped


NilsR0711:
## Summary
The preload keeps no parked/paused flag, and re-acquisition unconditionally plays and unmutes. A related main-side gap: `resyncSwappedView` re-applies mute/options/volume but not pause.

## Evidence
Preload — `packages/streamwall/src/preload/mediaPreload.ts`:

```ts
video.play()          // :503 — findMedia always plays
...
video.muted = false   // :520 — and unmutes
```

The `pause` IPC handler (`:672-686`) pauses `currentMedia`, but nothing records the desired state, so the `'emptied'` re-acquisition path (`:597-626`) brings the new element up playing.

Main — `packages/streamwall/src/main/viewStateMachine.ts:314-321`:

```ts
resyncSwappedView: ({ context }) => {
  const { view, options, volume, desiredAudio } = context
  view.webContents.audioMuted = desiredAudio === 'muted'
  if (options) view.webContents.send('options', options)
  view.webContents.send('volume', volume)   // no pause re-send
}
```

The `pause` region only emits `sendViewPause` on entry to `paused`, which an in-place swap or internal recovery does not re-trigger.

## Failure scenario
A view is parked behind a fullscreen expansion (`--park.pause`, issue #374 flow). An HLS teardown empties the `<video>`, triggering re-acquisition, which calls `video.play()` and unmutes. The hidden view resumes decoding — and potentially audio — defeating the park optimization. Today the swap path is masked because unparking sends `RESUME`, but that pairing is the only thing holding the invariant.

## Suggested fix
Track a `desiredPaused` flag in the preload (set/cleared by the `pause`/`resume` IPC handlers) and re-apply it after acquisition — pause via `HTMLMediaElement.prototype.pause.call(media)` and skip the unmute. In `resyncSwappedView`, re-send `pause` when `context.desiredPaused`. Add regression tests for both paths.

## Impact
Medium - hidden tiles can resume decoding/audio; the invariant currently depends on an unrelated RESUME message.

PR #652, opened 2026-07-23
fix(streamwall): re-apply operator-set rotation when media is re-acquired

### LINKED ISSUE #620: Operator-set rotation is lost when media is re-acquired after a stall
## Summary
Volume survives the `'emptied'` re-acquisition because `main()` tracks `latestVolume` and re-applies it, but rotation has no equivalent: the fresh `RotationController` starts at 0° and no `'options'` IPC is re-sent.

## Evidence
`packages/streamwall/src/preload/mediaPreload.ts`:

```ts
let latestVolume = initialVolume ?? 1                         // :575 — tracked
volumeController = new VolumeController(media, latestVolume)  // :587 — re-applied every acquire
...
rotationController = new RotationController(media)            // :591 — customRotation resets to 0
```

`updateOptions` (`:654-660`) only runs on the initial call and on incoming `'options'` IPC. On the main side, the internal stall recovery drives `VIEW_STALLED → playback.playing` only (`viewStateMachine.ts`), which does not re-send options.

## Failure scenario
Operator rotates a stream 90°. The stream briefly stalls (`'emptied'` → re-acquire, common for live streams). The new controller starts at 0°, so the tile silently snaps back to un-rotated until the operator re-rotates manually.

## Suggested fix
Track `latestRotation` in `main()` (mirroring `latestVolume`) and re-apply it when constructing the `RotationController` in `acquireMedia`. Add a regression test: dispatch `'options'`, fire `'emptied'`, assert the re-acquired element keeps its rotation class.

## Impact
Medium - operator-visible state silently reverts during routine stream hiccups.

NilsR0711:
## Problem

Volume survives the `'emptied'` re-acquisition because `main()` tracks `latestVolume` and re-applies it, but rotation had no equivalent: the fresh `RotationController` starts at 0 and no `'options'` IPC is re-sent by the internal stall recovery. An operator-set rotation silently snapped back to un-rotated whenever a live stream briefly stalled. The initial rotation from the `view-init` options was affected by the same gap, since `updateOptions(initialOptions)` runs before the first acquisition finishes constructing the controller.

## Solution

Track `latestRotation` in `main()` (mirroring `latestVolume`), seeded from the initial `view-init` options and updated by every `'options'` IPC, and re-apply it via `rotationController.setCustom(latestRotation)` whenever a controller is (re-)created in `acquireMedia`.

## Testing

- New regression tests in `mediaPreload.test.ts`:
  - initial rotation from `view-init` options is applied once media is acquired
  - an operator-set rotation is re-applied to the element re-acquired after an `'emptied'` stall
  - later options updates keep working on the re-acquired element
- Full quality gate: `npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` all pass.

Closes #620

### DIFF
--- a/packages/streamwall/src/preload/mediaPreload.test.ts
+++ b/packages/streamwall/src/preload/mediaPreload.test.ts
@@ -755,6 +755,106 @@ describe('mediaPreload MutationObserver lifecycle (issue #412)', () => {
   })
 })
 
+describe('mediaPreload rotation across re-acquisition (issue #620)', () => {
+  // Must match SCAN_THROTTLE in mediaPreload.ts: long enough for the
+  // re-acquisition's throttled scan to find the replacement element.
+  const SCAN_THROTTLE_MS = 500
+
+  beforeEach(() => {
+    vi.useFakeTimers()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
+    vi.resetModules()
+    invoke.mockClear()
+    send.mockClear()
+    on.mockClear()
+    exposeInMainWorld.mockClear()
+    document.body.innerHTML = ''
+  })
+
+  function registeredHandler(
+    channel: string,
+  ): (ev: unknown, ...args: unknown[]) => void {
+    const call = on.mock.calls.find(([ch]) => ch === channel)
+    if (!call) {
+      throw new Error(`no ipcRenderer.on('${channel}', ...) handler registered`)
+    }
+    return call[1] as (ev: unknown, ...args: unknown[]) => void
+  }
+
+  // happy-dom's HTMLVideoElement never implements videoWidth, so give it a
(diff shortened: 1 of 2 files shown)


NilsR0711:
Heads-up: PR #657 covers the same rotation fix (tracking the latest rotation in the preload and re-applying it on every acquisition, including the initial view-init options) plus the closely related paused-state loss from #621 (preload desiredPaused re-assert and a pause re-send in resyncSwappedView), with regression tests for both. If #657 lands first, this PR becomes redundant and can be closed; if this one lands first, #657 will rebase and drop the overlapping rotation change.

NilsR0711:
Closing: superseded by the merged PR #657, which fixes rotation (#620) and paused state (#621) across media re-acquisition with regression tests.

Issue #671, opened 2026-07-23
Parallel forge publish jobs race and split the release into two drafts


NilsR0711:
## Problem

The v0.10.2 release run (release.yml run 30041776569) produced **two** draft releases for the same tag:

- one draft holding the macOS/Windows assets plus `latest.yml`/`latest-mac.yml` and the generated release notes
- a second, notes-less draft holding only the Linux assets (`.rpm`, `.deb`)

The three `Publish (<os>)` jobs run in parallel and each lets electron-forge's GitHub publisher look up the release for the tag; when none exists yet, two jobs can pass that check concurrently and each creates its own draft (classic check-then-create race). v0.10.1 happened to serialize by chance, so this is intermittent.

## Impact

`gh release edit v0.10.2` becomes ambiguous, the published release would silently miss the Linux installers, and the update feeds (`latest.yml`) never reference them. For v0.10.2 this was repaired manually (assets moved to the notes-bearing draft, duplicate deleted) before publishing.

## Suggested solution

Create the draft release exactly once before the publish matrix fans out (e.g. a lightweight `prepare-release` job using `gh api` / `gh release create --draft`), so every forge publisher finds the existing draft and only appends assets. Alternatively, serialize the publish jobs with `needs:`/concurrency, at the cost of wall-clock time.

Related: #533 (asset-count warning) would have caught the missing Linux assets after publish, but preventing the split is the proper fix.

PR #677, opened 2026-07-26
test(ci): guard against a vitest environment resolved only by hoisting

### LINKED ISSUE #676: CI guard: fail fast when a test runtime dependency is only hoisted, not declared
Follow-up from #675.

Until #675 the vitest test environment ran on a `happy-dom` root `node_modules` entry that no `package.json` in this repo declared. npm had hoisted it because vitest lists `happy-dom` as an optional peer dependency (`"optional": true, "peer": true` in the lockfile). Any lockfile regeneration was free to drop it, and one eventually did — the find-my-way bump in #670 turned all three `Test` jobs red with `Cannot find package 'happy-dom'` and `no tests`, 49 errors, before a single assertion ran.

#675 fixes the specific case by declaring `happy-dom` at the root. What is missing is a guard against the class of failure: the symptom (`ERR_MODULE_NOT_FOUND` at worker startup, zero tests executed, exit 1) is indistinguishable at a glance from a genuine test failure, and it only surfaces in CI on an unrelated dependency PR.

Possible shapes for the guard, in rough order of cost:

- A check in `test/` (alongside the existing `ci-gate.test.mjs` style tests) that asserts every root-hoisted entry a test runner imports at runtime is declared in some `package.json` — or, more narrowly, that `happy-dom` and any future test-environment package appear in root `devDependencies`.
- A CI step that greps the lockfile for `"optional": true` combined with `"peer": true` on packages the test setup depends on, and fails with a pointed message.
- Distinguishing "no tests ran" from "tests passed" in the test job, so a zero-test run can never be reported as success by a future workflow change.

The first option is the most direct and matches how `test/ci-gate.test.mjs` already guards workflow invariants.

NilsR0711:
## Summary

Adds a CI-enforced invariant guarding against the class of failure from #671/#675: a vitest DOM environment (e.g. `happy-dom`) resolving only through npm's optional-peer hoisting instead of an explicit dependency. That inference is silently dropped by a lockfile regeneration unrelated to the environment package itself — the find-my-way bump in #670 did exactly this, turning all three `Test` jobs red with `Cannot find package 'happy-dom'` before a single assertion ran.

#675 fixed that specific instance by declaring `happy-dom` as a root `devDependency`. This closes the gap the issue identifies: nothing stopped the same class of failure from recurring for `happy-dom` again, or for any future test-environment package.

Closes #676

## Technical solution

`test/vitest-environment-deps.test.mjs` adds two invariant checks, run as part of `npm test`:

1. **Every vitest environment package actually used is declared as a root dependency.** The check scans every workspace's `vitest.config.ts` for a configured `test.environment`, and every `*.test.ts(x)` file for a `// @vitest-environment` pragma override, then asserts each distinct environment name (excluding vitest's built-in `node` default) appears in the root `package.json`'s `dependencies`/`devDependencies`.
2. **The lockfile doesn't resolve it through hoisting alone.** For the same set of environments, asserts the corresponding `package-lock.json` `node_modules/<name>` entry isn't still marked both `"optional": true` and `"peer": true` — the literal symptom of "installed only because vitest's peer dependency happened to pull it in."

Both checks currently pass with `happy-dom` as the only environment in use (configured in `streamwall-control-client` and `streamwall-control-ui`'s `vitest.config.ts`, and via pragma in several `streamwall` renderer/preload test files).

## Architecture decisions

- Followed the existing `test/ci-gate.test.mjs` / `test/workspace-metadata.test.mjs` convention: a plain `node:test` file reading real repository files directly, no separate `scripts/` module or standalone CLI entry point, since this only needs to run as part of `npm test` (unlike e.g. `check-deprecated-deps.mjs`, which is also invoked from its own scheduled workflow).
- Chose the issue's first suggested guard shape (declaration check in `test/`) and folded in its second shape (lockfile optional/peer grep) as a cheap second assertion, since both map naturally onto the same test file. The issue's third suggested shape — distinguishing "zero tests ran" from "tests passed" at the CI job level — is a materially separate, more invasive change to the test job's success detection across all three OS legs, so it's filed as its own follow-up instead of bundled in here.
- The root-dependency requirement (rather than "declared anywhere") deliberately matches the actual fix in #675: vitest itself is hoisted to the root `node_modules` and resolves its environment relative to its own location, so a workspace-local declaration alone does not prevent the root copy from being inferred as optional/peer (that state existed in this repo before #670 even though `streamwall`, `streamwall-control-client`, and `streamwall-control-ui` had already declared `happy-dom` in their own `package.json`s via an earlier Dependabot bump).

## Testing performed

- TDD red/green: temporarily restored `package.json`/`package-lock.json` to their pre-#675 (`f82eaa7~1`) state and confirmed both new assertions fail with the intended message; restored the current files and confirmed both pass.
- `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm test` (root + all workspaces) all pass locally.

## Risks/breaking changes

None — this only adds a new test file; no production code changes.

### DIFF
--- a/test/vitest-environment-deps.test.mjs
+++ b/test/vitest-environment-deps.test.mjs
@@ -1,0 +1,132 @@
+import assert from 'node:assert/strict'
+import { readFileSync, readdirSync } from 'node:fs'
+import { dirname, join } from 'node:path'
+import { test } from 'node:test'
+import { fileURLToPath } from 'node:url'
+
+const rootDir = dirname(dirname(fileURLToPath(import.meta.url)))
+
+function readJson(relativePath) {
+  return JSON.parse(readFileSync(join(rootDir, relativePath), 'utf8'))
+}
+
+function readDocIfExists(relativePath) {
+  try {
+    return readFileSync(join(rootDir, relativePath), 'utf8')
+  } catch (err) {
+    if (err.code === 'ENOENT') {
+      return null
+    }
+    throw err
+  }
+}
+
+function findTestFiles(workspace) {
+  const results = []
+
+  function walk(relativeDir) {
+    for (const entry of readdirSync(join(rootDir, workspace, relativeDir), {
+      withFileTypes: true,
+    })) {
+      if (entry.name === 'node_modules' || entry.name === 'dist') {
+        continue
+      }
+
+      const entryRelativeDir = join(relativeDir, entry.name)
+      if (entry.isDirectory()) {
+        walk(entryRelativeDir)
+      } else if (/\.test\.tsx?$/.test(entry.name)) {
+        results.push(join(workspace, entryRelativeDir))
+      }
+    }
+  }
+
+  walk('.')
+  return results
+}
+
+// vitest resolves a non-default `test.environment` (e.g. happy-dom) as a
+// package import, set either as a workspace-wide config default or a
+// per-file `@vitest-environment` pragma comment overriding it.
+function environmentFromConfig(source) {
+  return source.match(/environment:\s*['"]([^'"]+)['"]/)?.[1]
+}
+
+function environmentFromPragma(source) {
+  return source.match(/^\/\/\s*@vitest-environment\s+(\S+)/m)?.[1]
+}
+
+// Every vitest environment package this repo actually relies on to run its
+// tests, gathered from every workspace's config and test files. `node` is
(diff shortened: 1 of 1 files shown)


NilsR0711:
Follow-up filed: #678 (distinguishing a test job that ran zero tests from one that passed — the issue's third suggested guard shape, out of scope for this PR).

PR #718, opened 2026-09-02
fix(ci): let the release check see draft releases

### LINKED ISSUE #698: Scheduled workflow failing: Release tag
The scheduled workflow `Release tag` failed: https://github.com/NilsR0711/streamwall/actions/runs/30348928077

Scheduled runs notify nobody, so this issue is the notification. It is
reused for every further failure and closed automatically once the
workflow runs green again, so please leave it open until then.

Filed by `.github/workflows/scheduled-report.yml`.

NilsR0711:
## What was wrong

The daily `Release tag` run has failed on every scheduled run since 2026-07-28
(#698). Two things went wrong at once:

1. **The release for `v0.10.5` was never taken out of draft.** `release.yml`
   creates the release as a draft for the publish matrix to append to;
   publishing it is a manual step and was forgotten.
2. **The check could not say so.** `check-release-assets.mjs` asked
   `GET /repos/{owner}/{repo}/releases/tags/{tag}`, and GitHub attaches a
   release to its tag only once it is published — a draft answers 404 there no
   matter how privileged the token is. The check read the complete, ready-to-go
   draft as a tag that had produced nothing at all and reported:

   > `v0.10.5` has no GitHub Release, so the tag shipped no installers. Re-run
   > release.yml for the tag … or delete and re-push `v0.10.5`.

   Following that advice would have deleted the tag and thrown away the record
   of what shipped, without touching the actual problem.

## What this changes

- The release is looked up in the repository's release **listing**, which does
  carry drafts. Paging stops at the page that holds the release, and a tag past
  the paged window raises its own error instead of masquerading as "no
  release".
- When one tag carries two releases — the publish-matrix race of #671 — the
  published one decides the verdict.
- The listing includes drafts only for a token with push access, so the asset
  check moved into a job of its own with `contents: write` (it writes nothing;
  the checkout keeps `persist-credentials: false`). The tag check stays
  `contents: read`, and the two jobs are independent so a missing tag cannot
  hide a broken release.
- The missing-release report no longer suggests deleting a tag.
- `CONTRIBUTING.md` documents the job split and that a local run needs
  `GH_TOKEN` to see drafts at all.

## Verification

- The draft release `v0.10.5` was published (`--draft=false --latest`); its 7
  assets cover every expected artifact kind. Nothing was deleted.
- Against the live repository, the patched script reported
  `The release for v0.10.5 is still a draft … Publish it in the releases UI.`
  before publishing and
  `v0.10.5 has a published release with every expected artifact kind.` after —
  the exact diagnosis the old code could not produce.
- `node --test test/*.test.mjs`: 177 passing; `actionlint`, `eslint` and
  `prettier --check` clean.
- The next scheduled run closes #698 automatically; `release-tag.yml` is also
  dispatched manually after merge.

Refs #698

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -50,12 +50,37 @@ jobs:
       - name: Check the release version is tagged
         run: node scripts/check-release-tag.mjs
 
-      # A tag alone proves nothing: the publish matrix may have failed
-      # half-way through, or never run for the tag at all (#533). Runs even
-      # when the step above failed — the two report different halves of the
-      # same pipeline, and one missing tag must not hide a broken release.
+  # A tag alone proves nothing: the publish matrix may have failed half-way
+  # through, never have run for the tag at all (#533), or have left the release
+  # as an unpublished draft (#698). Deliberately independent of `check` — the
+  # two report different halves of the same pipeline, and one missing tag must
+  # not hide a broken release.
+  #
+  # A job of its own rather than a step of `check`, because it is the only part
+  # of this workflow that needs `contents: write`: GitHub lists draft releases
+  # only to a token with push access, and a draft is exactly what this check
+  # has to be able to see. With `contents: read` the unpublished release of
+  # v0.10.5 read as a tag that had produced nothing at all, and the report told
+  # the maintainer to re-push the tag. The scripts only read, and the checkout
+  # keeps `persist-credentials: false`, so the token never reaches git either.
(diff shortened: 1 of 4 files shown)


NilsR0711:
Follow-ups filed while working on this: #720 (release.yml leaves the finished release as a draft — the manual publish is what went unnoticed here) and #721 (a transient GitHub API error in the scheduled release checks files a failure report).

Issue #492, opened 2026-07-21
test(control-server): intermittent local test failure with fewer subtests reported


NilsR0711:
## Observation

While running the full workspace suite locally (macOS, Node 22) during work on #407, `npm test` failed once in the `streamwall-control-server` package and then passed on every subsequent run of the identical, unmodified tree.

The failing run reported:

```
# pass 145
# fail 0
npm error Lifecycle script `test` failed with error:
npm error command failed
```

Note the mismatch: `fail 0`, yet a non-zero exit — and only 145 subtests reported, versus 147 in the passing runs. That points at the test process dying (or a subtest never being registered/awaited) rather than an assertion failing. The surrounding output also contained the `failed to send client state delta` log line.

## Impact

- Non-deterministic red runs on developer machines and, potentially, in CI.
- A `node:test` run that exits non-zero while reporting `fail 0` is confusing to triage — the actual cause is not visible in the summary.

## Suggested investigation

- Re-run `npm -w packages/streamwall-control-server test` in a loop to reproduce and capture the full output of a failing run.
- Compare the registered subtest list between a passing and a failing run to identify which two subtests go missing.
- Check for unawaited async work / open handles in the WebSocket-based tests that could tear the process down early.

Related: #221 (previously fixed flake in the same package).

NilsR0711:
Root cause and fix landed in #537. Follow-up on the underlying per-test cost: #536.

Issue #776, opened 2026-09-02
Apply navigation and window-open guards to the overlay and background layer views


NilsR0711:
## Problem
`StreamWindow.createLayerView()` builds the background and overlay `WebContentsView`s — app pages loaded through `loadHTML`, carrying the `streamwallLayer` preload bridge — and passes them through no navigation hardening at all: no `setWindowOpenHandler`, no `will-navigate` / `will-redirect` guard.

That is the same gap #732 described for the Control Window, and after #732 it is fixable in one line with the `secureAppWindow()` helper that issue introduced.

## Evidence
- `packages/streamwall/src/main/StreamWindow.ts` — `createLayerView()` constructs both layers with `new WebContentsView({ webPreferences })` and never calls `secureStreamView`, `denyWindowOpen`, or `secureAppWindow`.
- By contrast `createRawView()` calls `secureStreamView()`, `bootstrap.ts`'s `createBrowseWindow()` calls `denyWindowOpen()`, and `ControlWindow` calls `secureAppWindow()` since #732.
- `packages/streamwall/src/main/navigationSecurity.ts` — `secureAppWindow(webContents, { isAppURL, openExternal })` is the helper meant for windows rendering the app's own bundled pages.

## Impact
No known trigger today: the `streamwallLayer` bridge is much narrower than `streamwallControl`, and no operator-supplied link is rendered as an anchor in those pages (overlay/background content is embedded in sandboxed iframes, whose navigations are a separate surface). This is a consistency and defence-in-depth gap rather than a live vulnerability — but it leaves two of the app's three app-page `webContents` outside the hardening the rest of the app has.

## Suggested fix
- Call `secureAppWindow(view.webContents, { isAppURL: isAppPageURL, openExternal: ... })` in `createLayerView()`, or at minimum `denyWindowOpen()` plus the navigation guard.
- Add a regression test alongside the existing `StreamWindow` tests asserting both layer views get the guards, mirroring the Control Window test added in #732.
- While there, consider whether `will-frame-navigate` is warranted for these views specifically, since unlike the Control Window they deliberately host third-party iframes.

## Related
#732 (control window navigation guards) — found while reviewing that fix; it introduced `secureAppWindow`. #5 (closed) scoped the original guards to stream views and the browse window. #733 covers the session/partition hardening of the same two layer views.

PR #803, opened 2026-09-02
fix(streamwall): let stream-view sub-frames resolve their own redirects

### LINKED ISSUE #794: Investigate whether secureStreamView wrongly cancels sub-frame redirects in stream views
## Problem
`secureStreamView`'s navigation guard treats a sub-frame redirect as the view trying to leave its page, and cancels it.

`will-redirect` fires for sub-frames as well as the main frame (see `isMainFrame` on `WebContentsWillRedirectEventParams` in `node_modules/electron/electron.d.ts`). `preventNavigationAway` compares the event's URL against `webContents.getURL()`, which is always the *main frame's* URL — a sub-frame's redirect target can never equal it, so every sub-frame 302 inside a stream view is `preventDefault()`ed.

## Evidence
- `packages/streamwall/src/main/navigationSecurity.ts` — `secureStreamView` registers the same `preventNavigationAway` guard on both `will-navigate` and `will-redirect`, and that function never reads `event.isMainFrame`.
- `packages/streamwall/src/main/StreamWindow.ts` — `createRawView()` applies `secureStreamView` to every stream view.
- The sibling helper `secureAppWindow` gained an opt-in `allowSubframeNavigation` for exactly this reason in #776; the `NavigationEvent.isMainFrame` field it introduced is deliberately optional because this guard does not read it.

## Impact
Stream views load arbitrary operator-supplied embed pages — precisely the content most likely to contain iframes that redirect: video players, CDN handoffs, consent frames, shortlinks. Any such sub-frame redirect is silently cancelled, which would surface as a broken or blank embed with no error, since the cancellation is not a main-frame `did-fail-load`.

It is not clear how often this bites in practice — no bug report is known — which is why it is filed for investigation rather than fixed blind. It is a pre-existing behaviour, not a regression from #776.

## Suggested fix
- Establish first whether real embeds are affected: load a stream whose player redirects a sub-frame and check for a cancelled navigation.
- If confirmed, let `preventNavigationAway` ignore `event.isMainFrame === false`, matching `secureAppWindow`'s opt-in. Sub-frame requests in a stream view are already governed at the network layer by the session's SSRF guard (`installRequestSSRFGuard`), so skipping them in the navigation guard does not remove the SSRF protection.
- Add a regression test alongside the existing `secureStreamView` cases in `navigationSecurity.test.ts`.

## Related
#5 (closed) introduced these guards. #776 added the sub-frame opt-in to `secureAppWindow` and left this asymmetry deliberately untouched; found while reviewing that PR.

NilsR0711:
## Summary

`secureStreamView` cancelled every sub-frame redirect inside a committed stream view.

The investigation the issue asked for came out positive, from Electron's own semantics:

- Electron's navigation throttle (`shell/browser/electron_navigation_throttle.cc`) gates `will-navigate` on `handle->IsInMainFrame()`, but emits `will-redirect` for **every** frame — no main-frame check. The pinned typings agree: `will-navigate` is documented as "start navigation on the main frame", while `WebContentsWillRedirectEventParams` carries an `isMainFrame` flag with no such restriction.
- `preventNavigationAway` compared the event URL against `webContents.getURL()`, which is always the *main frame's* URL. A sub-frame's redirect target can never equal it, so once a page had committed (`currentURL !== ''`) every sub-frame 302 was `preventDefault()`ed.

Stream views exist to host operator-supplied embeds — video players, CDN handoffs, consent frames, shortlinks — i.e. exactly the content whose iframes redirect. Such a cancellation would show up as a blank embed with no main-frame `did-fail-load` to report it.

The guard now returns early for `event.isMainFrame === false`, matching the opt-in `secureAppWindow` gained in #776. The skip is unconditional here rather than opt-in because `secureStreamView` has exactly one caller (`StreamWindow.createRawView()`), and that caller is the iframe-hosting case.

## Security

- Main-frame behaviour is untouched: navigations and redirects away are still cancelled, self-reloads still allowed, and the uncommitted-view allowance for the initial load's own redirect chain is unchanged.
- The comparison is `=== false`, not `!== true`, so an event that carries no frame information at all is still guarded as a main frame — it fails closed.
- No protection is lost for sub-frames: every stream view's session is hardened by `hardenSession` → `installRequestSSRFGuard` (`StreamWindow.ts:491`, before `secureStreamView` at `:498`), which re-checks every request hop against the non-public-address policy. A sub-frame's *initial* load was never gated by this navigation guard either.

## How was this tested?

`npm run format`, `npm run lint`, `npm run typecheck`, `npm run test` (2337 tests, green) at the repo root.

New unit coverage in `navigationSecurity.test.ts`:

- a sub-frame redirect (`isMainFrame: false`) resolves — this test fails on `main`, and is the reproduction of the reported bug;
- the same redirect reported as the main frame is still blocked (negative control);
- the existing "302 escape" case now explicitly documents that it dispatches an event with **no** frame information and must still be blocked. `FakeWebContents.dispatchNavigation` was adjusted to omit the `isMainFrame` key entirely in that case instead of setting it to `undefined`, so the fail-closed path is exercised against a genuinely key-less event shape. That is a no-op for every existing assertion (both shapes compare `!== false`).

No live Electron run was performed — no GUI environment is available here — so the "load a real redirecting embed" bullet was answered from Electron's throttle source and typings plus the unit-level reproduction above.

## Pre-PR review

Three rounds with an independent reviewer that had none of the implementing session's context.

- Round 1 — two minor findings, both accepted: (a) the new "reports no frame at all" test duplicated an existing case and the test double never produced a truly key-less event → the double now omits the key and the duplicate test was dropped in favour of documenting the existing case; (b) the commit body asserted the real-world impact as observed fact although no live embed was loaded → wording softened to the conditional and the substitution documented here.
- Round 2 — one minor finding, accepted: the two "still blocks a main-frame …" tests were redundant (the guard treats `true` and an absent key identically) → the `will-navigate` one was dropped, the `will-redirect` one kept as the negative control beside the new sub-frame test.
- Round 3 — `NO FINDINGS`.

No findings were dismissed.

Closes #794

## Checklist

### DIFF
--- a/packages/streamwall/src/main/navigationSecurity.test.ts
+++ b/packages/streamwall/src/main/navigationSecurity.test.ts
@@ -54,13 +54,16 @@ class FakeWebContents {
     isMainFrame?: boolean,
   ): boolean {
     let prevented = false
-    const navEvent: NavEvent = {
-      url,
-      isMainFrame,
-      preventDefault: () => {
-        prevented = true
-      },
+    const preventDefault = () => {
+      prevented = true
     }
+    // Omitted rather than set to `undefined`, so a caller that passes no frame
+    // information reproduces an event shape that carries no `isMainFrame` key
+    // at all -- the case every guard has to fail closed on.
+    const navEvent: NavEvent =
+      isMainFrame === undefined
+        ? { url, preventDefault }
+        : { url, isMainFrame, preventDefault }
     for (const listener of this.navHandlers[event] ?? []) {
       listener(navEvent)
     }
@@ -135,6 +138,9 @@ test("secureStreamView keeps a stream URL's signed-token query out of the reload
 })
 
 test('secureStreamView blocks a redirect away once a page has committed (302 escape)', () => {
+  // Dispatched without any frame information, so this pins the fail-closed
+  // path too: an event shape that reports no frame is guarded as a main-frame
+  // navigation rather than skipped.
   const wc = new FakeWebContents('https://example.com/stream')
   secureStreamView(asWebContents(wc))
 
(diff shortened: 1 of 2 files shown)


NilsR0711:
Follow-up filed for an out-of-scope finding noticed while running the quality gate for this change: #804 (control-server test runner exited non-zero once with zero reported failures; unreproducible afterwards, same signature as the closed #492). It is unrelated to this PR's diff.

Issue #790, opened 2026-09-02
Tell the operator when an overlay or background URL is rejected instead of failing blank


NilsR0711:
## Problem
An overlay or background whose URL the SSRF guard blocks now fails invisibly. Since #733 the chrome layers run in a hardened session, so a `link` pointing at a private-network or loopback address is cancelled at the network layer — which is the intended protection — but the operator sees only a blank layer. Nothing in the UI says the URL was rejected, and nothing tells them why.

## Evidence
- `packages/streamwall/src/main/partitions.ts` — `installRequestSSRFGuard`'s own doc comment justifies silent cancellation with "a cancelled main-frame navigation surfaces as a view error on the wall via the existing `did-fail-load` handler". That reasoning holds for stream views, whose `registerView()` wires `did-fail-load` to a view error. It does not hold for the layers: `StreamWindow.createLayerView()` registers no `did-fail-load`, and a cancelled load inside an *iframe* would not reach it anyway.
- `packages/streamwall/src/renderer/OverlayRoot.tsx` and `packages/streamwall/src/renderer/background.tsx` render `<OverlayIFrame src={s.link}>` / `<BackgroundIFrame src={s.link}>` with no error state.
- `s.link` never passes through `ensureValidURL`; that check lives only on the wall-tile path (`viewStateMachine.ts`).
- The only trace today is a `log.warn` from the request guard in the main process.

## Impact
An operator who mistypes a URL, or supplies one on the LAN, gets a blank layer and no explanation — a support burden, and easy to mistake for a bug in the layer rendering. This is a usability gap opened by a security fix, not a security hole; the pre-#733 behaviour (loading the internal URL) was worse.

## Suggested fix
- Validate `s.link` with `ensureValidURL` before rendering the iframe, so a rejected URL becomes a deliberate, reported outcome rather than a cancelled request.
- Surface it where the operator can see it: an error state on the layer tile and/or a `dataSourceHealth`-style report back to the control UI. The point of this issue is the operator-visible feedback; the validation is the means.
- Cover both layers (`OverlayRoot.tsx` and `background.tsx`) and add tests for the rejected-URL rendering path.

## Related
Raised in the suggested fix of #733 and again while reviewing its PR; deliberately left out of that change, which was scoped to the session hardening.

PR #809, opened 2026-09-02
feat(control-ui): name a refused layer URL where the operator typed it

### LINKED ISSUE #797: Report a refused overlay or background URL to the control UI, not only the wall
## Problem
A refused overlay or background URL is now named on the wall (#790), but not in the control UI — where the operator actually typed it.

In the control-server deployment the person adding a custom stream can be on a different machine entirely, with no view of the wall. For them the behaviour is unchanged from before #790: the layer silently does nothing.

## Evidence
- `packages/streamwall/src/main/StreamWindow.ts` — `createLayerView`'s `onBlocked` reports to the overlay layer's webContents only.
- `packages/streamwall/src/renderer/BlockedLayerNotice.tsx` renders the notice on the wall.
- Nothing carries the report into `StreamwallState`, so neither the local control window nor a remote control-UI client sees it.

## Impact
Usability, not security. The wall-side notice covers the single-operator desktop case; the remote-operator case still gets no explanation for a URL that quietly does nothing.

## Suggested fix
- Carry the blocked URLs into the state the control UI already receives — `dataSourceHealth` is the closest existing shape, and the control UI already renders a banner for it (`DataSourceHealthBanner.tsx`).
- Keep the wall notice: the two audiences are different, and the wall one needs no round trip.
- Mind what #790 established the hard way: the guard reports the request it cancelled, which for a redirect hop or a refused sub-resource is not the link the operator typed, so the report must not be presented as "this stream is broken" without evidence that it is.
- Reuse the existing report path rather than adding a second one: `installRequestSSRFGuard`'s `onBlocked` option already exists.

## Related
#790 (the wall-side notice) — raised while reviewing its PR. #733 introduced the hardened layer session whose guard does the refusing.

NilsR0711:
## Summary

Since #790 the wall names an overlay or background URL its hardened session refused, instead of leaving a blank layer behind. That only reaches somebody standing in front of the wall: in a control-server deployment the person adding a custom overlay can be on another machine entirely, and for them the layer still silently does nothing.

The refusal now travels into the broadcast state and is named in the control UI, beside the custom-stream inputs the link was typed into.

**How it works**

- `StreamWindow` emits every refusal the SSRF guard's existing `onBlocked` reporter hands it — alongside the wall's own notice, not instead of it, and without waiting for the overlay renderer to announce itself, which the control UI's path has no reason to inherit. Both deliveries are guarded, so neither audience can lose its notice to the other's transport failing.
- `createBlockedLayerURLReporter` (bootstrap) folds the reports into `StreamwallState.blockedLayerURLs` and takes every assembled state on its way out, which is where an operator's edit to a layer link clears the list.
- `BlockedLayerURLNotice` renders the addresses in the sidebar's Custom Streams section, dismissable per address.

**Architecture decisions**

- **A new bounded state field, not `dataSourceHealth`.** The issue named it as the closest existing shape; it is not the same thing (these are not configured data sources, and the banner's wording says so), so reusing it would have meant lying about what failed. The field carries the same enforcement instead: `MAX_BLOCKED_LAYER_URLS` and `MAX_BLOCKED_LAYER_URL_LENGTH` in `streamwallStateSchema`.
- **Bounded because it is layer content, not operator input.** Each URL is truncated, the list is capped, a URL already held is not re-broadcast, and a full list drops the newcomer rather than evicting the operator's own refused link — so a framed page polling a refused endpoint with a cache-busting query cannot grow the state or re-broadcast at request rate. Two URLs differing only past the truncation point collapse into one entry deliberately: they are the same string to anybody reading the notice.
- **`.default([])` on the schema field.** The desktop and the control server are released and deployed separately; rejecting a pre-#797 desktop's payload would take its whole state broadcast down rather than costing it one notice. `useStreamwallState` defaults it too, because a snapshot reaches the client without passing through the schema at all.
- **Withheld from `monitor`.** The addresses are on the wall's own network, and that role cannot manage custom streams in the first place, so `StateWrapper.view` hands it an empty list.
- **`layerLinksKey` moved into `streamwall-shared`.** The renderer's copy remounts the layer frames so a still-refused URL reports itself again; the main process's copy clears the broadcast list. Two copies could drift, and the notice would then be cleared with no remount behind it, losing the refusal for good. The single implementation is also order-insensitive (layer streams arrive from a polled data source, free to reorder) and no longer collides two link sets that embed its separator.
- **Says what was refused, never that a stream is broken.** As on the wall: the guard reports the request it cancelled, which for a redirect hop or a refused sub-resource is not the link the operator typed. The addresses render as plain text, never as links.

**Known limitation (documented in the component):** a dismissal is dropped once the desktop stops reporting that address, but a client disconnected across the operator's edit reconnects to a snapshot already naming the same address again — indistinguishable from it never having gone away, so it stays dismissed. Telling the two apart would mean carrying a clear generation through the state, which is more machinery than a dismissed notice is worth.

Closes #797

## How was this tested?

`npm run lint`, `npm run format:check`, `npm run typecheck` and `npm test` all pass locally (2384 tests).

New coverage on all three layers:

- **shared** (`schemas.test.ts`, `layerLinks.test.ts`): the field's bounds accepted at the limit and rejected past it in both dimensions, the omitted-field default, and the clear key's order-insensitivity and separator safety.
- **main** (`blockedLayerURLs.test.ts`, `bootstrap.test.ts`, `StreamWindow.test.ts`, `index.test.ts`): truncation, de-duplication, the cap dropping newcomers, the defensive copy, clear-on-edit and its no-op cases, the report reaching `updateState`, the absence of a broadcast loop, the emit happening before the overlay renderer is up and not after disposal, and each delivery surviving the other throwing. `index.test.ts` pins that every state update still passes through the reporter — reverting that one line left the suite green before.
- **control UI** (`BlockedLayerURLNotice.test.tsx`, `blockedLayerURLPlacement.test.tsx`, `streamwallState.test.tsx`): rendering, the live region staying mounted, addresses never becoming links, dismissal and its expiry, the full-list caption, the state default, and that the notice reaches an operator in the same section as the stream URL input while a `monitor` sees none.
- **control server** (`auth.test.ts`): the role gate in both directions.

## Risks / breaking changes

None expected. `StreamwallState` gains a required field, so the desktop, the control server and the control UI all move together; the schema default and the client-side default cover a version-skewed pair in both directions.

## Pre-PR review

Eight rounds with an independent reviewer that had none of the implementing context; the last returned no findings. Fixed along the way: the missing client-side default for a pre-#797 server (would have crashed the sidebar); the duplicated `layerLinksKey` and its order/separator sensitivity; the unbounded notice size in the sidebar; the tracker handing out its own array; the untested `updateState` seam, later made a real guard; the "list is full" caption describing the visible list rather than the desktop's; the wall and control-UI report paths being able to lose a notice to each other; and two tests that could not fail. No findings were dismissed.

## Checklist


NilsR0711:
Follow-ups filed while working on this: #810 (a dismissal outliving a clear the client did not see — the limitation documented in `BlockedLayerURLNotice`) and #811 (the wall-clock-dependent warm-session test in the control server, hit twice during review on this branch and unrelated to it).

